PR #2, opened 2018-04-19
Output appropriate client library namespace.

lukesneeringer:
This commit makes it such that library code is output into an appropriate import structure.

Prior to this, each service was written to the output directory at the top level. Now, the API's namespace and name are used. Therefore, for an API named "Foo" with version "v1", which has `service Bar`...

  * Before: `import bar`
  * Now: `from foo_v1 import bar`

And if the namespace is `("Spam", "Eggs")`...

  * Before: `import bar`
  * Now: `from spam.eggs.foo_v1 import bar`

Additionally, the API's name is directly importable, with and without the version:

  * Now: `from spam.eggs import foo`
  * Now: `from spam.eggs import foo_v1`

These are actually the preferred import approach.

Internally, this commit changes the file rename approach from just renaming filenames with templates beginning with `service/` to now substituting four variables (as documented).


lukesneeringer:
Good catch, thanks.
Note to self: Do not use `$` in `git commit` in bash. :-p

lukesneeringer:
The protocol buffer compiler itself will solve for that, no? It is hard to see `protoc` not choking on `service Foo/Bar {`.

Certainly the [style guide](https://developers.google.com/protocol-buffers/docs/style) mandates valid upper camel names, which precludes characters which can not be coerced into that.

landrito:
Namespace segments and version are not validated currently correct?

lukesneeringer:
Ah, yes, that is true.

PR #5, opened 2018-06-13
Add field headers

eoogbe:
Related googleapis/googleapis#505

Looks like
```diff
@functools.singledispatch
def create_device_registry(self,
        request: device_manager_pb2.CreateDeviceRegistryRequest, *,
        retry: retry.Retry = None,
        timeout: float = None,
        metadata: Sequence[Tuple[str, str]] = (),
        ) -> resources_pb2.DeviceRegistry:
    """Creates a device registry that contains devices.

    Args:
        request (~.device_manager_pb2.CreateDeviceRegistryRequest):
            The request object. Request for `CreateDeviceRegistry`.
        retry (~.retry.Retry): Designation of what errors, if any,
            should be retried.
        timeout (float): The timeout for this request.
        metadata (Sequence[Tuple[str, str]]): Strings which should be
            sent alont with the request as metadata.

    Returns:
        ~.resources_pb2.DeviceRegistry:
           A container for a group of devices.
    """
    # Wrap the RPC method; this adds retry and timeout information,
    # and friendly error handling.
    rpc = gapic_v1.method.wrap_method(
        self._transport.create_device_registry,
        default_retry=None,  # FIXME
        default_timeout=None,  # FIXME
        client_info=self.client_info,
    )

+  # Ensure metadata is a mutable sequence.
+  metadata = list(metadata)

+  # Add request headers to metadata.

+  try:
+      header_value = request.parent
+  except AttributeError:
+      # Suppress when attribute missing.
+      pass
+  else:
+      metadata.append((
+          'x-goog-request-params',
+          '{}={}'.format(
+              parse.quote_plus('parent'),
+              parse.quote_plus(header_value),
+          ),
+      ))

+  try:
+      header_value = request.device_registry.name
+  except AttributeError:
+      # Suppress when attribute missing.
+      pass
+  else:
+      metadata.append((
+          'x-goog-request-params',
+          '{}={}'.format(
+              parse.quote_plus('device_registry.name'),
+              parse.quote_plus(header_value),
+          ),
+      ))

    # Send the request.
    response = rpc(request, retry=retry, timeout=timeout, metadata=metadata)

    # Done; return the response.
    return response
```

### DIFF
--- a/api_factory/generator/templates/$namespace/$name_$version/$service/client.py.j2
+++ b/api_factory/generator/templates/$namespace/$name_$version/$service/client.py.j2
@@ -4,6 +4,9 @@
 import functools
 import pkg_resources
 from typing import Mapping, Optional, Sequence, Tuple, Union
+{%- if service.has_field_headers %}
+from urllib import parse
+{%- endif %}
 
 from google.api_core import gapic_v1
 {%- if service.has_lro %}
@@ -54,8 +57,8 @@ class {{ service.name }}:
         else:
             Transport = get_transport_class(transport)
             self._transport = Transport(credentials=credentials)
+    {%- for method in service.methods.values() %}
 
-    {% for method in service.methods.values() -%}
     @functools.singledispatch
     def {{ method.name|snake_case }}(self,
             request: {{ method.input.pb2_module }}.{{ method.input.name }}, *,
@@ -97,6 +100,38 @@ class {{ service.name }}:
             default_timeout=None,  # FIXME
             client_info=self.client_info,
         )
+        {%- if method.field_headers %}
+
+        # Ensure metadata is a mutable sequence.
+        if metadata is None:
+            metadata = []
+        else:
+            metadata = list(metadata)
+
+        # Add request headers to metadata.
+        try:
+            request_headers = (
+                {%- for field_header in method.field_headers %}
+                ('{{ field_header.header }}', '{{ field_header.field }}', request.{{ field_header.field }}),
+                {%- endfor %}
+            )
+        except AttributeError:
+            # Suppress when attribute missing.
+            pass
+        else:
+            request_metadata = [
+                (
+                    header,
+                    '{}={}'.format(
+                        parse.quote_plus(key),
+                        parse.quote_plus(value),
+                    ),
+                )
+                for (header, key, value)
+                in request_headers
+            ]
+            metadata.extend(request_metadata)
+        {%- endif %}
 
         # Send the request.
         response = rpc(request, retry=retry, timeout=timeout, metadata=metadata)
--- a/api_factory/generator/templates/$namespace/$name_$version/$service/client.py.j2
+++ b/api_factory/generator/templates/$namespace/$name_$version/$service/client.py.j2
@@ -103,34 +103,25 @@ class {{ service.name }}:
         {%- if method.field_headers %}
 
         # Ensure metadata is a mutable sequence.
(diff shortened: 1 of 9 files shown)


api_factory/generator/templates/$namespace/$name_$version/$service/client.py.j2
```diff
@@ -97,6 +100,38 @@ class {{ service.name }}:
             default_timeout=None,  # FIXME
             client_info=self.client_info,
         )
+        {%- if method.field_headers %}
+
+        # Ensure metadata is a mutable sequence.
+        if metadata is None:
```
lukesneeringer: Since `metadata` defaults to empty-tuple, this check is not necessary; `list(metadata)` is sufficient.

api_factory/generator/templates/$namespace/$name_$version/$service/client.py.j2
```diff
@@ -97,6 +100,38 @@ class {{ service.name }}:
...
+
+        # Add request headers to metadata.
+        try:
+            request_headers = (
+                {%- for field_header in method.field_headers %}
+                ('{{ field_header.header }}', '{{ field_header.field }}', request.{{ field_header.field }}),
+                {%- endfor %}
+            )
+        except AttributeError:
+            # Suppress when attribute missing.
```
lukesneeringer: Is this correct? Seems like this will make _any_ attribute's being missing suppress _all_ of them.

lukesneeringer: Also, since this is protocol buffers, the attribute is likely to be present-but-falsy often; do you want a truthiness check?

api_factory/schema/wrappers.py
```diff
@@ -187,4 +193,22 @@ def pb2_modules(self) -> Sequence[Tuple[str, str]]:
...
         """Return whether the service has a long-running method."""
-        return any(method.lro_payload for method in self.methods.values())
+        return self._any_method('lro_payload')
+
+    @property
+    def has_field_headers(self) -> bool:
+        """Return whether the service has a method containing field headers."""
+        return self._any_method('field_headers')
+
+    def _any_method(self, attr: str) -> bool:
```
lukesneeringer: Clever. I wonder if this is a concept worth trying to expose more generically.

What if we sent down `any` and `all` methods that accepted an iterable for the first argument and lambda function for the second?

Not sure if Jinja syntax would support sending the lambda though.

api_factory/generator/templates/$namespace/$name_$version/$service/client.py.j2
eoogbe: I couldn't think of a way to check the individual fields without being super repetitive or depending on an external function. Since most of the existing things that use request header params use 1 or at most 2 fields, I figured it wouldn't hurt too much to throw the whole thing away for one missing attribute.

`gapic-generator` has [the same problem](https://github.com/googleapis/gapic-generator/blob/master/src/test/java/com/google/api/codegen/testdata/py/python_library.baseline#L1564). I'm wondering if moving more of this logic to `api_core` is a better solution for both generators.

api_factory/schema/wrappers.py
eoogbe: I don't think this is possible in Jinja due to allowing only "view logic." I could make the function take in a lambda from its callers within the same wrapper class instead.

PR #6, opened 2018-07-12
Refactor to support the finalized-ish input format.

lukesneeringer:
This is a refactor to support the input format we landed on.

It does a lot of things:

  * Uses the new annotations.
  * Moves templates to `templates/` in the project root.
  * Removes support for flat files, which are no longer needed with @garrettjonesgoogle's package generation work.
  * It refactors the `API` class to maintain some separation between the source proto files. (I want this because I eventually may want some per-proto functionality.)
  * `API` and `Proto` are now frozen dataclasses. In order to support this, there are `build` class methods on both.
  * A new `Naming` class is added, and naming-based things are moved there (from `API`).
  * Updates documentation to be current.
  * Officially supports Python 3.7.


api_factory/cli/generate.py
```diff
@@ -40,7 +42,12 @@ def generate(
 
     # Translate into a protobuf CodeGeneratorResponse;
     # if there are issues, error out appropriately.
```
pongad: These comments seem more related to just the last line in this "paragraph" right?

lukesneeringer:
The wrapper classes generally have "their name" and nothing else. This has _a bunch_ of names and variants.

api_factory/schema/api.py
```diff
@@ -219,37 +281,34 @@ def _get_methods(self, methods: List[descriptor_pb2.MethodDescriptorProto],
...
-                input=self.messages[method_pb.input_type.lstrip('.')],
-                lro_metadata=self.messages.get(types.lro_metadata_type, None),
-                lro_payload=self.messages.get(types.lro_return_type, None),
-                method_pb=method_pb,
+        for meth_pb, i in zip(methods, range(0, sys.maxsize)):
+            types = meth_pb.options.Extensions[operations_pb2.operation_types]
+            answer[meth_pb.name] = wrappers.Method(
+                input=self.all_messages[meth_pb.input_type.lstrip('.')],
+                lro_metadata=self.all_messages.get(types.metadata, None),
+                lro_payload=self.all_messages.get(types.response, None),
```
landrito: This will cause a silent failure if the payload message is not part of the import tree. Since it's a breaking for the generated client to fix the lro method, would we want to consider adding something to ensure that if a non empty payload type string is provided, a message type must be found?

docs/api-configuration.rst
```diff
@@ -37,55 +40,67 @@ When specifying an annotation, your proto will need to import the file
...
 The most important piece of information this plugin requires is information
-about the client library itself: what should it be called, how is it licensed,
-and so on.
+about the client library itself: what should it be called, what is its proper
+namespace, and so on.
 
-This is rolled up into a strcuture spelled ``Client``, and the annotation
-is defined in `google/api/experimental/client.proto`_.
+This is rolled up into a strcuture spelled ``Metadata``, and the annotation
+is defined in `google/api/metadata.proto`_.
```
landrito: I think this sentence needs rewording.

I am mainly unsure about using the word `spelled`.

api_factory/schema/api.py
```diff
@@ -258,34 +317,34 @@ def _load_descriptor(self, message_pb: descriptor_pb2.DescriptorProto,
         fields = self._get_fields(
             message_pb.field,
             address=nested_addr,
-            info=info.get(2, {}),
+            path=path + (2,),
         )
         fields.update(self._get_fields(
```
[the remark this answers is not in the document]
lukesneeringer: Protocol buffers requires them to be distinct; the separation within proto itself is a (rather annoying) detail.

api_factory/schema/api.py
```diff
@@ -258,34 +317,34 @@ def _load_descriptor(self, message_pb: descriptor_pb2.DescriptorProto,
...
         )
 
         # Load all nested items.
         self._load_children(message_pb.nested_type, address=nested_addr,
-                            loader=self._load_descriptor, info=info.get(3, {}))
+                            loader=self._load_message, path=path + (3,))
         self._load_children(message_pb.enum_type, address=nested_addr,
-                            loader=self._load_enum, info=info.get(4, {}))
+                            loader=self._load_enum, path=path + (4,))
         # self._load_children(message.oneof_decl, loader=self._load_field,
```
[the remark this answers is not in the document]
lukesneeringer: Just the oneof aspect itself. This is not a change in this PR; it was not implemented before, and still is not now.

api_factory/schema/api.py
```diff
@@ -43,116 +76,149 @@ class API:
...
-        #                   address=address, info=comments_by_path.get(7, {}))
-
-        # Merge any client directives with what we have going so far.
-        self.client.MergeFrom(fdp.options.Extensions[client_pb2.client])
-
-    def _load_children(self, children: list, loader: Callable,
-                       address: metadata.Address, info: dict) -> None:
-        """Load arbitrary children from a Descriptor.
+
+    def _load_children(self, children: Sequence, loader: Callable, *,
```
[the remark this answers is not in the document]
lukesneeringer: We can, but it is a lot of work, since this is essentially a dispatch function. It is also a private method, so I am not particularly concerned.

lukesneeringer:
> There are a lot of frozen dataclasses that are using the @property annotation rather than the @cached_property annotation. What is your threshold for using @cached_property?

I use `@cached_property` if something non-trivial is being computed, or in cases where we return instances and it matters that the instances be the same object rather than an equivalent copy.

api_factory/schema/api.py
lukesneeringer: I agree with this, but it is not a change from the status quo. Can I fix that in a follow-up? (I have another PR I need to base off this one.)

api_factory/schema/naming.py
```diff
@@ -0,0 +1,182 @@
...
+        # because protoc is guaranteed to give us valid package names.
+        pattern = r'^((?P<namespace>[a-z0-9_.]+)\.)?(?P<name>[a-z0-9_]+)'
+
+        # Only require the version portion of the regex if the version is
+        # present.
+        #
+        # This code may look counter-intuitive (why not use ? to make it
+        # optional), but the engine's greediness routine will decide that
+        # the version is the name, which is not what we want.
+        version = r'\.(?P<version>v[0-9]+(p[0-9]+)?((alpha|beta|test)[0-9])*)'
```
landrito: This part of the regex seems wrong although I might be understanding it wrong: `((alpha|beta|test)[0-9])*)`

Shouldn't the star be scoped to the [0-9]?

api_factory/schema/wrappers.py
```diff
@@ -105,14 +105,17 @@ def __getattr__(self, name):
...
 
     @property
-    def overloads(self):
-        """Return the overloads defined for this method."""
-        return self.method_pb.options.Extensions[overload_pb2.overloads]
+    def field_headers(self) -> Sequence[str]:
+        """Return the field headers defined for this method."""
+        http = self.options.Extensions[annotations_pb2.http]
+        if http.get:
+            return tuple(re.findall(r'\{([a-z][\w\d_.]+)=', http.get))
```
landrito: This RE is capturing the initial brace and the ending equals but does not strip them. I think this was not caught because this method is being implicitly tested through the `has_field_headers` test but not explicitly tested.

api_factory/templates/$namespace/$name_$version/$service/transports/http.py.j2
```diff
@@ -88,7 +88,12 @@ class {{ service.name }}HttpTransport({{ service.name }}Transport):
 
         # Send the request.
         response = self._session.post(
-            f'https://{self.SERVICE_ADDRESS}/$rpc/{{ '.'.join(method.meta.address.package) }}.{{ service.name }}/{{ method.name }}',
+            'https://{host}/$rpc/{package}.{service}/{method}'.format(
```
landrito: Do we plan on making this interface similar to the grpc transport interface in which this transport also returns callables? I have a feeling this inconsistency will bite us in the future.

docs/api-configuration.rst
```diff
@@ -15,8 +15,11 @@ published protocol buffers.
 This plugin *will* successfully publish a library on a valid protobuf API
 even without any additional information set, but may require some
 post-processing work by a human in this case before the resulting client
-library will install or work. (Look for values enclosed by ``<<<`` and
-``>>>`` to quickly spot these.)
+library will work.
+
+Look for values enclosed by ``<<<`` and ``>>>`` to quickly spot these.
+As of this writing, these are the ``SERVICE_ADDRESS`` and ``OAUTH_SCOPES``
```
landrito: I think `SERVICE_ADDRESS` is called `HOSTNAME` now. Also check oauth scopes name.

tests/unit/schema/test_api.py
```diff
@@ -12,355 +12,377 @@
...
 
         # The services should come third.
-        _, args, kwargs = lc.mock_calls[2]
+        _, args, _ = lc.mock_calls[2]
         assert args[0][0] == sentinel_service
-        assert kwargs['loader'] == api._load_service
+        assert args[1] == pb._load_service
+
+
+def test_not_taget_file():
```
landrito: Spelling

api_factory/schema/api.py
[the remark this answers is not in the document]
landrito: Ack.

api_factory/schema/api.py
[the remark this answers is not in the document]
landrito: Ack.

api_factory/schema/api.py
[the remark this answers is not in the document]
landrito: Ack.

api_factory/schema/api.py
landrito: Lets fix this in a followup. Also, lets make sure that the issue is tracked.

api_factory/schema/naming.py
lukesneeringer: It is correct, because `v1` is a valid name.
This regex would fail `alpha10`, though, which is an error (albeit likely not a very serious one).

api_factory/schema/wrappers.py
lukesneeringer: The capture group does not include the brace.

api_factory/templates/$namespace/$name_$version/$service/transports/http.py.j2
lukesneeringer: This is a good point. There is no "HTTP stub" equivalent in this case, but returning a callable that can be called with the request argument has merit.

This is a significant change, so I will open an issue.

docs/api-configuration.rst
lukesneeringer: Nope, it is still `SERVICE_ADDRESS` in the client library.

api_factory/schema/naming.py
lukesneeringer: Nevermind; looking again, you are right and I am wrong.

api_factory/schema/wrappers.py
landrito: Ack.

lukesneeringer: To be clear, because I understand why this is non-obvious:

```
    Return a list of all non-overlapping matches in the string.
    
    If one or more capturing groups are present in the pattern, return
    a list of groups; this will be a list of tuples if the pattern
    has more than one group.
```

api_factory/templates/$namespace/$name_$version/$service/transports/http.py.j2
lukesneeringer: Filed #7.

api_factory/schema/naming.py
```diff
@@ -0,0 +1,182 @@
...
+        # name validity (e.g. avoiding .. or segments starting with numbers)
+        # because protoc is guaranteed to give us valid package names.
+        pattern = r'^((?P<namespace>[a-z0-9_.]+)\.)?(?P<name>[a-z0-9_]+)'
+
+        # Only require the version portion of the regex if the version is
+        # present.
+        #
+        # This code may look counter-intuitive (why not use ? to make it
+        # optional), but the engine's greediness routine will decide that
+        # the version is the name, which is not what we want.
```
alexander-fenster: Note: the engine's greediness can be limited by appending `?` to `+` or `*`. If the namespace pattern is changed to `(?P<namespace>[a-z0-9_.]+?)` it will never eat more than it's supposed to.

PR #10, opened 2018-07-24
Flattening implementation

### LINKED ISSUE #9: LRO Silent Failure
Currently, if the LRO annotation is absent but `google.longrunning.Operation` is the response type, it will generate a non-working library. This should become an error.

lukesneeringer:
This adds a flattening implementation. I am posting the PR now before tests so it can be reviewed in parallel, but the system (correctly) will prevent merge until tests are complete.

This does the following:

  * It adds a `MethodSignature` and `MethodSignatures` wrapper object to reason about the `google.api.method_signature` annotation.
  * It refactors the LRO implementation slightly to use an `OperationType`; this means templates have far fewer "if LRO" forks. (Fixes #9.)
  * It fixes the routine headers implementation, which was broken (the joys of pre-alpha).
  * It creates the names `python_ident` and `sphinx_ident` for use in templates.
      * These replace the use of `{{ thing.python_module }}.{{ thing.name }}` and the manual use of the `~.` prefix for sphinx.
      * This is useful because sometimes these things are primitives, so this avoids template `if` logic.


api_factory/templates/$namespace/$name_$version/$service/client.py.j2
```diff
@@ -1,23 +1,17 @@
...
-{%- endif %}
 from google.api_core import retry
 from google.auth import credentials
 
-{% for package, pb2_module in service.pb2_modules -%}
-from {{ package }} import {{ pb2_module }}
+{% for package, python_module in service.python_modules -%}
+from {{ package }} import {{ python_module }}
 {% endfor %}
+from ..utils import dispatch
```
landrito: Any plans on moving this to api_core?

api_factory/templates/$namespace/$name_$version/$service/transports/grpc.py.j2
```diff
@@ -103,8 +103,8 @@ class {{ service.name }}GrpcTransport({{ service.name }}Transport):
         if '{{ method.name|snake_case }}' not in self._stubs:
```
landrito: I didn't realize we were doing this. What case does this cover?

api_factory/templates/$namespace/$name_$version/$service/client.py.j2
lukesneeringer: Yes. However, since `api_core` is GA, I want to let things go through some testing before shifting them there. Ultimately, I do not want a `utils.py` to be generated, however.

api_factory/templates/$namespace/$name_$version/$service/transports/grpc.py.j2
lukesneeringer: It just covers not creating them all at instance instantiation. I wrote this before getting Thea's feedback in gapic-generator, and should revisit it.

api_factory/templates/$namespace/$name_$version/$service/client.py.j2
landrito: Ack.

api_factory/templates/$namespace/$name_$version/$service/transports/grpc.py.j2
landrito: Ack.

PR #12, opened 2018-07-25
Remove sending len to Jinja.

lukesneeringer:
The `length` filter exists, and we are no longer using `len` anyway.


landrito:
Where was `len` being used in the templates. I see nothing being removed.

PR #15, opened 2018-07-25
Refactor the wrap filter.

lukesneeringer:
This replaces the `initial_width` and `subsequent_indent` keyword arguments with a single `offset` argument. The `offset` can be a int or a two-tuple of ints (first line and subsequent lines, respectively), and the behavior is the same.

This does remove the ability to have a `subsequent_indent` which is anything other than spaces, which did not end up being used in real life (at least not yet). If it is needed, the `subsequent_indent` stand-alone filter could be resurrected (see #14).

This change markedly improves readability of templates where `wrap` is being used. Similarly, the syntactic change from "initial width" to "initial offset" ends up being far more intuitive (I had used "initial width" incorrectly about half the time and had to debug it).

### DIFF
--- a/api_factory/templates/$namespace/$name_$version/$service/client.py.j2
+++ b/api_factory/templates/$namespace/$name_$version/$service/client.py.j2
@@ -17,7 +17,7 @@ from .transports import {{ service.name }}Transport
 
 
 class {{ service.name }}:
-    """{{ service.meta.doc|wrap(width=72, offset=(7, 4)) }}
+    """{{ service.meta.doc|wrap(width=72, offset=7, indent=4) }}
     """
     def __init__(self, *,
             credentials: credentials.Credentials = None,
@@ -57,12 +57,12 @@ class {{ service.name }}:
             timeout: float = None,
             metadata: Sequence[Tuple[str, str]] = (),
             ) -> {{ method.output.python_ident }}:
-        """{{ method.meta.doc|wrap(width=72, offset=(11, 8)) }}
+        """{{ method.meta.doc|wrap(width=72, offset=11, indent=8) }}
 
         Args:
             request ({{ method.input.sphinx_ident }}):
-                The request object.{{ ' ' }}
-                {{- method.input.meta.doc|wrap(width=72, offset=(36, 16)) }}
+                The request object.{{ ' ' -}}
+                {{ method.input.meta.doc|wrap(width=72, offset=36, indent=16) }}
             retry (~.retry.Retry): Designation of what errors, if any,
                 should be retried.
             timeout (float): The timeout for this request.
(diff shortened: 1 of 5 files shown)


lukesneeringer:
This is useful, thanks. I think these are two separate issues:

 ### The int/tuple[int, int] overload

I went back and forth on whether to do this. I liked it because `offset=8` is so clean, and I thought it was what a naive person would try first. However, I would be happy to remove this (it would be fine to just say `offset=(8, 8)` in templates).

It is documented [here](https://github.com/googleapis/gapic-generator-python/pull/15/files#diff-98ae6debeaf2aa693d8d4ce5e8f5a940R33), incidentally, but if you failed to find it, then others will also. And arguably using a tuple all the time would be a better way to present to a naive reader, since an `(X, X)` tuple will cause the reader to wonder why it is a tuple and go find out.

Thoughts?

### The behavioral difference between the first and subsequent lines.

You are completely correct on the implementation, and I have the same pause.

However, if I think about the function as an _expression of user intent_, this construction makes a lot of sense. Basically what this means is, "The first line is offset X spaces, and all remaining lines are offset Y spaces." The differences in implementation actually complement this, because of how a template author would want the template to look. You _want_ the visual indentation on the first line and there is obviously no way to provide it on the subsequent lines.

I also found that when I changed to this implementation, I caught several bugs, and almost all of them were forgetting that the then-`initial_width` argument was needed. (You can go look at the template diffs to find them.) Making this an expression of intent rather than an expression of implementation makes it feel more intuitive to me.

What do you think? Challenges are welcome if I have not convinced you.

landrito:
I think the method does make sense from that _user intent_ viewpoint but I think the mental path to get to that view of the method is hard to get to. Offset takes on two meanings, current position, and subsequent indents and getting to the understanding of those requires added work of reading two paragraphs in documentation. Reading two paragraphs in documentation is non trivial for methods used in templates since a user unfamiliar with the codebase would need to figure out where the method is defined to find the documentation. This is made harder since these filters are magically in the templates namespace. I assert that the params for the filter methods must be as self documenting as possible.

Personally, I think changing wrap to be would be most understandable (albeit too long):
`def wrap(width: int = 79, current_position: int = 0, subsequent_indent: int =0) => string`

and make another method (which I can't figure out a good name for right now) that covers the situation where `current_position` and `subsequent_indent` are equal.

Furthermore, In a perfect world we could extract the position out of some metadata from the template rather than counting the prior characters (but my thirty minutes of scouring the jinja2 couldn't find anything like that).

lukesneeringer:
Thank you for pushing back on this. I need to think about it more. :-)

> Furthermore, In a perfect world we could extract the position out of some metadata from the template rather than counting the prior characters (but my thirty minutes of scouring the jinja2 couldn't find anything like that).

I mused over this idea but I do not want to do it. It fails an edge case that I do actually use:

```
Call the {{ method.name.replace('_', ' ').wrap... }} method over gRPC.
```

Clever position parsing, even if possible, would do the wrong thing in this case, since what I actually want is an _even shorter_ line because of the presence of a tail.

lukesneeringer:
> Personally, I think changing wrap to be would be most understandable (albeit too long):
> ```
> def wrap(width: int = 79, current_position: int = 0, subsequent_indent: int =0) => string

lukesneeringer:
@landrito and I discussed in chat and we are going to ask a couple folks unfamiliar with the codebase to look at the template diffs to help us understand what is intuitive.

lukesneeringer:
Thanks to @landrito (and others), we identified a better interface. :-)

PR #16, opened 2018-07-25
Add simple whitespace post-processing.

lukesneeringer:
This commit adds very basic whitespace formatting to nullify certain whitespace issues that are sufficiently difficult to get right in Jinja (particularly when `if` and `for` are involved).

This is not meant as a replacement for `yapf`. It is completely reasonable to want to run the resulting code through `yapf`, but I do not want to force this on all template authors. This commit is more humble and just applies very specific fixes that are more or less universally adopted and sufficiently difficult to do correctly in templates in all situations.

### DIFF
--- a/api_factory/generator/formatter.py
+++ b/api_factory/generator/formatter.py
@@ -32,10 +32,10 @@ def fix_whitespace(code: str) -> str:
     code = re.sub(r'[ ]+\n', '\n', code)
 
     # Ensure at most two blank lines before top level definitions.
-    code = re.sub(r'\s*\n\n\n(class|def|@)', r'\n\n\n\1', code)
+    code = re.sub(r'\s+\n\n\n(class|def|@)', r'\n\n\n\1', code)
 
     # Ensure at most one line before nested definitions.
-    code = re.sub(r'\s*\n\n(    )+(class|def|@)', r'\n\n\1\2', code)
+    code = re.sub(r'\s+\n\n(    )+(class|def|@)', r'\n\n\1\2', code)
 
     # All files shall end in one and exactly one line break.
     return f'{code.rstrip()}\n'
--- a/api_factory/generator/formatter.py
+++ b/api_factory/generator/formatter.py
@@ -1,0 +1,41 @@
+# Copyright 2018 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


api_factory/generator/formatter.py
```diff
@@ -0,0 +1,41 @@
...
+        code (str): A string of code to be formatted.
+
+    Returns
+        str: Formatted code.
+    """
+    # Remove trailing whitespace from any line.
+    code = re.sub(r'[ ]+\n', '\n', code)
+
+    # Ensure at most two blank lines before top level definitions.
+    code = re.sub(r'\s+\n\n\n(class|def|@)', r'\n\n\n\1', code)
```
landrito: How bothersome is it to maintain these things at the template level? I think it is good practice to have the templates be as clean as possible.

Maybe we can introduce a logger that emits a developer warning if any of these substitutions are called and have PRs fail on these developer warnings.

lukesneeringer: In general, formatting at the template level is easy, but I have found a couple issues where _this one thing_ is hard.

Basically the problem comes down to `{% for %}` and `{% if %}`. You can use the `{%-` thing in Jinja to do line collapses and such, but it becomes very tricky to do them where a block may or may not be present, or will be present multiple times (which tends to end up with first/last idiosyncrasies).

I have gotten the templates pretty close but there tends to be cases where there are very slight errors in some cases, which these fix.

However, I am also _very_ good with Jinja (having used it for many non-trivial projects over a decade); for a naive template writer, this would be the thing that would cause probably more pain than every other formatting issue combined, and this eliminates it.

Essentially, I agree with you in principle. This case is the exception that proves the rule.

lukesneeringer: For clarity, because I did not explicitly call this out:

> Maybe we can introduce a logger that emits a developer warning if any of these substitutions are called and have PRs fail on these developer warnings.

No, because there are literally a couple cases where they are not avoidable at all, which is why I added this.

landrito: Ah gotcha, if there are things that are more pain that it's worth as well as being impossible this method is welcome!

landrito: We should add a template style guide to the docs and enumerate these specific places where enforcing python style through the templates is very hard/impossible as exceptions.

PR #27, opened 2018-09-06
Be tolerant of no version.

### LINKED ISSUE #26: kiosk/__init__.py: Tried to write the same file twice.
Running in [kiosk/protos](https://github.com/googleapis/kiosk/tree/master/protos) (after having run `COMPILE-PROTOS.sh` in the parent directory), I get this message:
```
$ protoc kiosk.proto --proto_path=./api-common-protos --proto_path=. --pyclient_out=.
kiosk/__init__.py: Tried to write the same file twice.
```
This is with no configuration options, so I'm not surprised that it failed, but was surprised by the type of problem. Are there error messages that I can watch for? (`protoc` plugins can write errors to stdout)

lukesneeringer:
Correct an issue where the generator fails outright with an unfriendly error if the protos do not have a version (due to a collision between `$name/__init__.py` and `$name_$version/__init__.py`).

Fixes #26.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,10 @@ workflows:
     jobs:
       - unit-3.6
       - unit-3.7
+      - showcase:
+          requires:
+            - unit-3.6
+            - unit-3.7
       - docs
 jobs:
   unit-3.6:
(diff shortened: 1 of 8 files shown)


landrito:
I'm having trouble connecting the dots from the changes of the tests to how the issue is fixed. Can you make a more explicit test?

lukesneeringer:
Thanks for keeping me honest, @landrito! Done.

PR #31, opened 2018-09-16
Make a minor adjustment to the whitespace stripper.

lukesneeringer:
This will now clear whitespace even if the intermediate lines themselves have whitespace, and recognizes a comment as a valid start to a new block.


lukesneeringer:
The docs failure is unrelated and dodged by #32.

PR #33, opened 2018-09-17
Directory restructure.

lukesneeringer:
This shifts the directory structure to more closely match the proto structure, and will be the baseline for adding in proto types.


tests/system/conftest.py
```diff
@@ -16,7 +16,7 @@
 
 from google.auth.credentials import AnonymousCredentials
 from google.showcase import Showcase
-from google.showcase_v1alpha1.showcase.transports.grpc import (
+from google.showcase.v1alpha1.services.showcase.transports.grpc import (
```
lukesneeringer: @landrito This is the "payload" test that shows the underlying change. The reason for adding `services` is to disambiguate from the upcoming `types`, which will have proto messages and enums. (After a lot of debating with myself, I want to keep them separate to keep the files more readable. I would love to hear a dissenting opinion if you disagree.)

(The case where someone needs to import this entireeeeeee thing is pretty rare, but I am still considering whether I can shorten this, because it's long.)

lukesneeringer:
The docs are broken because #32 needs to be merged before this can be.

tests/system/conftest.py
landrito: This directory structure seems a lot like the `protos` directory structure in the current generator. I remember you saying that the `protos` directory could have been done better. Would you be able to provide some context about how that strategy and it's fallbacks and in comparison to this strategy. I just want to ensure that we don't make the same mistakes.

lukesneeringer: Honestly, I would consider this to be still an iteration, and not necessarily a statement about "this will be the exact final structure".

Yesterday when I wrote this, I played with a few ideas:

### Each proto becomes a Python module.

This is how protoc works. In this case I would take the package straight-up (unless annotations override it), and the client classes would live together with the message types. I ended up not liking this very much because the files were becoming very long, and also the client classes would ideally be on the bottom (so the messages can come first and be in scope for type annotations), which is not great for users because it is the first thing they need.

Another problem with this is that the transports really should be multiple files (otherwise the file gets _even longer_, and the gRPC import is less clean), and there was no good approach to that.

### Bifurcate below version (or name).

This is what I am currently iterating on. Essentially the idea is that proto gives you this structure:

    {namespace}.{name}.{version}.{subpackages}

The Python generator would bifurcate after version:

    {namespace}.{name}.{version}.services.{subpackages}.{iterate by service}
    {namespace}.{name}.{version}.types.{subpackages}.{iterate by proto}

This is still _way_ less intrusive than what we do now. (Note that the subpackages are not yet maintained in this PR; I intend to add that.)

### Status quo ante.

Also a possibility.

------

My feeling at the moment is that I am not 100% confident that this is the blessed, right way to do it. But I am confident that it removes complexity as opposed to what was there, and it is the right starting point against which to start adding the message types.

In other words -- _this_ PR might end up "making the same mistakes", but the only way for me to find that out is to tinker more with the steps that come next.

-------

One more thing -- my testing strategy to avoid the same mistakes is to hit four cases:

  * Language ("normal")
  * Ads (subpackages)
  * The grafeas folks (an entirely different package)
  * Koisk (unversioned)

I think if I get all of those, I should be in good shape.

landrito: > This is still way less intrusive than what we do now. 

Correct me if I'm wrong but this would also solve issues where the current types.py file _could_ have naming collisions.

Slightly orthogonal, how do we plan on getting the pb2.py files into the correct locations? The copying situation we have now is really fragile. Maybe there is a way to solve that problem here as well (although I doubt it since the protoc interface very strict)?

lukesneeringer: > Correct me if I'm wrong but this would also solve issues where the current types.py file could have naming collisions.

This PR will not solve that, but what I am tinkering with ultimately should.

> Slightly orthogonal, how do we plan on getting the pb2.py files into the correct locations? The copying situation we have now is really fragile. Maybe there is a way to solve that problem here as well (although I doubt it since the protoc interface very strict)?

My intent is to not have _pb2.py files.

lukesneeringer: Update: @theacodes is requesting a couple reversions to this offline, so going to do those first.

PR #44, opened 2018-10-04
Add sorting of imports.

lukesneeringer:


lukesneeringer:
I sort of decided I was not that worried. No one is harmed if we stop sorting.

PR #56, opened 2018-10-11
Update root package error message.

lukesneeringer:
@timburks This is an improvement, but perhaps not improved enough. Suggestions welcome.


lukesneeringer:
> Maybe providing a list of the conflicting package names would make this error message more helpful.

Done.

PR #57, opened 2018-10-11
Determine naming collisions up front.

lukesneeringer:
This is an attempt to determine and resolve naming collisions logically prior to templates (in other words, so templates do not have to worry about it and they just get the "right thing").

This is the promised follow-up to #51. To be honest, I am a little conflicted about it. The removal of the constant `.context(...)` calls in templates are an obvious improvement, but this feels more brittle.

Tests are not migrated yet; I would like feedback on the approach before updating them.

### DIFF
--- a/gapic/schema/api.py
+++ b/gapic/schema/api.py
@@ -256,7 +256,8 @@ class _ProtoBuilder:
         # for each item as it is loaded.
         self.address = metadata.Address(
             api_naming=naming,
-            collisions=self._get_names(file_descriptor),
+            collisions=self._get_names(file_descriptor)
+                if file_to_generate else frozenset(),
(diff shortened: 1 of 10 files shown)


landrito:
I'm not exactly following how this is more brittle than the prior solution. Can you provide some context?

lukesneeringer:
> I think we would still need some iteration of Address#context() to update python_import. It's non-obvious from the method name what context supposed to do. Perhaps something that indicates its use for resolving naming collisions or even just update_import would be more clear.

I all kinds of dislike the method name `context` but have struggled to come up with anything better. I thought about `bind` and `scope`. I could make it something longer, like `register_collisions`, but honestly even _that_ is unclear.

PR #58, opened 2018-10-16
Determine naming collisions up front. (Take 2.)

lukesneeringer:
This is an attempt to determine and resolve naming collisions logically prior to templates (in other words, so templates do not have to worry about it and they just get the "right thing").

Closes #57.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -5,4 +5,4 @@ ignore =
   E123, E124
   # Line over-indented for visual indent.
   # This works poorly with type annotations in method declarations.
-  E128, E131
+  E126, E128, E131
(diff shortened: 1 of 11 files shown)


lukesneeringer:
@eoogbe @landrito I was still bothered by #57, so I took another crack at it. This approach is significantly cleaner, at the cost of being less efficient (we build up naive proto and service objects and then apply the context to them _post hoc_).

One benefit to this approach is that all existing tests pass perfectly (with full coverage even), although I intend to add some to specifically exercise the `context` methods.

Eva, I would love help thinking of a better name for the `context` method -- ideas are welcome. I have considered `bind`, `scope`, `locals`...thoughts?

PR #64, opened 2018-11-29
Docker

lukesneeringer:
Adds a Dockerfile and infrastructure for building an image, as well as documentation for using said image.

This PR does _not_ provide automated image building on releases, since there is no discrete release concept yet. This is coming next.


docs/getting-started/docker.rst
```diff
@@ -0,0 +1,190 @@
...
+
+    This matters because of how proto imports are resolved. The ``import``
+    statement imports a *file*, relative to a base directory or set of
+    base directories, called the ``proto_path``. This is assumed
+    (and hard-coded) to ``/input/`` in the Docker image, and so any directory
+    structure present in the imports of the proto files must be preserved
+    beneath this for compilation to succeed.
+
+
+Running a Client Library
```
landrito: This the docker docs doesn't seem like the right place for this section. Maybe link out to it somewhere.

lukesneeringer: This is just copied from `local.rst` outright. I agree with shifting it.

lukesneeringer: Good call on this, thanks. I shifted the file and document structure around slightly and made it so that there is not (significant) repeated documentation. This will make it much easier to keep things up to date.

I do like the narrative flow of having verification on the getting started page, but it was subsumed under "usage" in a way that did not make sense. Fixed that.

PR #65, opened 2018-12-07
Automated publication (PyPI, Docker image)

lukesneeringer:
This wires up automated publication when release tags are added.
(Some thrashing is possible to make sure the automation works, since the only real way to test it is to do so live, here.)


.circleci/config.yml
```diff
@@ -103,3 +137,44 @@ jobs:
...
           command: nox -s docs
+  publish_package:
+    docker:
+      - image: python:3.7-slim
+    steps:
+      - checkout
+      - run:
+          name: Decrypt the PyPI key.
+          command: |
+            openssl aes-256-cbc -d \
```
alexander-fenster: Note: default algorithm may change. Details here: https://github.com/googleapis/synthtool/pull/159

.circleci/config.yml
```diff
@@ -103,3 +137,44 @@ jobs:
...
+          name: Publish to PyPI.
+          command: python setup.py sdist upload
+  publish_image:
+    docker:
+      - image: google/cloud-sdk
+    steps:
+      - checkout
+      - run:
+          name: Build Docker image.
+          command: docker build . -t gcr.io/gapic-images/gapic-generator-python:latest
```
alexander-fenster: To run docker inside docker, you need to do something special. Example here: https://github.com/googleapis/artman/blob/a429a2c05238fb65470e9e490ee83cec64bc8a56/.circleci/config.yml#L101

.circleci/config.yml
```diff
@@ -103,3 +137,44 @@ jobs:
...
+      - run:
+          name: Set up authentication to Google Container Registry.
+          command: |
+            echo ${GCLOUD_SERVICE_KEY} > ${GOOGLE_APPLICATION_CREDENTIALS}
+            gcloud auth configure-docker
```
alexander-fenster: This will go inside docker container, something like `docker run -i ... gcloud auth configure-docker`

.circleci/config.yml
lukesneeringer: Acknowledged. Will address in a follow-up.

.circleci/config.yml
lukesneeringer: Done.

.circleci/config.yml
lukesneeringer: Done.

PR #72, opened 2018-12-18
[chore] Run CI Showcase in its own image.

lukesneeringer:


.circleci/config.yml
```diff
@@ -93,6 +93,7 @@ jobs:
   showcase:
     docker:
       - image: python:3.7-slim
+      - image: gcr.io/gapic-showcase/gapic-showcase:0.0.7
```
alexander-fenster: How does it work with two images?

lukesneeringer: The gapic-showcase image has `EXPOSE 7469`, and CircleCI is smart. The first image is where all steps are run, and all subsequent images are just run and their `EXPOSE` statements are sent to the first image's localhost.

lukesneeringer: The documentation on this is [here](https://circleci.com/docs/2.0/executor-types/#using-multiple-docker-images).

PR #83, opened 2018-12-30
[fix] Address bad import against nested messages.

lukesneeringer:
This commit fixes a spurious import that tried to import nested messages directly from module scope.


lukesneeringer:
Also filed #84 to refactor a related piece of this.

PR #88, opened 2019-01-03
[feat] Multiple template sources.

lukesneeringer:
This commit allows for multiple sources for templates, which are searched in order.


lukesneeringer:
Also adds documentation covering both local and Docker-based invocation. No change to `gapic.sh` at this time since I am not sure I want to add the directory mounting aspect to that.

PR #92, opened 2019-02-17
[refactor] New annotations

lukesneeringer:
This updates gapic-generator-python to use the current annotations.


lukesneeringer:
The Showcase errors are expected since _Showcase_ is likely not updated to the new annotations yet.

PR #96, opened 2019-02-23
[fix] Docstrings with generated content should be raw.

lukesneeringer:
This fixes an issue where comments with backslashes can cause a syntax error.


alexander-fenster:
Ah. I have just realized it _is_ covered by tests (that currently fail).

lukesneeringer:
Yep. :-)

PR #103, opened 2019-02-28
[feat] Retries.

lukesneeringer:
This commit adds automatic retries for both idempotent and
non-idempotent requests.

### DIFF
--- a/tests/unit/schema/wrappers/test_method.py
+++ b/tests/unit/schema/wrappers/test_method.py
@@ -69,6 +69,11 @@ def test_method_idempotent_no():
     assert method.idempotent is False
 
 
+def test_method_idempotent_no_http_rule():
+    method = make_method('DoSomething')
+    assert method.idempotent is False
+
+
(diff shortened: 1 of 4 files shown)


tests/unit/schema/wrappers/test_method.py
```diff
@@ -51,10 +51,22 @@ def test_method_field_headers_none():
...
+    assert method.field_headers == ('parent',)
+
+
+def test_method_idempotent_yes():
+    http_rule = http_pb2.HttpRule(get='/v1/{parent=projects/*}/topics')
+    method = make_method('DoSomething', http_rule=http_rule)
+    assert method.idempotent is True
+
+
+def test_method_idempotent_no():
```
alexander-fenster: Can we have a test of `assert method.idempotent is False` for a method with no `HttpRule`?

lukesneeringer: Good idea. Done.

PR #105, opened 2019-02-28
[chore] Bump to 0.7.0, beta.

lukesneeringer:
As best as I can tell, this is feature complete (every "must" in
the spec is fulfilled).

I still want to add more tests and particularly generated unit
tests, but I feel confident bumping this to beta at this point.


lukesneeringer:
Note: Holding off on this until #103 is merged.

Issue #113, opened 2019-03-27
Error in protoc google/cloud/vision/v1/image_annotator.proto


AnupKumarPanwar:
Reference docs : https://gapic-generator-python.readthedocs.io/en/stable/getting-started/local.html

Running the following command to setup the project locally produces error
`protoc google/cloud/vision/v1/image_annotator.proto     --proto_path=../api-common-protos/ --proto_path=.     --python_gapic_out=../dest/`

Output
```
google/longrunning/operations.proto:30:8: Option "php_namespace" unknown.
google/cloud/vision/v1/image_annotator.proto: Import "google/longrunning/operations.proto" was not found or had errors.
google/cloud/vision/v1/image_annotator.proto:63:16: "google.longrunning.Operation" is not defined.
```

lukesneeringer:
Hi @AnupKumarPanwar,
Thanks for trying out gapic-generator-python.

I believe `php_namespace` was an option introduced into protocol buffers relatively recently, and may not play nicely with older versions of `protoc`. Can you check and make sure you are running a recent version? I think you need 3.6.0 or higher, but I am not entirely certain. You can get a new release [here](https://github.com/protocolbuffers/protobuf/releases).

Note that if you upgrade, you may need to manually copy the `google/protobuf/` protos to an include location on your disk (see the README in their tarballs).

lukesneeringer:
Hi @AnupKumarPanwar,
I am closing this as I believe it is not something we can act upon. Please feel free to reopen if I have missed something. :-)

PR #117, opened 2019-04-01
[chore] Update to protoc 3.7.1

lukesneeringer:


AnupKumarPanwar:
It will fix #113 right?

lukesneeringer:
I think #113 is an issue on your local development environment. protoc 3.6.x (which is what this is currently on) already supports the annotation that #113 is failing on.

PR #121, opened 2019-05-19
[feat] Add automatic pagination support.

lukesneeringer:
This commit adds support for automatic pagination. Paginated
methods now return a "Pager" class that wraps the underlying
response, but adds an `__iter__` method which returns a generator
that yields results until no more results remain.

Spec reference: [AIP-4233](https://aip.dev/4233)


lukesneeringer:
This PR also ships a refactor for wrapping arbitary Python types, bringing more consistency to how things like `OperationType` are handled, and greatly simplifying the relationship between "output type" (what the transport returns) and "client output type" (what the GAPIC client returns).

gapic/schema/wrappers.py
```diff
@@ -489,17 +528,34 @@ def idempotent(self) -> bool:
...
+                           (self.input, str, 'page_token'),
+                           (self.output, str, 'next_page_token')):
+            field = page_field[0].fields.get(page_field[2], None)
+            if not field or field.type != page_field[1]:
+                return None
+
+        # Return the first repeated field.
+        for field in self.output.fields.values():
+            if field.repeated and field.message:
+                return field
```
alexander-fenster: will this method implicitly return `None` if there are no `repeated` fields in the output message? can we return `None` explicitly to make it clear?

lukesneeringer: Yes, and sure. Will do that before merging.

lukesneeringer: Done. Thanks.

PR #126, opened 2019-05-23
[feat] Add a CLI option to set the Python module

lukesneeringer:
This allows us to translate between the actual proto package and the desired Python module when automatic inference does the wrong thing (e.g. think `google.spanner.v1` -> `google.cloud.spanner_v1`).

### DIFF
--- a/gapic/generator/options.py
+++ b/gapic/generator/options.py
@@ -26,7 +26,9 @@ class Options:
     on unrecognized arguments (essentially, we throw them away, but we do
     warn if it looks like it was meant for us).
     """
-    templates: Tuple[str]
+    templates: Tuple[str] = dataclasses.field(default=('DEFAULT',))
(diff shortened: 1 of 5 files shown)


gapic/generator/options.py
```diff
@@ -57,26 +59,28 @@ def build(cls, opt_string: str) -> 'Options':
             # Set the option.
             # Just assume everything is a list at this point, and the
             # final instantiation step can de-list-ify where appropriate.
-            opts.setdefault(opt, [])
-            opts[opt].append(value)
+            opts.setdefault(opt[13:], [])
+            opts[opt[13:]].append(value)
```
alexander-fenster: I'm struggling to understand why it's `opt[13:]`. A comment or a named constant would've made sense here.

lukesneeringer: `len('python-gapic-')` is equal to 13.
I will change it to that.

PR #127, opened 2019-05-28
[feat] Improve enums.

lukesneeringer:
This uses proto-plus enums, which add enum support to the marshal and cause enums (rather than ints) to be returned, and add support for accepting strings.


lukesneeringer:
This is thoroughly tested in proto plus, but I have no good way to write an integration test here due to the lack of enums in Showcase. Filed https://github.com/googleapis/gapic-showcase/issues/142 on Showcase to address this.

PR #129, opened 2019-05-28
[fix] Do not silently drop `--python-gapic-` options.

lukesneeringer:
This is a follow-up to #126, where I added a CLI option, but neglected to update the Docker entrypoint script to actually pass it on to the generator, making the option useless when using Docker (the common case for now).

This updates the script to be more robust, and check the prefix rather than an index of every known option.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -25,10 +25,10 @@ while [ -n "$1" ]; do
     * )
       # If this switch begins with "--python-gapic-" or "--gapic-", then it is
       # meant for us.
-      if [ "${1:0:15}" == "--python-gapic-" ]; then
+      if [[ $1 == --python-gapic-* ]]; then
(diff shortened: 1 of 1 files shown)


docker-entrypoint.sh
```diff
@@ -18,17 +18,23 @@ PLUGIN_OPTIONS=""
...
     -- )
       shift
       break
       ;;
     * )
-      # Ignore anything we do not recognize.
-      shift
+      # If this switch begins with "--python-gapic-" or "--gapic-", then it is
+      # meant for us.
+      if [ "${1:0:15}" == "--python-gapic-" ]; then
```
alexander-fenster: Since we use `#!/bin/bash` and not `#!/bin/sh`, let's just do 
```sh
if [[ $var == --python-gapic-* ]]; then
```
here and below (note double square brackets `[[` `]]` for conditional expression evaluation).

lukesneeringer: Today I learned. Thanks.

PR #137, opened 2019-07-03
Initial commit for generating an integrated sample

software-dov:
Adds basic functionality and tests for end-to-end sample generation.

Includes tests for comprehensive sample generation

Conduct translation from special variable '$resp' to 'response' in
template rendering

Includes macros to set up method invocation, adds code to include
license and copyright information,

Includes minor tweaks to various names, tests, and macros.

Note: sample generation is still ignorant of method and message
      signatures, types, attributes, and so forth.
      TODOs mark these areas in the code.
      More full-featured semantic analysis will be conducted
      in subsequent PRs.

### DIFF
--- a/gapic/generator/generator.py
+++ b/gapic/generator/generator.py
@@ -72,13 +72,13 @@ class Generator:
         output_files: Dict[str, CodeGeneratorResponse.File] = OrderedDict()
 
         # TODO: handle sample_templates specially, generate samples.
-        sample_templates, other_templates = utils.partition(
+        sample_templates, client_templates = utils.partition(
             self._env.loader.list_templates(),
             lambda fname: os.path.basename(fname) == samplegen.TEMPLATE_NAME)
 
         # Iterate over each template and add the appropriate output files
         # based on that template.
-        for template_name in other_templates:
+        for template_name in client_templates:
             # Sanity check: Skip "private" templates.
             filename = template_name.split('/')[-1]
             if filename.startswith('_') and filename != '__init__.py.j2':
--- a/gapic/generator/generator.py
+++ b/gapic/generator/generator.py
@@ -55,7 +55,7 @@ class Generator:
         self._env.filters['snake_case'] = utils.to_snake_case
         self._env.filters['sort_lines'] = utils.sort_lines
         self._env.filters['wrap'] = utils.wrap
-        self._env.filters['coerce_variable_name'] = samplegen.coerce_variable_name
+        self._env.filters['coerce_response_name'] = samplegen.coerce_response_name
 
     def get_response(self, api_schema: api.API) -> CodeGeneratorResponse:
         """Return a :class:`~.CodeGeneratorResponse` for this library.
--- a/gapic/generator/generator.py
+++ b/gapic/generator/generator.py
@@ -55,7 +55,7 @@ class Generator:
         self._env.filters['snake_case'] = utils.to_snake_case
         self._env.filters['sort_lines'] = utils.sort_lines
         self._env.filters['wrap'] = utils.wrap
-        self._env.filters['resp_to_response'] = samplegen.resp_to_response
+        self._env.filters['coerce_variable_name'] = samplegen.coerce_variable_name
 
     def get_response(self, api_schema: api.API) -> CodeGeneratorResponse:
         """Return a :class:`~.CodeGeneratorResponse` for this library.
@@ -74,8 +74,8 @@ class Generator:
 
         # TODO: handle sample_templates specially, generate samples.
         sample_templates, client_templates = utils.partition(
-            self._env.loader.list_templates(),
-            lambda fname: os.path.basename(fname) == samplegen.TEMPLATE_NAME)
+            lambda fname: os.path.basename(fname) == samplegen.TEMPLATE_NAME,
+            self._env.loader.list_templates())
 
         # Iterate over each template and add the appropriate output files
         # based on that template.
--- a/gapic/generator/generator.py
+++ b/gapic/generator/generator.py
@@ -70,9 +70,14 @@ class Generator:
         """
         output_files: Dict[str, CodeGeneratorResponse.File] = OrderedDict()
 
+        # TODO: handle sample_templates specially, generate samples.
+        sample_templates, other_templates = utils.partition(
+            self._env.loader.list_templates(),
+            lambda fname: os.path.basename(fname) == "sample.py.j2")
+
         # Iterate over each template and add the appropriate output files
         # based on that template.
-        for template_name in self._env.loader.list_templates():
+        for template_name in other_templates:
             # Sanity check: Skip "private" templates.
             filename = template_name.split('/')[-1]
             if filename.startswith('_') and filename != '__init__.py.j2':
--- a/gapic/generator/generator.py
+++ b/gapic/generator/generator.py
@@ -55,6 +55,7 @@ class Generator:
         self._env.filters['snake_case'] = utils.to_snake_case
         self._env.filters['sort_lines'] = utils.sort_lines
         self._env.filters['wrap'] = utils.wrap
+        self._env.filters['resp_to_response'] = samplegen.resp_to_response
 
     def get_response(self, api_schema: api.API) -> CodeGeneratorResponse:
         """Return a :class:`~.CodeGeneratorResponse` for this library.
--- a/gapic/generator/generator.py
+++ b/gapic/generator/generator.py
@@ -24,6 +24,7 @@ from google.protobuf.compiler.plugin_pb2 import CodeGeneratorResponse
 from gapic import utils
 from gapic.generator import formatter
 from gapic.generator import options
+from gapic.samplegen import samplegen
 from gapic.schema import api
 
 
@@ -73,7 +74,7 @@ class Generator:
         # TODO: handle sample_templates specially, generate samples.
         sample_templates, other_templates = utils.partition(
             self._env.loader.list_templates(),
-            lambda fname: os.path.basename(fname) == "sample.py.j2")
+            lambda fname: os.path.basename(fname) == samplegen.TEMPLATE_NAME)
 
         # Iterate over each template and add the appropriate output files
         # based on that template.
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -251,9 +251,6 @@ class Validator:
             response: list[dict{str:?}]: The structured data representing
                                          the sample's response.
 
-        Returns:
-            bool: True if response is valid.
-
         Raises:
             InvalidStatement: If an unexpected key is found in a statement dict
                               or a statement dict has more than or less than one key.
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -20,6 +20,7 @@ import re
 from gapic.samplegen import utils
 
 from collections import (defaultdict, namedtuple)
+from textwrap import dedent
 from typing import (List, Mapping, Set, Tuple)
 
 # Outstanding issues:
@@ -30,19 +31,22 @@ MIN_SCHEMA_VERSION = (1, 2, 0)
 
 VALID_CONFIG_TYPE = 'com.google.api.codegen.SampleConfigProto'
 
-FILE_HEADER = {"copyrightLines": ["Copyright (C) 2019  Google LLC"],
-               "licenseLines": ["Licensed under the Apache License, Version 2.0 (the \"License\");",
-                                "you may not use this file except in compliance with the License.",
-                                "You may obtain a copy of the License at",
-                                "",
-                                "     http://www.apache.org/licenses/LICENSE-2.0",
-                                "",
-                                "Unless required by applicable law or agreed to in writing, software",
-                                "distributed under the License is distributed on an \"AS IS\" BASIS,",
-                                "WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.",
-                                "See the License for the specific language governing permissions and",
-                                "limitations under the License."
-                                ]}
+FILE_HEADER = {"copyrightLines": "Copyright (C) 2019  Google LLC",
+               "licenseLines": dedent(
+                   """
+                   Licensed under the Apache License, Version 2.0 (the "License");
+                   you may not use this file except in compliance with the License.
+                   You may obtain a copy of the License at
+                      
+                        http://www.apache.org/licenses/LICENSE-2.0
+                      
+                   Unless required by applicable law or agreed to in writing, software
+                   distributed under the License is distributed on an "AS IS" BASIS,
+                   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+                   See the License for the specific language governing permissions and
+                   limitations under the License.
+                   """).strip()
+               }
 
 RESERVED_WORDS = frozenset(
     itertools.chain(keyword.kwlist,
@@ -232,8 +236,8 @@ class Validator:
         if (calling_form in {utils.CallingForm.RequestStreamingClient,
                              utils.CallingForm.RequestStreamingBidi} and
                 len(base_param_to_attrs) > 1):
-            raise InvalidRequestSetup("""There can be at most 1 base request in a
-            sample for a method with client side streaming""")
+            raise InvalidRequestSetup(("There can be at most 1 base request in a sample"
+                                       " for a method with client side streaming"))
 
         return [TransformedRequest(base, body)
                 for base, body in base_param_to_attrs.items()]
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -32,7 +32,8 @@ MIN_SCHEMA_VERSION = (1, 2, 0)
 VALID_CONFIG_TYPE = 'com.google.api.codegen.SampleConfigProto'
 
 # TODO: read in copyright and license from files.
-FILE_HEADER: Dict[str, str] = {}
+FILE_HEADER: Dict[str, str] = {"copyright": "TODO: add a copyright",
+                               "license": "TODO: add a license"}
 
 RESERVED_WORDS = frozenset(
     itertools.chain(keyword.kwlist,
@@ -103,7 +104,7 @@ class InvalidRequestSetup(SampleError):
     pass
 
 
-def coerce_variable_name(s):
+def coerce_response_name(s: str) -> str:
     # In the sample config, the "$resp" keyword is used to refer to the
     # item of interest as received by the corresponding calling form.
     # For a 'regular', i.e. unary, synchronous, non-long-running method,
@@ -156,7 +157,7 @@ class Validator:
            are that "field" maps only to the second part of a dotted variable name.
            Other key/value combinations in the dict are unmodified for the time being.
 
-           Note: gRPC API methods only take one parameter (ignoring client side streaming).
+           Note: gRPC API methods only take one parameter (ignoring client-side streaming).
                  The reason that GAPIC client library API methods may take multiple parameters
                  is a workaround to provide idiomatic protobuf support within python.
                  The different 'bases' are really attributes for the singular request parameter.
@@ -350,6 +351,39 @@ class Validator:
             raise UndefinedVariableReference("Reference to undefined variable: {}"
                                              .format(rval_base))
 
+    def _validate_write_file(self, body):
+        """Validate 'write_file' statements.
+
+        The body of a 'write_file' statement is a two-element dict
+        with known keys: 'filename' and 'contents'.
+        'filename' maps to a list of strings which constitute a format string
+        and variable-based rvalues defining the fields.
+        'contents' maps to a single variable-based rvalue.
+
+        Raises:
+            MismatchedFormatSpecifier: If the filename formatstring is badly formed.
+            UndefinedVariableReference: If any of the formatstring variables
+                                        or the file contents variable are undefined.
+            InvalidStatement: If either 'filename' or 'contents' are absent keys.
+        """
+
+        fname_fmt = body.get("filename")
+        if not fname_fmt:
+            raise InvalidStatement(
+                "Missing key in 'write_file' statement: 'filename'")
+
+        self._validate_format(fname_fmt)
+
+        contents_var = body.get("contents")
+        if not contents_var:
+            raise InvalidStatement(
+                "Missing key in 'write_file' statement: 'contents'")
+
+        base = contents_var.split(".")[0]
+        if base not in self.var_defs_:
+            raise UndefinedVariableReference("Reference to undefined variable: {}"
+                                             .format(base))
+
     def _validate_loop(self, body):
         """Validates loop headers and statement bodies.
 
@@ -438,6 +472,7 @@ class Validator:
         "define": _validate_define,
         "print": _validate_format,
         "comment": _validate_format,
+        "write_file": _validate_write_file,
         "loop": _validate_loop,
     }
 
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -13,15 +13,16 @@
 # limitations under the License.
 
 import itertools
+import jinja2
 import keyword
 import re
-import gapic.samplegen.utils
+
+from gapic.samplegen import utils
 
 from collections import (defaultdict, namedtuple)
-from typing import (List, Mapping, Set)
+from typing import (List, Mapping, Set, Tuple)
 
 # Outstanding issues:
-# * Neither license nor copyright are defined
 # * In real sample configs, many variables are
 #   defined with an _implicit_ $resp variable.
 
@@ -29,13 +30,33 @@ MIN_SCHEMA_VERSION = (1, 2, 0)
 
 VALID_CONFIG_TYPE = 'com.google.api.codegen.SampleConfigProto'
 
-# There are a couple of other names that should be reserved
-# for sample variable assignments, e.g. 'client', but there are diminishing returns:
-# the sample config author isn't a pathological adversary, they're a normal person
-# who may make mistakes occasionally.
-RESERVED_WORDS = frozenset(itertools.chain(keyword.kwlist, dir(__builtins__)))
+FILE_HEADER = {"copyrightLines": ["Copyright (C) 2019  Google LLC"],
+               "licenseLines": ["Licensed under the Apache License, Version 2.0 (the \"License\");",
+                                "you may not use this file except in compliance with the License.",
+                                "You may obtain a copy of the License at",
+                                "",
+                                "     http://www.apache.org/licenses/LICENSE-2.0",
+                                "",
+                                "Unless required by applicable law or agreed to in writing, software",
+                                "distributed under the License is distributed on an \"AS IS\" BASIS,",
+                                "WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.",
+                                "See the License for the specific language governing permissions and",
+                                "limitations under the License."
+                                ]}
 
-TEMPLATE_NAME = 'samplegen_template.j2'
+RESERVED_WORDS = frozenset(
+    itertools.chain(keyword.kwlist,
+                    dir(__builtins__),
+                    {"client",
+                     "f",            # parameter used in file I/O statements
+                     "operation",    # temporary used in LROs
+                     "page",         # used in paginated responses
+                     "page_result",  # used in paginated responses
+                     "response",     # basic 'response'
+                     "stream",       # used in server and bidi streaming
+                     }))
+
+TEMPLATE_NAME = "sample.py.j2"
 
 TransformedRequest = namedtuple("TransformedRequest", ["base", "body"])
 
@@ -48,15 +69,15 @@ class ReservedVariableName(SampleError):
     pass
 
 
-class SchemaVersion(SampleError):
-    pass
-
-
 class RpcMethodNotFound(SampleError):
     pass
 
 
-class InvalidConfigType(SampleError):
+class UnknownService(SampleError):
+    pass
+
+
+class InvalidConfig(SampleError):
     pass
 
 
@@ -113,6 +134,7 @@ class Validator:
     # TODO: this will eventually need the method name and the proto file
     # so that it can do the correct value transformation for enums.
     def validate_and_transform_request(self,
+                                       calling_form: utils.CallingForm,
                                        request: List[Mapping[str, str]]) -> List[TransformedRequest]:
         """Validates and transforms the "request" block from a sample config.
 
@@ -174,6 +196,7 @@ class Validator:
         """
         base_param_to_attrs: Mapping[str,
                                      List[Mapping[str, str]]] = defaultdict(list)
+
         for field_assignment in request:
             field_assignment_copy = dict(field_assignment)
             input_param = field_assignment_copy.get("input_parameter")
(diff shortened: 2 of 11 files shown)


gapic/generator/generator.py
```diff
@@ -70,9 +71,14 @@ def get_response(self, api_schema: api.API) -> CodeGeneratorResponse:
...
 
+        # TODO: handle sample_templates specially, generate samples.
+        sample_templates, other_templates = utils.partition(
+            self._env.loader.list_templates(),
+            lambda fname: os.path.basename(fname) == samplegen.TEMPLATE_NAME)
+
         # Iterate over each template and add the appropriate output files
         # based on that template.
-        for template_name in self._env.loader.list_templates():
+        for template_name in other_templates:
```
hzyi-google: Would it make sense to change `other_templates` to `client_templates` or `library_templates`?

tests/unit/samplegen/test_template.py
```diff
@@ -301,6 +312,226 @@ def test_dispatch_map_loop():
...
+    # The callingForm and callingFormEnum parameters are dummies,
+    # which we can get away with because of duck typing in the template.
+    check_template(
+        '''
+        {% import "feature_fragments.j2" as frags %}
+        {{ frags.renderMethodCall({"rpc": "Categorize", "request": [{"base": "video"}]},
+                                  "client", {"RequestStreamingBidi": "bidi",
+                                             "RequestStreamingClient": "client"}) }}
+       ''',
+        "\nclient.Categorize([video])\n"
```
hzyi-google: Should it be `categorize` here? I _think_ public methods in Python gapics are in `snake_case`.

tests/unit/samplegen/test_template.py
```diff
@@ -301,6 +312,226 @@ def test_dispatch_map_loop():
     )
 
 
+def test_print_input_params():
+    check_template(
+        '''
+        {% import "feature_fragments.j2" as frags %}
+        {{ frags.printInputParams([{"base": "squid",
+                                    "body": [{"field": "mass",
```
hzyi-google: Do you think it would be useful to have an input variable with an underscore to test casing?

hzyi-google:
There are some changes which seem to be code formatting only. Is it possible to have a separate PR dedicated to fix formatting issues?

software-dov:
I'll open an issue for formatting tweaks and make a separate PR to address the issue.

tests/unit/samplegen/test_template.py
software-dov: You're right: both the existing samples from https://github.com/beccasaurus/gapic-docs-samples and the output of the microgenerator use snake case. Will address.

gapic/generator/generator.py
software-dov: I like `client_templates`.

tests/unit/samplegen/test_template.py
software-dov: The `input_parameter` in `test_render_attr_file` has underscores.

tests/unit/samplegen/test_template.py
```diff
@@ -153,9 +153,9 @@ def test_render_print_args():
     check_template(
         '''
         {% import "feature_fragments.j2" as frags %}
-        {{ frags.renderPrint(["Molluscs: %s, %s, %s", "squid", "clam", "whelk"]) }}
+        {{ frags.renderPrint(["$resp %s %s", "$resp.squids", "$resp.clams"]) }}
         ''',
-        '\nprint("Molluscs: {}, {}, {}", squid, clam, whelk)\n'
+        '\nprint("$resp {} {}", response.squids, response.clams)\n'
```
vchudnov-g: nitty nit nit suggestion: Consider having the expected string also be specified using `'''`: this will make it more readable and take care of the `\n`s

tests/unit/samplegen/test_template.py
```diff
@@ -301,6 +312,226 @@ def test_dispatch_map_loop():
...
+        ''',
+        "\nmass, length, color"
+    )
+
+
+def test_render_calling_form_request():
+    check_template(
+        '''
+        {% import "feature_fragments.j2" as frags %}
+        {% set callingFormEnum  = {
```
vchudnov-g: Is there any way you can use `utils.CallingForm` here and below? That way you're checking against the actual enum values being used.

gapic/utils/code.py
```diff
@@ -21,3 +21,25 @@ def empty(content: str) -> bool:
...
+def partition(iterator, predicate=bool):
+    """Partitions an iterable into two lists based on a predicate
+
+    Args:
+        iterator : An iterable on any type.
+        predicate: A callable predicate on a single argument
+                   of whatever type is in iterator.
+
+    Returns:
+        Tuple(List, List): The contents of iterator partitoned into two lists.
```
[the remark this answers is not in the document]
software-dov: Doh! Nice catch.

gapic/generator/generator.py
```diff
@@ -70,9 +70,14 @@ def get_response(self, api_schema: api.API) -> CodeGeneratorResponse:
...
 
+        # TODO: handle sample_templates specially, generate samples.
+        sample_templates, other_templates = utils.partition(
+            self._env.loader.list_templates(),
+            lambda fname: os.path.basename(fname) == "sample.py.j2")
+
         # Iterate over each template and add the appropriate output files
         # based on that template.
-        for template_name in self._env.loader.list_templates():
+        for template_name in other_templates:
```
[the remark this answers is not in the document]
software-dov: Hanzhen suggested `client_templates`, which I think is good, as is `library_templates`. I think that `gapic_templates` is misleading, since from once perspective _all_ the templates are gapic templates, and `non_gapic_templates` is what an old math teacher of mine would have called "true but not useful."

gapic/samplegen/samplegen.py
```diff
@@ -13,29 +13,50 @@
...
 VALID_CONFIG_TYPE = 'com.google.api.codegen.SampleConfigProto'
 
-# There are a couple of other names that should be reserved
-# for sample variable assignments, e.g. 'client', but there are diminishing returns:
-# the sample config author isn't a pathological adversary, they're a normal person
-# who may make mistakes occasionally.
-RESERVED_WORDS = frozenset(itertools.chain(keyword.kwlist, dir(__builtins__)))
-
-TEMPLATE_NAME = 'samplegen_template.j2'
+FILE_HEADER = {"copyrightLines": ["Copyright (C) 2019  Google LLC"],
```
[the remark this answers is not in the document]
software-dov: In this instance it's a pure formality, but if/when the time comes to read these from files the separation makes more sense. It also makes the change for a new year more straightforward.

tests/unit/samplegen/test_integration.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # [START mollusc_classify_sync_core]
+
+    client = mollusca_v1.MolluscClient()
+
+    classify_request = {}
+    # video = "path/to/mollusc/video.mkv"
+    with open(video, "rb") as f:
+        classify_request["video"] = f.read()
+
```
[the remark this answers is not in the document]
software-dov: Agreed, will try fighting with jinja again :package:

tests/unit/samplegen/test_template.py
software-dov: Yeah, bad call on my part. My initial judgement was that using triple-quoted strings for one-liners was less readable, but the alternative is not great.

tests/unit/samplegen/test_samplegen.py
```diff
@@ -405,27 +393,47 @@ def test_validate_request_multiple_arguments():
...
 
 def test_validate_request_reserved_input_param():
     with pytest.raises(samplegen.ReservedVariableName):
+        samplegen.Validator().validate_and_transform_request(utils.CallingForm.Request,
+                                                             [{"field": "mollusc.class",
+                                                               "value": "cephalopoda",
+                                                               "input_parameter": "class"}])
+
+
+def test_single_request_client_streaming():
```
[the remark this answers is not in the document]
software-dov: Not quite, it's because methods with client side streaming can't use the positional/keyword attribute notation when describing the request. These methods need a single 'base' dict that's then wrapped in a list and passed to the client method.

In the general case, request construction needs to be much more sophisticated, but for right now it's a quick sanity check.

gapic/samplegen/samplegen.py
```diff
@@ -206,6 +229,12 @@ def validate_and_transform_request(self,
             field_assignment_copy["field"] = attr
             base_param_to_attrs[base].append(field_assignment_copy)
 
+        if (calling_form in {utils.CallingForm.RequestStreamingClient,
+                             utils.CallingForm.RequestStreamingBidi} and
+                len(base_param_to_attrs) > 1):
+            raise InvalidRequestSetup("""There can be at most 1 base request in a
```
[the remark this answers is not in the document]
software-dov: Can't remember why I did this. Will fix.

gapic/samplegen/samplegen.py
```diff
@@ -13,29 +13,50 @@
...
 
-# There are a couple of other names that should be reserved
-# for sample variable assignments, e.g. 'client', but there are diminishing returns:
-# the sample config author isn't a pathological adversary, they're a normal person
-# who may make mistakes occasionally.
-RESERVED_WORDS = frozenset(itertools.chain(keyword.kwlist, dir(__builtins__)))
-
-TEMPLATE_NAME = 'samplegen_template.j2'
+FILE_HEADER = {"copyrightLines": ["Copyright (C) 2019  Google LLC"],
+               "licenseLines": ["Licensed under the Apache License, Version 2.0 (the \"License\");",
```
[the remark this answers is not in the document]
software-dov: Binding the copyright and license to the template feels wrong. Using literals in the python code is marginally better because it leaves the door open to obtaining them from input files.

+1 to multiline string and `split("\n")`.

tests/unit/samplegen/test_template.py
software-dov: My initial thought was "technically doable but tricky," but I started messing with it and found a neat solution.

gapic/samplegen/samplegen.py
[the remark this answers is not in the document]
software-dov: Went with splitting lines in the template.

gapic/samplegen/samplegen.py
```diff
@@ -410,3 +439,36 @@ def _validate_loop(self, body):
...
+            "Could not find rpc in service {}: {}".format(service_name, rpc_name))
+
+    calling_form = utils.CallingForm.method_default(rpc)
+
+    v = Validator()
+    sample["request"] = v.validate_and_transform_request(calling_form,
+                                                         sample["request"])
+    v.validate_response(sample["response"])
+
+    region_tag = sample["region_tag"]
```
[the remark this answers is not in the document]
software-dov: Done

gapic/templates/examples/feature_fragments.j2
```diff
@@ -25,44 +25,78 @@ There is a little, but not enough for it to be important because
...
+#
+{% for line in fileHeader["copyrightLines"] -%}
+# {{ line }}
+{% endfor -%}
+#
+{% for line in fileHeader["licenseLines"] -%}
+# {{ line }}
+{% endfor -%}
+#
+# DO NOT EDIT! This is a generated sample ("{{ callingForm }}",  "{{ sample.region_tag }}")
```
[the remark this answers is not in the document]
software-dov: Done

gapic/templates/examples/feature_fragments.j2
```diff
@@ -25,44 +25,78 @@ There is a little, but not enough for it to be important because
...
+{% endmacro %}
+
 {% macro renderPrint(elts) %}
   {# First elment is a format string, remaining elements are the format string parameters #}
   {# Validating that the number of format params equals #}
   {# the number of remaining params is handled by real python code #}
   {% with fmtStr = ('\"' + elts[0] + '\"') |replace("%s", "{}") %}
-print({{ ([fmtStr] + elts[1:])|join(', ') }})
-  {% endwith -%}
+  {% with params = elts[1:]|map("replace","$resp", "response")|list %}
```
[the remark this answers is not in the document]
software-dov: Created a new filter that does this.

gapic/generator/generator.py
```diff
@@ -54,6 +55,7 @@ def __init__(self, opts: options.Options) -> None:
         self._env.filters['snake_case'] = utils.to_snake_case
         self._env.filters['sort_lines'] = utils.sort_lines
         self._env.filters['wrap'] = utils.wrap
+        self._env.filters['resp_to_response'] = samplegen.resp_to_response
```
lukesneeringer: Bikeshed: This name is a bit weird. It makes me think "response to response". Not having read the code yet, I am not sure what it actually is.

gapic/samplegen/samplegen.py
```diff
@@ -13,29 +13,54 @@
...
 VALID_CONFIG_TYPE = 'com.google.api.codegen.SampleConfigProto'
 
-# There are a couple of other names that should be reserved
-# for sample variable assignments, e.g. 'client', but there are diminishing returns:
-# the sample config author isn't a pathological adversary, they're a normal person
-# who may make mistakes occasionally.
-RESERVED_WORDS = frozenset(itertools.chain(keyword.kwlist, dir(__builtins__)))
-
-TEMPLATE_NAME = 'samplegen_template.j2'
+FILE_HEADER = {"copyrightLines": "Copyright (C) 2019  Google LLC",
```
lukesneeringer: This seems awfully Google-specific. I avoided putting license stuff in these files (opting instead to have Yoshi add it) since the generator was supposed to be company-agnostic.

If we are to add these, why not put it in the template rather than here?

gapic/samplegen/samplegen.py
```diff
@@ -92,6 +117,20 @@ class InvalidRequestSetup(SampleError):
     pass
 
 
+def resp_to_response(s):
```
lukesneeringer: I do not love making this a new filter in the Jinja environment; `|replace('$resp', 'response')` does the same thing.

If we need a "single home" for this logic, I would suggest naming it something more generic (e.g. `coerce_variable_name`) so that it can accommodate future massaging that may come up.

gapic/schema/api.py
```diff
@@ -68,10 +68,10 @@ def build(cls, file_descriptor: descriptor_pb2.FileDescriptorProto,
                 These are needed to look up messages in imported protos.
         """
         return _ProtoBuilder(file_descriptor,
-            file_to_generate=file_to_generate,
-            naming=naming,
-            prior_protos=prior_protos or {},
-        ).proto
+                             file_to_generate=file_to_generate,
```
lukesneeringer: What auto-formatter are you using? Perhaps we should settle on one and reify it in CI.

gapic/templates/examples/feature_fragments.j2
```diff
@@ -25,44 +25,78 @@ There is a little, but not enough for it to be important because
...
+{% endfor -%}
+#
+{% for line in fileHeader["licenseLines"].split("\n") -%}
+# {{ line }}
+{% endfor -%}
+#
+# DO NOT EDIT! This is a generated sample ("{{ callingForm }}",  "{{ sample.id }}")
+#
+# To install the latest published package dependency, execute the following:
+#   pip3 install TODO
```
lukesneeringer: `{{ api.naming.warehouse_package_name }}`

gapic/templates/examples/feature_fragments.j2
```diff
@@ -25,44 +25,78 @@ There is a little, but not enough for it to be important because
 
 {# response handling macros #}
 
+{% macro licenseSection(fileHeader, sample, callingForm) %}
+# -*- coding: utf-8 -*-
+#
+{% for line in fileHeader["copyrightLines"].split("\n") -%}
```
lukesneeringer: I do not really see why this is not just template text.

gapic/templates/examples/feature_fragments.j2
```diff
@@ -25,44 +25,78 @@ There is a little, but not enough for it to be important because
 
 {# response handling macros #}
 
+{% macro licenseSection(fileHeader, sample, callingForm) %}
```
lukesneeringer: Please use `snake_case` macro names and variable names (see PEP 8).

gapic/utils/code.py
```diff
@@ -21,3 +23,29 @@ def empty(content: str) -> bool:
     """
     return not any([i.lstrip() and not i.lstrip().startswith('#')
                     for i in content.split('\n')])
+
+
+T = TypeVar('T')
+
+
+def partition(iterator: Iterable[T],
```
lukesneeringer: Recommend flipping the function signature to `predicate, iterable` to match [itertools.filterfalse](https://docs.python.org/3/library/itertools.html#itertools.filterfalse) and others.

gapic/schema/api.py
software-dov: autopep8, part of my python-mode-hook.
Opened issue #142 as a feature task for CI. Do you have strong feelings on autopep8 vs black or something else? My vote is black because there are no choices to make.

gapic/templates/examples/feature_fragments.j2
software-dov: No really good reason, just unit testing and peace of mind.

gapic/generator/generator.py
software-dov: Sample configs use the special variable `$resp` to refer to the interesting proto message returned by the API call, i.e. for 'vanilla' methods it's just the return value, for paginated methods it's the repeated element, and so forth. Some part of the sample generation process needs to remove the `$`, and while that's happening it might as well get a less terse name. I really, really didn't want to lump that into the validation logic since it's just a dumb text replacement, so strategic template logic via filter it is.

gapic/samplegen/samplegen.py
software-dov: @vchudnov-g commented earlier that he was not fond of the repeated `|replace('$resp', 'response')`, so quick and dirty filter it was.
`coerce_variable_name` is more helpful from a slightly more abstract viewpoint but less descriptive of what is actually happening. The ideal solution would be to somehow use an in-template macro as a filter and avoid messing with the environments.

gapic/samplegen/samplegen.py
software-dov: The original intention was to read in copyright and license from input files, but somehow (can't remember where) I was convinced that would not be happening in the near future.
I'd rather just yank them and make it Yoshi's responsibility to add them then to add it to the template; keeping them as constants in python code makes it more straightforward to read them in from files later.

gapic/templates/examples/feature_fragments.j2
software-dov: Thanks!

gapic/utils/code.py
software-dov: Will do; my initial rationale was to give the predicate a default value, but consistency with stdlib and builtins is better.

gapic/templates/examples/feature_fragments.j2
software-dov: Given that that ship has sailed in the other parts of the fragments template, would you be okay with a separate, subsequent PR that snakeifies all the macro names?

software-dov: Tracked by #144

gapic/samplegen/samplegen.py
software-dov: Going with `coerce_variable_name` for the time being.

gapic/samplegen/samplegen.py
software-dov: After some second thoughts, I move to punt the issue and just remove the copyright and license until publishing samples and microgenerator created libraries is the topic at hand. We can then hash it out between us and Yoshi as to where the responsibility falls.

gapic/samplegen/samplegen.py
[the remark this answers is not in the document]
software-dov: Referenced below in a comment chain with @lukesneeringer : I move that the copyright and license are removed for the time being brought back upon discussion with Yoshi on where to get the information/who is responsible for it.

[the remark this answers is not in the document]
vchudnov-g: Let me TAL at those comments, but I am inclined, in order of preference, (a)to not remove them for now so that by default they're there (we can alter the mechanism later) or (b) to have a big ugly hulking TODO to minimize (but I don't dare hope eliminate) the chance of us forgetting to add the license before release.

gapic/samplegen/samplegen.py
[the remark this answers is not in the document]
vchudnov-g: OK, sounds good.

Addressing your previous comment above: IMO it's marginally easier to change the license+copyright in a template file than in a code file.

gapic/samplegen/samplegen.py
vchudnov-g: I suggest that for the short-term we either keep the license as a placeholder or add a really obvious TODO. My main driver is ensuring that we do not accidentally publish Google libraries generated with this tool that lack the legal boilerplate.

Longer term, I think this makes it high priority to require that all microgenerators have a way of specifying a license file which will then be commented into every generated artifact. Counterargument: users can do this themselves, post-generation. Counter-counter argument: we want to make it easy for users to have generated artifacts, and since we're generating, it should be trivial for us to do it.

gapic/samplegen/samplegen.py
vchudnov-g: sg

gapic/samplegen/samplegen.py
```diff
@@ -13,29 +13,40 @@
...
 
 VALID_CONFIG_TYPE = 'com.google.api.codegen.SampleConfigProto'
 
-# There are a couple of other names that should be reserved
-# for sample variable assignments, e.g. 'client', but there are diminishing returns:
-# the sample config author isn't a pathological adversary, they're a normal person
-# who may make mistakes occasionally.
-RESERVED_WORDS = frozenset(itertools.chain(keyword.kwlist, dir(__builtins__)))
+# TODO: read in copyright and license from files.
+FILE_HEADER: Dict[str, str] = {}
```
vchudnov-g: Could you have some placeholder values until we implement the TODO? Two reasons:

1. They will increase the odds we remember to implement the TODO before GA (especiallyif you make the placeholders really LOUD, and maybe with a special key phrase like `TODO` or `FIXME`
2. Until we implement the TODO, we can still see how the header will show up in the generated artifacts.

gapic/samplegen/samplegen.py
```diff
@@ -92,6 +103,20 @@ class InvalidRequestSetup(SampleError):
     pass
 
 
+def coerce_variable_name(s):
```
vchudnov-g: If the intent of this is to remain specific to replacing `$resp` (ie, it only takes one argument), I suggest renaming to `coerce_response_name`

Also, is the intent for the signature above and logic below to grow to accommodate the extra cases where the target is not `"response"`? (eg for paged methods, etc.)

gapic/samplegen/samplegen.py
```diff
@@ -206,6 +233,12 @@ def validate_and_transform_request(self,
             field_assignment_copy["field"] = attr
             base_param_to_attrs[base].append(field_assignment_copy)
 
+        if (calling_form in {utils.CallingForm.RequestStreamingClient,
+                             utils.CallingForm.RequestStreamingBidi} and
+                len(base_param_to_attrs) > 1):
+            raise InvalidRequestSetup(("There can be at most 1 base request in a sample"
+                                       " for a method with client side streaming"))
```
vchudnov-g: nitty nit: hyphen in "client-side"

gapic/templates/examples/sample.py.j2
```diff
@@ -0,0 +1,50 @@
...
+{# python code is responsible for all transformations: all we do here is render #}
+{% for importStatement in imports %}
+{{ importStatement }}
+{% endfor %}
+
+{# also need calling form #}
+def sample_{{ frags.renderMethodName(sample.rpc) }}({{ frags.printInputParams(sample.request) }}):
+    """{{ sample.description }}"""
+
+    # [START {{ sample.id }}_core]
```
vchudnov-g: We no longer include the `_core` region tag pair for now (cf https://github.com/googleapis/gapic-generator/pull/2810)

We may add it back when we update our plan for dealing with reference page (APIs Explorer) samples.

gapic/templates/examples/feature_fragments.j2
```diff
@@ -25,44 +25,69 @@ There is a little, but not enough for it to be important because
 
 {# response handling macros #}
 
+{% macro sample_header(fileHeader, sample, callingForm) %}
```
vchudnov-g: You're not using `fileHeader` here. I recommend we have a dummy placeholder as the value for now, and implement the logic here to render it.

gapic/templates/examples/feature_fragments.j2
```diff
@@ -25,44 +25,69 @@ There is a little, but not enough for it to be important because
...
+{% endmacro %}
+
 {% macro renderPrint(elts) %}
   {# First elment is a format string, remaining elements are the format string parameters #}
   {# Validating that the number of format params equals #}
   {# the number of remaining params is handled by real python code #}
   {% with fmtStr = ('\"' + elts[0] + '\"') |replace("%s", "{}") %}
-print({{ ([fmtStr] + elts[1:])|join(', ') }})
-  {% endwith -%}
+  {% with params = elts[1:]|map("coerce_variable_name")|list %}
```
vchudnov-g: Just curious: what does `elts` stand for? Is it a standard abbreviation?

gapic/utils/code.py
```diff
@@ -21,3 +23,29 @@ def empty(content: str) -> bool:
...
+
+T = TypeVar('T')
+
+
+def partition(predicate: Callable[[T], bool],
+              iterator: Iterable[T]) -> Tuple[List[T], List[T]]:
+    """Partitions an iterable into two lists based on a predicate
+
+    Args:
+        predicate Callable((T), bool) : A callable predicate on a single argument
```
vchudnov-g: nit: In other methods you have the type inside parentheses. Do so here too for consistency.

(BTW, is the style to repeat the type even though it's already listed in the type hints in the function signature?)

gapic/templates/examples/feature_fragments.j2
software-dov: It's intended to be short for 'elements'. I've seen other people use it a few places.

gapic/templates/examples/sample.py.j2
software-dov: Good to know. The region tag comments are one of the (thankfully) few remaining cargo cult elements of the rendered output.

gapic/utils/code.py
software-dov: A quick experiment indicates that pydoc is aware of type annotations. I'll remove the type info from the docstrings.

gapic/samplegen/samplegen.py
software-dov: Done

gapic/samplegen/samplegen.py
software-dov: I'm not sure I understand your question. The ideal solution to the problem this functionlet/filter is trying to solve would be
```
sed -i "s/\$resp/response/g" sample_file.py
```
Handling it in the validation code is fiddly, actually using sed in the pipeline won't work and doing a big string in python code that invokes rendering feels like a hack, so the substitution is punted to the template.

gapic/samplegen/samplegen.py
software-dov: Added them back with loud, dummy contents.

gapic/templates/examples/feature_fragments.j2
software-dov: Done

gapic/samplegen/samplegen.py
software-dov: +1 for making it standard that microgenerators add license and copyright information to samples.

gapic/samplegen/samplegen.py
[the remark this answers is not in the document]
software-dov: Went with (b) in the most recent commit.

gapic/samplegen/samplegen.py
lukesneeringer: Well, right now, the microgenerator adds it to samples and not the main files. Let's definitely not do that. :-)

Can we back this portion out of this PR and make it a separate item, and determine what the right thing to do is?

gapic/schema/api.py
lukesneeringer: I am not willing to use black because it generates non-PEP-8 compliant code. autopep8 is fine with me.

gapic/templates/examples/feature_fragments.j2
lukesneeringer: Yes.

gapic/utils/code.py
lukesneeringer: Python type hints are new-ish and not all tooling supports them yet, so repetition is semi common.

These things are parsed and have meaning though, and the one here is wrong; it should be `Callable[[T], bool]`.

gapic/samplegen/samplegen.py
software-dov: The license and copyright are now dummy values with TODO so that we don't forget.

gapic/schema/api.py
software-dov: Most of what I've read about black indicates that its output is PEP-8 compliant. What parts does it violate?

gapic/templates/examples/feature_fragments.j2
vchudnov-g: Ah! I've not seen it before. I've seen `elems`. Another abbrev added to my data banks.

gapic/samplegen/samplegen.py
vchudnov-g: The question was referring to the fact that sometimes we may choose to have $resp have a slightly different name, eg [here](https://github.com/googleapis/gapic-generator/blob/master/src/test/java/com/google/api/codegen/gapic/testdata/py/python_library.baseline#L6452) and [here](https://github.com/googleapis/gapic-generator/blob/master/src/test/java/com/google/api/codegen/gapic/testdata/py/python_library.baseline#L7637)

I think it's fine for each µgen to decide independently whether or not to always `$resp`→`response` in these cases or use an alternate name, though the set of alternate names should be small across languages to minimize forbidden user-provided identifiers.

gapic/samplegen/samplegen.py
```diff
@@ -320,6 +351,39 @@ def _validate_define(self, body: str):
...
+        contents_var = body.get("contents")
+        if not contents_var:
+            raise InvalidStatement(
+                "Missing key in 'write_file' statement: 'contents'")
+
+        base = contents_var.split(".")[0]
+        if base not in self.var_defs_:
+            raise UndefinedVariableReference("Reference to undefined variable: {}"
+                                             .format(base))
+
```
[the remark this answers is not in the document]
software-dov: Done

tests/unit/samplegen/test_template.py
```diff
@@ -153,9 +160,9 @@ def test_render_print_args():
     check_template(
         '''
         {% import "feature_fragments.j2" as frags %}
-        {{ frags.renderPrint(["Molluscs: %s, %s, %s", "squid", "clam", "whelk"]) }}
+        {{ frags.renderPrint(["$resp %s %s", "$resp.squids", "$resp.clams"]) }}
         ''',
-        '\nprint("Molluscs: {}, {}, {}", squid, clam, whelk)\n'
+        '\nprint("$resp {} {}".format(response.squids, response.clams))\n'
```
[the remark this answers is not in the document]
software-dov: I'm planning on turning all of these one-liners into triple-quote strings in the same review that snakifies all the macro names; just one big, shallow cleanup.

gapic/samplegen/samplegen.py
software-dov: The consistency that I aimed for in the calling-form handling template code is that `response` is always the most interesting top level proto response, i.e. it's the repeated element in paginated or streamed return values, it's the result of waiting on a promise, and so forth.

Issue #142, opened 2019-07-09
Format checking task in CI 


software-dov:
Add a format-checking task in that runs autopep8 (or black) on the checked in code. If there's a non-trivial diff, error out.

PR #146, opened 2019-07-11
Add an autopep8 CI task and run autopep8 on all py files

software-dov:
Implementation of #142

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -303,7 +303,9 @@ jobs:
           command: |
             pip install autopep8
       - run:
-          find gapic tests -name "*.py" | xargs autopep8 --in-place
+          name: Format files
+          command: |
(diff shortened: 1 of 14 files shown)


.circleci/config.yml
```diff
@@ -283,3 +289,22 @@ jobs:
...
+      - run:
+          name: Install git
+          command: |
+            apk add git
+      - run:
+          name: Install autopep8
+          command: |
+            pip install autopep8
+      - run:
+          find gapic tests -name "*.py" | xargs autopep8 --in-place
```
vchudnov-g: I see this is a valid shorthand syntax for a run step, but would it be helpful to also give it a name for consistency/descriptiveness.

software-dov: Done.

PR #148, opened 2019-07-12
Test and impl for generating samplegen manifest structure

software-dov:
This change includes machinery required to render the manifest but doesn not tie it together.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,6 +3,7 @@ branch = True
 omit =
   gapic/cli/*.py
   *_pb2.py
+  # Abstract methods play hob with coverage
 
 [report]
 fail_under = 100
@@ -14,3 +15,5 @@ exclude_lines =
     Impossible; skip coverage checks.
     # Ignore debug-only repr
     def __repr__
+    # Abstract methods by definition are not invoked
+    @abstractmethod
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -16,6 +16,7 @@ import itertools
 import jinja2
 import keyword
 import re
+import time
 
 from gapic.samplegen import utils
 
@@ -48,6 +49,7 @@ RESERVED_WORDS = frozenset(
                      }))
 
 TEMPLATE_NAME = "sample.py.j2"
+
 
 TransformedRequest = namedtuple("TransformedRequest", ["base", "body"])
 
@@ -501,8 +503,7 @@ def generate_sample(sample,
                                                          sample["request"])
     v.validate_response(sample["response"])
 
-    sample_id = sample["id"]
-    sample_fpath = sample_id + str(calling_form) + ".py"
+    sample_fpath = sample["id"] + str(calling_form) + ".py"
 
     sample["package_name"] = api_schema.naming.warehouse_package_name
 
@@ -511,3 +512,59 @@ def generate_sample(sample,
                                                 imports=[],
                                                 calling_form=calling_form,
                                                 calling_form_enum=utils.CallingForm)
+
+
+def generate_manifest(fpaths_and_samples,
+                      api_schema,
+                      *,
+                      manifest_time: int = None):
+    """Generate a samplegen manifest for use by sampletest
+
+    Args:
+        fpaths_and_samples (Iterable[Tuple[str, Mapping[str, ?]]]):
+                         The file paths and samples to be listed in the manifest
+
+        api_schema (~api.API): An API schema object.
+        manifest_time (int): Optional. An override for the timestamp in the name of the manifest filename.
+                             Primarily used for testing.
+
+    Returns:
+        Tuple[str, Dict[str,?]]: The filename of the manifest and the manifest data as a dictionary.
+
+    """
+    python_env = {
+        "environment": "python",
+        "bin": "python3",
+        "base_path": "sample/base/directory",
+        "invocation": "{bin} {path} @args",
+    }
+    all_info = {
+        "type": "manifest/samples",
+        "schema_version": 3,
+        "python": python_env,
+        "samples": [{
+            "python": python_env,
+            "sample": sample["id"],
+            "path": "{base_path}/%s" % fpath,
+            "region_tag": sample.get("region_tag", "")
+        } for fpath, sample in fpaths_and_samples],
+    }
+
+    dt = time.gmtime(manifest_time)
+    manifest_fname_template = ("{api}.{version}.python."
+                               "{year:04d}{month:02d}{day:02d}."
+                               "{hour:02d}{minute:02d}{second:02d}."
+                               "manifest.yaml")
+
+    manifest_fname = manifest_fname_template.format(
+        api=api_schema.naming.name,
+        version=api_schema.naming.version,
+        year=dt.tm_year,
+        month=dt.tm_mon,
+        day=dt.tm_mday,
+        hour=dt.tm_hour,
+        minute=dt.tm_min,
+        second=dt.tm_sec,
+    )
+
+    return manifest_fname, all_info
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -18,11 +18,11 @@ import keyword
 import re
 import time
 
-from gapic.samplegen import utils
+from gapic.samplegen import utils, yaml
 
-from collections import (defaultdict, namedtuple)
-from textwrap import dedent
-from typing import (Dict, List, Mapping, Set, Tuple)
+
+from collections import defaultdict, namedtuple
+from typing import Dict, List, Mapping, Optional, Set, Tuple
 
 # Outstanding issues:
 # * In real sample configs, many variables are
@@ -30,23 +30,29 @@ from typing import (Dict, List, Mapping, Set, Tuple)
 
 MIN_SCHEMA_VERSION = (1, 2, 0)
 
-VALID_CONFIG_TYPE = 'com.google.api.codegen.SampleConfigProto'
+VALID_CONFIG_TYPE = "com.google.api.codegen.SampleConfigProto"
 
 # TODO: read in copyright and license from files.
-FILE_HEADER: Dict[str, str] = {"copyright": "TODO: add a copyright",
-                               "license": "TODO: add a license"}
+FILE_HEADER: Dict[str, str] = {
+    "copyright": "TODO: add a copyright",
+    "license": "TODO: add a license",
+}
 
 RESERVED_WORDS = frozenset(
-    itertools.chain(keyword.kwlist,
-                    dir(__builtins__),
-                    {"client",
-                     "f",            # parameter used in file I/O statements
-                     "operation",    # temporary used in LROs
-                     "page",         # used in paginated responses
-                     "page_result",  # used in paginated responses
-                     "response",     # basic 'response'
-                     "stream",       # used in server and bidi streaming
-                     }))
+    itertools.chain(
+        keyword.kwlist,
+        dir(__builtins__),
+        {
+            "client",
+            "f",  # parameter used in file I/O statements
+            "operation",  # temporary used in LROs
+            "page",  # used in paginated responses
+            "page_result",  # used in paginated responses
+            "response",  # basic 'response'
+            "stream",  # used in server and bidi streaming
+        },
+    )
+)
 
 TEMPLATE_NAME = "sample.py.j2"
 
@@ -140,9 +146,9 @@ class Validator:
 
     # TODO: this will eventually need the method name and the proto file
     # so that it can do the correct value transformation for enums.
-    def validate_and_transform_request(self,
-                                       calling_form: utils.CallingForm,
-                                       request: List[Mapping[str, str]]) -> List[TransformedRequest]:
+    def validate_and_transform_request(
+        self, calling_form: utils.CallingForm, request: List[Mapping[str, str]]
+    ) -> List[TransformedRequest]:
         """Validates and transforms the "request" block from a sample config.
 
            In the initial request, each dict has a "field" key that maps to a dotted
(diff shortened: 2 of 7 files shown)


tests/unit/samplegen/test_samplegen.py
```diff
@@ -523,3 +523,62 @@ def test_coerce_response_name():
...
+        "samples": [
+            {
+                "python": {
+                    "environment": "python",
+                    "bin": "python3",
+                    "base_path": "sample/base/directory",
+                    "invocation": "{bin} {path} @args",
+                },
+                "sample": "squid_sample",
+                "path": "{base_path}/squid_fpath",
```
hzyi-google: Would you mind adding a `.py` extension here? So that it feels more real.

tests/unit/samplegen/test_samplegen.py
```diff
@@ -523,3 +523,62 @@ def test_coerce_response_name():
...
+    fname, info = samplegen.generate_manifest(
+        fpath_to_dummy_sample.items(),
+        DummyApiSchema(DummyNaming("Mollusc", "v1")),
+        # Empirically derived number such that the
+        # corresponding time_struct tests the zero
+        # padding in the returned filename.
+        manifest_time=4486525628
+    )
+
+    assert fname == "Mollusc.v1.python.21120304.090708.manifest.yaml"
```
hzyi-google: @vchudnov-g Is this format universally required?

gapic/samplegen/samplegen.py
```diff
@@ -501,8 +503,7 @@ def generate_sample(sample,
                                                          sample["request"])
     v.validate_response(sample["response"])
 
-    sample_id = sample["id"]
-    sample_fpath = sample_id + str(calling_form) + ".py"
+    sample_fpath = sample["id"] + str(calling_form) + ".py"
```
hzyi-google: IIRC when sample id is unique, that id will be used as the file name. Calling forms are only used to disambiguate when id is not unique (a sample config has more than one valid calling form specified).

@vchudnov-g can confirm if I am wrong.

tests/unit/samplegen/test_samplegen.py
software-dov: Done

tests/unit/samplegen/test_samplegen.py
software-dov: From the samplegen manifest doc:
"The name of the manifest file SHOULD be API.version.lang.YYYYMMDD.hhmmss.manifest.yaml, including the local date/time of generation"

gapic/samplegen/samplegen.py
```diff
@@ -511,3 +512,59 @@ def generate_sample(sample,
...
+    """
+    python_env = {
+        "environment": "python",
+        "bin": "python3",
+        "base_path": "sample/base/directory",
+        "invocation": "{bin} {path} @args",
+    }
+    all_info = {
+        "type": "manifest/samples",
+        "schema_version": 3,
```
vchudnov-g: FYI and reminder to myself: we've been (somewhat inconsistently) using `config_schema_version` because that's what gapic config uses because it's what the API compiler expects, IIUC[*]. However, I am thinking now that for stuff that does not need to go through that, we could be fine using "schema_version" as a shorter alias for the same metadatum.

[*] I cannot find a primary source for this at the moment. I was probably told this and I would assume I had verified it, but we should double-check in the interests of not cargo-culting. @hzyi-google , since you're currently working on gapic-generator, is this something you could check really quickly? Try changing the name of this field in the gapic config and see where what blows up?

gapic/samplegen/samplegen.py
```diff
@@ -511,3 +512,59 @@ def generate_sample(sample,
                                                 imports=[],
                                                 calling_form=calling_form,
                                                 calling_form_enum=utils.CallingForm)
+
+
+def generate_manifest(fpaths_and_samples,
```
vchudnov-g: Reference implementation: [sample-tester/gen_manifest](https://github.com/googleapis/sample-tester/blob/master/gen_manifest/gen_manifest.py#L76).

In particular, to emit YAML anchors I chose to emit the YAML by hand, partly because it wasn't obvious that PyYaml could do it. I just found a [link](https://www.wenbiancheng.com/question/yaml/18065427-Generating%20anchors%20with%20PyYAML.dump()?) that suggests that it can, though I'd want the anchor name to be human-readable.

gapic/samplegen/samplegen.py
```diff
@@ -511,3 +512,59 @@ def generate_sample(sample,
...
+    python_env = {
+        "environment": "python",
+        "bin": "python3",
+        "base_path": "sample/base/directory",
+        "invocation": "{bin} {path} @args",
+    }
+    all_info = {
+        "type": "manifest/samples",
+        "schema_version": 3,
+        "python": python_env,
```
vchudnov-g: I assume you have this so it can be an anchor that is referenced by each individual sample. That is correct.

gapic/samplegen/samplegen.py
```diff
@@ -511,3 +512,59 @@ def generate_sample(sample,
...
+
+
+def generate_manifest(fpaths_and_samples,
+                      api_schema,
+                      *,
+                      manifest_time: int = None):
+    """Generate a samplegen manifest for use by sampletest
+
+    Args:
+        fpaths_and_samples (Iterable[Tuple[str, Mapping[str, ?]]]):
```
lukesneeringer: Nit: `Any`, not `?`

gapic/samplegen/samplegen.py
```diff
@@ -511,3 +512,59 @@ def generate_sample(sample,
...
+                         The file paths and samples to be listed in the manifest
+
+        api_schema (~api.API): An API schema object.
```
lukesneeringer: Nit: `~.api.API`

gapic/samplegen/samplegen.py
```diff
@@ -511,3 +512,59 @@ def generate_sample(sample,
...
+        "bin": "python3",
+        "base_path": "sample/base/directory",
+        "invocation": "{bin} {path} @args",
+    }
+    all_info = {
+        "type": "manifest/samples",
+        "schema_version": 3,
+        "python": python_env,
+        "samples": [{
+            "python": python_env,
```
vchudnov-g: This should not be a _child_ of the sample, but rather merged into the sample. In other words, the final state when the YAML is parsed should be

```
type: manifest/samples
schema_version: 3
samples:
- environment: python  
  bin: python3  # used to run these items
  base_path: "/home/nobody/api/samples/"  
  invocation: "{bin} {path} @args"
  sample: "alice2"                               
  path: "{base_path}trivial/method/sample_alice" 
  region_tag: "sample_alice"
  foo: "bar"
```

We want to accomplish this succinctly by using anchors and references when emitting the YAML file as follows, even though the _parsed_ YAML will be identical to the above:

```
type: manifest/samples
schema_version: 3
python: &python                               
  environment: python                           
  bin: python3  
  base_path: "/home/nobody/api/samples/"  
  invocation: "{bin} {path} @args"               
samples:
- <<: *python
  sample: "alice2"                                
  path: "{base_path}trivial/method/sample_alice"  
  region_tag: "sample_alice"
  foo: "bar"
```

Check out [this link](https://yaml-online-parser.appspot.com/?yaml=---+++++++++++++++++++++++++++++++++++++++++++++++%0Atype%3A+manifest%2Fsamples%0Aschema_version%3A+3%0Apython%3A+%26python+++++++++++++++++++++++++++++++++++%0A++environment%3A+python+++++++++++++++++++++++++++++%0A++bin%3A+python3++%23+used+to+run+these+items%0A++base_path%3A+"%2Fhome%2Fnobody%2Fapi%2Fsamples%2F"++%0A++invocation%3A+"%7Bbin%7D+%7Bpath%7D+%40args"++++++++++++++++%0Asamples%3A%0A-+<<%3A+*python%0A++sample%3A+"alice2"++%0A++path%3A+"%7Bbase_path%7Dtrivial%2Fmethod%2Fsample_alice"++%0A++region_tag%3A+"sample_alice"%0A++foo%3A+"bar"&type=json) to an online parser with the above data.

Typing this comment made me realize I had two errors in http://go/actools-samplegen-manifest-generation: the format of the reference, and the fact that the elements of the anchor should not be in a list element in this case (though according to the online validator, they lead to the same outcome). These have been fixed.

gapic/samplegen/samplegen.py
software-dov: For the time being, since there is no code that iterates over the sample config, this is taking the form of a single boolean flag in the `generate_sample` function. The sample iterating code will do two passes, one to gather counts for ids and the second to generate the samples.

gapic/samplegen/samplegen.py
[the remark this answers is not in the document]
software-dov: You're right: it's something I have to remember to do, and I'm not very good at that.
I'd argue that both are important, and I'll add both during polishing.

gapic/samplegen/samplegen.py
software-dov: That's the plan.

gapic/samplegen/samplegen.py
software-dov: I'm in the middle of that judgement call, balancing with the competing forces of "generating yaml by hand is deeply tedious and unrewarding" vs "fighting with PyYaml makes me want to cry tears of frustration because of all the boilerplate".
There's no _good_ solution, just a matter of which one sounds less unpleasant.

gapic/samplegen/samplegen.py
software-dov: Thanks for clarifying that.

gapic/samplegen/samplegen.py
software-dov: The 'samplegen manifest files' document uses `schema_version`, not `config_schema_version` in its example. Is it in error?

gapic/samplegen/samplegen.py
```diff
@@ -511,3 +512,59 @@ def generate_sample(sample,
...
+                             Primarily used for testing.
+
+    Returns:
+        Tuple[str, Dict[str,?]]: The filename of the manifest and the manifest data as a dictionary.
+
+    """
+    python_env = {
+        "environment": "python",
+        "bin": "python3",
+        "base_path": "sample/base/directory",
```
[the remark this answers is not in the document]
software-dov: Went with something completely different; this whole section is drastically changed.

gapic/samplegen/samplegen.py
software-dov: Wound up rolling my own yaml emitter.

gapic/samplegen/samplegen.py
software-dov: Fixed.

gapic/samplegen/samplegen.py
hzyi-google: `config_schema_version` is not required by api-compiler. It's just a name used by gapic config. 
 When implementing sample config in the monolith, I used the same name for consistency (though yes, the initial design doc uses `schema_version`). 

I think it's OK to just call it `schema_version` in the manifest file. (After call, the manifest file is generated metadata, not exactly a `config` file).

.coveragerc
```diff
@@ -14,3 +15,5 @@ exclude_lines =
     Impossible; skip coverage checks.
     # Ignore debug-only repr
     def __repr__
+    # Abstract methods by definition are not invoked
+    @abstractmethod
```
hzyi-google: Sorry one more thing: newline here :)

gapic/samplegen/samplegen.py
vchudnov-g: +1 to going with `schema_version` (everywhere). Simpler and shorter, and as @hzyi-google points out, more accurate.

gapic/samplegen/samplegen.py
vchudnov-g: Nice!

gapic/samplegen/yaml.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+@dataclasses.dataclass()
+class Anchor(Element):
+    """An anchor to a map."""
+    target: str
+
+    def render(self, spaces: int = 0) -> str:
+        whitespace = " " * spaces
+        return f"{whitespace}<<: *{self.target}"
```
vchudnov-g: nit: IIUC, `<<: *label` is a reference or alias,  and the `&label` is the anchor

setup.py
```diff
@@ -46,6 +46,7 @@
         'jinja2 >= 2.10',
         'protobuf >= 3.7.1',
         'pypandoc >= 1.4',
+        'PyYAML >= 5.1.1',
```
vchudnov-g: Do we need the PyYAML inclusion here right now? AFAIK, the only place where we use PyYAML is in the tests.

I assume when you do read an actual external sample config YAML file, you will need it.

gapic/samplegen/yaml.py
```diff
@@ -0,0 +1,101 @@
...
+
+import dataclasses
+
+from abc import abstractmethod, ABC
+from textwrap import indent
+from typing import List, Optional
+
+"""Module containing classes used for dramatically simplified yaml rendering.
+
+The yaml used for generating the samplegen manifest is simple and self contained,
```
vchudnov-g: Very elegant and concise!

gapic/samplegen/yaml.py
software-dov: Good to know! I'll clean it up in a subsequent PR.

gapic/samplegen/yaml.py
software-dov: Thanks!

setup.py
software-dov: Right now, PyYAML is only needed for the test because the code to read in the sample config isn't hooked up. I suppose this then counts as a spurious dependency.

PR #151, opened 2019-07-18
Add a 'validate_expression' method (and tests) to Validator

software-dov:
The 'validate_expression' method validates that a dotted attribute expression, 
e.g. "squid.clam.whelk",
references a defined variable, and that each attribute in the chain is
valid, i.e. 'squid' is a defined variable, and its type has a 'clam'
attribute, whose type has a 'whelk' attribute.

Also, do initial attribute checking
**NOTE**: not all validation methods check expressions yet. 
This is a deliberate decision because
*) Determining whether an attribute is a map (and what the types of the key and value are) is non trivial and requires additional work.
*) This review was getting large as is.

All "attribute expressions", e.g. mollusc.cephalopod.squid, are
checked to make sure that the chain of attributes is valid.
Validity includes:
*) Only the final attribute may be repeated
*) The type of an attribute at position n in the chain has a field
   with the name of the attribute at position n+1.

Defines check that repeated fields can only come at the end of an expression.
Loop statements now check that the collection is a repeated field.

Also, move the manifest test into its own file because it got large.

### DIFF
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -355,8 +355,12 @@ class Validator:
             r"^(?P<attr_name>\$?\w+)(?:\[(?P<index>\d+)\])?$")
 
         toks = exp.split(".")
-        base_tok, previous_was_indexed = indexed_exp_re.match(toks[0]).groups()
+        match = indexed_exp_re.match(toks[0])
+        if not match:
+            raise BadAttributeLookup(f"Badly formatted attribute expression: {exp}")
 
+        base_tok, previous_was_indexed = (match.groupdict()["attr_name"],
+                                          bool(match.groupdict()["index"]))
         base = self.var_field(base_tok)
         if not base:
             raise UndefinedVariableReference(
@@ -365,9 +369,13 @@ class Validator:
             raise BadAttributeLookup(
                 "Cannot index non-repeated attribute: {}".format(base_tok))
 
-        name_and_lookup_generator = (indexed_exp_re.match(tok).groups()
-                                     for tok in toks[1:])
-        for attr_name, lookup_token in name_and_lookup_generator:
+        for tok in toks[1:]:
+            match = indexed_exp_re.match(tok)
+            if not match:
+                raise BadAttributeLookup(
+                    f"Badly formatted attribute expression: {tok}")
+
+            attr_name, lookup_token = match.groups()
             if base.repeated and not previous_was_indexed:
                 raise BadAttributeLookup(
                     "Cannot access attributes through repeated field: {}".format(attr_name))
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -23,7 +23,7 @@ from gapic.schema import (api, wrappers)
 
 from collections import (defaultdict, namedtuple, ChainMap as chainmap)
 from textwrap import dedent
-from typing import (Dict, List, Mapping, Optional, Set, Tuple)
+from typing import (ChainMap, Dict, List, Mapping, Optional, Set, Tuple)
 
 # Outstanding issues:
 # * In real sample configs, many variables are
@@ -154,27 +154,22 @@ class Validator:
         elif method.lro:
             response_type = method.lro.response_type
 
-        class FieldMocker:
-            # This is a shameless hack to work around the design of wrappers.Field
-            def __init__(self, val):
-                self.val = val
-
-            def __getattr__(self, name):
-                return self.val
+        # This is a shameless hack to work around the design of wrappers.Field
+        MockField = namedtuple("MockField", ["message", "repeated"])
 
         # TODO: pass var_defs_ around during response verification
         #       instead of assigning/restoring.
-        self.var_defs_: Dict[str, wrappers.Field] = chainmap({
+        self.var_defs_: ChainMap[str, wrappers.Field] = chainmap(
             # When validating expressions we need to store the Field,
             # not just the message type, because there are additional data we need:
             # whether a name refers to a repeated value (or a map),
             # and whether it's an enum or a message or a primitive type.
             # The method call response isn't a field, so construct an artificial
             # field that wraps the response.
-            "$resp": wrappers.Field(field_pb=FieldMocker(False),
-                                    message=response_type,
-                                    enum=None,
-                                    meta=None)})
+            {  # type: ignore
+                "$resp": MockField(response_type, False)
+            }
+        )
 
     def var_field(self, var_name: str) -> Optional[wrappers.Field]:
         return self.var_defs_.get(var_name)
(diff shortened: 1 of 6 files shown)


gapic/samplegen/samplegen.py
```diff
@@ -214,7 +242,7 @@ def validate_and_transform_request(
             field_assignment_copy = dict(field_assignment)
             input_param = field_assignment_copy.get("input_parameter")
             if input_param:
-                self._handle_lvalue(input_param)
+                self._handle_lvalue(input_param, str)
```
hzyi-google: Not quite following this. `input_parameter` isn't always a `string`, it can be any primitive types IIUC.

gapic/samplegen/samplegen.py
```diff
@@ -297,7 +324,49 @@ def validate_response(self, response):
 
             validater(self, body)
 
-    def _handle_lvalue(self, lval):
+    def validate_expression(self, exp: str) -> wrappers.Field:
+        """Validate an attribute chain expression.
+
+        Given a lookup expression, e.g. squid.clam.whelk,
```
hzyi-google: The expression could contain `[]`, `{}` and `%` as well. e.g., `$resp.books[0].name`.

gapic/samplegen/samplegen.py
```diff
@@ -297,7 +324,49 @@ def validate_response(self, response):
 
             validater(self, body)
```
hzyi-google: nit: validater -> validator?

gapic/samplegen/samplegen.py
```diff
@@ -297,7 +324,49 @@ def validate_response(self, response):
...
 
-    def _handle_lvalue(self, lval):
+    def validate_expression(self, exp: str) -> wrappers.Field:
```
hzyi-google: question: why doesn't this method start with `_`? It does not seem to be used outside this class.

gapic/samplegen/samplegen.py
software-dov: Good catch!

gapic/samplegen/samplegen.py
software-dov: That answers a question I added to the features sheet, which is whether indexed lookups are permitted.

gapic/samplegen/samplegen.py
software-dov: I wanted to call it directly in the unit tests, which from one point of view promotes it into a first class portion of the class's API.

gapic/samplegen/samplegen.py
software-dov: Yes, but from the point of view of the `validate_expression` it doesn't matter: all that the code needs to know is that it's _not_ a `MessageType`. Input parameters can be enums though, which I plan on addressing in a subsequent review.

gapic/samplegen/samplegen.py
software-dov: Wait, how can the expression contain "%"?

gapic/samplegen/samplegen.py
hzyi-google: Okay that makes sense. Would you mind adding a comment here?

gapic/samplegen/samplegen.py
hzyi-google: That's to access an entity of a resource name.

For example if the response contains a resource name location:

```
message FooResponse {

  // this string is a resource name, it follows the pattern "projects/{project}/locations/{location}"
  string location = 1;

}
```

Then

`$resp.location%project` refers to the `{project}` part of the `location` field.

The `%` is not needed for 1.2 though, so it's totally fine if you don't implement it for now. But we'll need to do this in the near future so just reserve the flexibility to support it later.

gapic/samplegen/samplegen.py
software-dov: Sure thing.

gapic/samplegen/samplegen.py
software-dov: That is not a thing I knew anything about. Is there documentation about it in the samplegen docs? Or documentation elsewhere?

hzyi-google: Sorry the doc missed this :( Tagged you where I updated the doc.

software-dov: I've added support for subscripting repeated attributes, e.g. `$resp.squids[0]`. I'm going to hold off on adding mapping lookup until a subsequent review, as I still have some work to do before supporting maps in general. I'll add a TODO.

tests/unit/samplegen/test_samplegen.py
```diff
@@ -12,453 +12,670 @@
...
+            DummyMethod(output=message_factory("mollusc"))
+        ).validate_response([define])
 
 
 def test_define_add_var():
-    samplegen.Validator().validate_response([{"define": "squid=$resp"},
-                                             {"define": "name=squid.name"}])
+    samplegen.Validator(
+        DummyMethod(output=message_factory("mollusc.name"))
+    ).validate_response([{"define": "squid=$resp"}, {"define": "name=squid.name"}])
```
vchudnov-g: Minor: I suggest consistency in whether you have the `response` inline or via a separate variable (cf previous cases above)

gapic/samplegen/samplegen.py
```diff
@@ -297,7 +329,72 @@ def validate_response(self, response):
...
+        """Validate an attribute chain expression.
+
+        Given a lookup expression, e.g. squid.clam.whelk,
+        recursively validate that each base has an attr with the name of the
+        next lookup lower down and that no attributes, besides possibly the final,
```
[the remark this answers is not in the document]
software-dov: You understand correctly. Will update.

gapic/samplegen/samplegen.py
```diff
@@ -297,7 +329,72 @@ def validate_response(self, response):
...
+                                          bool(match.groupdict()["index"]))
+        base = self.var_field(base_tok)
+        if not base:
+            raise UndefinedVariableReference(
+                "Reference to undefined variable: {}".format(base_tok))
+        if previous_was_indexed and not base.repeated:
+            raise BadAttributeLookup(
+                "Cannot index non-repeated attribute: {}".format(base_tok))
+
+        for tok in toks[1:]:
```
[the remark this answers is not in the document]
software-dov: This function is the result of really, really gross TDD and is in desperate need of a refactor. Will take a whack at that after I double check the test coverage you mention below.

gapic/samplegen/samplegen.py
```diff
@@ -297,7 +329,72 @@ def validate_response(self, response):
...
+            attr_name, lookup_token = match.groups()
+            if base.repeated and not previous_was_indexed:
+                raise BadAttributeLookup(
+                    "Cannot access attributes through repeated field: {}".format(attr_name))
+            if previous_was_indexed and not base.repeated:
+                raise BadAttributeLookup(
+                    "Cannot index non-repeated attribute: {}".format(attr_name))
+
+            # TODO: handle enums, primitives, and so forth.
+            attr = base.message.fields.get(attr_name)  # type: ignore
```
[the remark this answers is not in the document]
software-dov: You're right: I don't think that's currently covered by tests. I'll check and possibly add another.

gapic/samplegen/samplegen.py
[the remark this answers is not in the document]
software-dov: The problem is that the _field_ is what carries info about whether it's repeated, whereas the parent token in the parsed string is what carries info about whether the lookup is indexed.

tests/unit/samplegen/test_samplegen.py
software-dov: I'll clean it up in a followup review.

PR #156, opened 2019-07-19
Bump version to 0.12.0

software-dov:


lukesneeringer:
Leaving this at approved (so you do not have to wait for me again), but note that `setup.py` says 0.11.1, even though the PR says 0.12.0.

vchudnov-g:
lgtm, but TAL @ Luke's comment regarding the version in `setup.py`

software-dov:
I'll be sure to update the version to 0.12.0 in `setup.py`. I just want to keep this PR in the chamber, as it were, so that if we ever need to do another version bump quickly there's less hassle.

You've been pre-approved! :credit_card:

PR #162, opened 2019-07-26
Test request setup against method request type

software-dov:
Fields in the request are checked against the fields in the message type.

Also includes changes for properly handling and rendering top level
fields, plus associated tests.

The data model for transformed requests has changed substantially: instead of being primarily dict based, TransformedRequest now has a de-facto tagged union for a list of AttributeRequestSetup objects XOR a singleton AttributeRequestSetup. This change was made to facilitate top level fields in requests.

Also includes broad, shallow, standardizing cleanup to samplegen unit tests

### DIFF
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -270,17 +270,20 @@ class Validator:
                                 in the request message type.
 
         """
-        base_param_to_attrs: Dict[str, List[AttributeRequestSetup]] = defaultdict(list)
+        base_param_to_attrs: Dict[str,
+                                  List[AttributeRequestSetup]] = defaultdict(list)
 
         for r in request:
             duplicate = dict(r)
             val = duplicate.get("value")
             if not val:
-                raise InvalidRequestSetup("Missing keyword in request entry: 'value'")
+                raise InvalidRequestSetup(
+                    "Missing keyword in request entry: 'value'")
 
             field = duplicate.get("field")
             if not field:
-                raise InvalidRequestSetup("Missing keyword in request entry: 'field'")
+                raise InvalidRequestSetup(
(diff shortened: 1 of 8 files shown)


gapic/samplegen/samplegen.py
```diff
@@ -58,7 +59,28 @@
...
 
 
-TransformedRequest = namedtuple("TransformedRequest", ["base", "body"])
+@dataclasses.dataclass(frozen=True)
+class AttributeRequestSetup:
+    value: str
+    field: Optional[str] = None
+    value_is_file: bool = False
+    input_parameter: Optional[str] = None
+    comment: Optional[str] = None
```
lukesneeringer: Empty string is probably better than None for these last two.

gapic/samplegen/samplegen.py
```diff
@@ -58,7 +59,28 @@
...
+
+
+@dataclasses.dataclass(frozen=True)
+class TransformedRequest:
+    """Class representing a single field in a method call.
+
+    The Optional[single]/Optional[body] is workaround for not having tagged unions;
```
lukesneeringer: I read this twice and still do not understand it.

software-dov: Can you highlight what about the comment is confusing or lacking or otherwise suboptimal so I can better inform subsequent readers? I've been swimming in these waters all week, and _I_ know what I mean :stuck_out_tongue_closed_eyes:

gapic/samplegen/samplegen.py
software-dov: The only keys the request input dict is required to have are "field" and "value"; the others are optional. I feel that `Optional[str]` is a better representation, but if you disagree I will change to an empty str default.

PR #165, opened 2019-07-29
Add tests and implementation for enums in requests

software-dov:
Request setup directives can set enums. Setup directives pass the enum variant as a string.

Includes tests for invalid enum variants and attempting to set
non-existent fields of the enum itself.

### DIFF
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -335,27 +335,24 @@ class Validator:
                 elif attr.enum:
                     # A little bit hacky, but 'values' is a list, and this is the easiest
                     # way to verify that the value is a valid enum variant.
-                    witness = next((e.name
-                                    for e in attr.enum.values
-                                    if e.name == val),
-                                   None)
+                    witness = any(e.name == val for e in attr.enum.values)
                     if not witness:
                         raise InvalidEnumVariant(
                             "Invalid variant for enum {}: '{}'".format(attr, val))
                     # Python code can set protobuf enums from strings.
                     # This is preferable to adding the necessary import statement
                     # and requires less munging of the assigned value
-                    duplicate["value"] = f"'{witness}'"
+                    duplicate["value"] = f"'{val}'"
                     break
                 else:
                     raise TypeError
 
             if i != len(attr_chain) - 1:
+                # We broke out of the loop after processing an enum.
                 extra_attrs = ".".join(attr_chain[i:])
                 raise InvalidEnumVariant(
-                    f"Subsequent attributes in enum request setup: '{extra_attrs}'")
+                    f"Attempted to reference attributes of enum value: '{extra_attrs}'")
 
-                # TODO: what if there's more stuff in the chain?
             if len(attr_chain) > 1:
                 duplicate["field"] = ".".join(attr_chain[1:])
             else:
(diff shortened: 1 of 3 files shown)


gapic/samplegen/samplegen.py
```diff
@@ -332,23 +332,30 @@ def validate_and_transform_request(self,
 
                 if attr.message:
                     base = attr.message
-
+                elif attr.enum:
+                    # A little bit hacky, but 'values' is a list, and this is the easiest
+                    # way to verify that the value is a valid enum variant.
+                    witness = next((e.name
```
vchudnov-g: Can't you use `any`? `witness = any([e.name == val for e in attr.enum.values])` It reads a lot cleaner, IMO. And in line 348, you can use `f"'{val}'"`

gapic/samplegen/samplegen.py
```diff
@@ -332,23 +332,30 @@ def validate_and_transform_request(self,
...
-                #     # Python code can set protobuf enums from strings.
-                #     # This is preferable to adding the necessary import statement
-                #     # and requires less munging of the assigned value
-                #     duplicate["value"] = f"'{witness}'"
-
-            # TODO: what if there's more stuff in the chain?
+            if i != len(attr_chain) - 1:
+                extra_attrs = ".".join(attr_chain[i:])
+                raise InvalidEnumVariant(
+                    f"Subsequent attributes in enum request setup: '{extra_attrs}'")
```
vchudnov-g: It's probably just me, but I didn't find this message illuminating. How about: `"Attempted to reference attributes of enum value"`? Also, inside this `if` statement I would add a comment saying `# we broke out of the loop after processing an enum`

gapic/samplegen/samplegen.py
```diff
@@ -332,23 +332,30 @@ def validate_and_transform_request(self,
...
+
+                # TODO: what if there's more stuff in the chain?
```
vchudnov-g: Is this TODO still relevant? If so, could you clarify it a little (and probably dedent it)?

gapic/samplegen/samplegen.py
software-dov: `any` returns a boolean. If we were on python 3.8 I could use `any` and assignment expressions in something like
```
if any((witness := e.name). == val for e  in attr.enum.values):
...
else:
   raise ...
```
which is only marginally clearer, IMO.

gapic/samplegen/samplegen.py
software-dov: It is not.

tests/unit/samplegen/common_types.py
```diff
@@ -17,6 +17,8 @@
 from collections import namedtuple
 from typing import(Iterable, Optional)
 
+from google.protobuf import descriptor_pb2
```
hzyi-google: Should it be `descriptor_pb2` or `descriptor`?

software-dov: Other code in the microgenerator uses `descriptor_pb2`.

gapic/samplegen/samplegen.py
```diff
@@ -335,27 +335,24 @@ def validate_and_transform_request(self,
                 elif attr.enum:
                     # A little bit hacky, but 'values' is a list, and this is the easiest
```
vchudnov-g: Do you still need this comment?

software-dov: I suppose it's not that helpful any more. I'll remove it in a subsequent PR.

PR #166, opened 2019-07-31
All response variable/attribute expressions are validated

software-dov:
All statements and expressions that reference the method response are validated,
e.g. {"print": ["%s", "$resp.cephalopods[0].mantle"]},
will be checked to make sure that the response message type has a
repeated "cephalopods" field, and that that type has a "mantle" field.

Includes a major refactor of Validator.validate_expression

### DIFF
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -525,7 +525,7 @@ class Validator:
 
         return validate_recursively(exp, self.var_defs_)
 
-    def _handle_lvalue(self, lval: str, type_=None):
+    def _handle_lvalue(self, lval: str, type_: wrappers.Field):
         """Conducts safety checks on an lvalue and adds it to the lexical scope.
 
         Raises:
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -26,6 +26,8 @@ from collections import (defaultdict, namedtuple, ChainMap as chainmap)
 from textwrap import dedent
 from typing import (ChainMap, Dict, List, Mapping, Optional, Set, Tuple, Union)
 
+from google.protobuf import descriptor_pb2
+
 # Outstanding issues:
 # * In real sample configs, many variables are
 #   defined with an _implicit_ $resp variable.
@@ -179,6 +181,18 @@ def coerce_response_name(s: str) -> str:
     return s.replace("$resp", "response")
 
 
+@dataclasses.dataclass(frozen=True)
+class LoopParameterField(wrappers.Field):
+    # This class is a hack for assigning the iteration variable in a collection loop.
+    # The variable is assigned to the field that represents the collection,
+    # but while the collection is a repeated field, the iteration variable is not.
+    # When adding the iteration variable to the lexical scope,
+    # instead of being directly assigned to the collection variable,
+    # it is copied into an instance of this class, which has a value
+    # override for 'repeated'.
+    repeated: bool = False
+
+
 class Validator:
     """Class that validates a sample.
 
@@ -319,7 +333,8 @@ class Validator:
 
             input_parameter = duplicate.get("input_parameter")
             if input_parameter:
-                self._handle_lvalue(input_parameter, str)
+                self._handle_lvalue(input_parameter, wrappers.Field(
+                    field_pb=descriptor_pb2.FieldDescriptorProto(type=9)))
 
             attr_chain = field.split(".")
             base = self.request_type_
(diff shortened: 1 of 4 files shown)


gapic/samplegen/samplegen.py
```diff
@@ -436,48 +451,79 @@ def validate_expression(self, exp: str) -> wrappers.Field:
...
-                "Reference to undefined variable: {}".format(base_tok))
-        if previous_was_indexed and not base.repeated:
-            raise BadAttributeLookup(
-                "Cannot index non-repeated attribute: {}".format(base_tok))
-
-        for tok in toks[1:]:
-            match = indexed_exp_re.match(tok)
+        chain_link_re = re.compile(
+            r"""
+            (?P<attr_name>\$?\w+)(?:\[(?P<index>\d+)\]|\{["'](?P<key>[^"']+)["']\})?$
```
hzyi-google: This works like magic.

hzyi-google:
LGTM but let's have @vchudnov-g TAL as well.

gapic/samplegen/samplegen.py
software-dov: Is that good or bad? It's far from the worst regex I've ever written :P

gapic/samplegen/samplegen.py
```diff
@@ -179,6 +181,18 @@ def coerce_response_name(s: str) -> str:
     return s.replace("$resp", "response")
 
 
+@dataclasses.dataclass(frozen=True)
+class LoopParameterField(wrappers.Field):
+    # This class is a hack for assigning the iteration variable in a collection loop.
```
vchudnov-g: This comment was really hard to parse..... Maybe putting the class after `_validate_loop()` would give more context (it certainly helped me). I think I partially see what you're getting at, but where is the requirement that `repeated=False` coming from? (as in, what code line beak if you don't do the override?)

(Is this crying out for a non-hacky approach?)

gapic/samplegen/samplegen.py
```diff
@@ -586,12 +619,7 @@ def _validate_write_file(self, body):
...
 
-        # TODO: check the rest of the elements for valid subfield attribute
-        base = contents_var.split(".")[0]
-        if base not in self.var_defs_:
-            raise UndefinedVariableReference(
-                "Reference to undefined variable: {}".format(base)
-            )
+        self.validate_expression(contents_var)
 
     def _validate_loop(self, body):
```
vchudnov-g: I think the name `body` here is confusing, because you're referring to the whole `loop` directive, but one of its fields is itself called `body`. Maybe you can s/`body`/`loop`/ ? wdyt?

tests/unit/samplegen/test_samplegen.py
```diff
@@ -1048,6 +1317,175 @@ def test_validate_expression_base_attr_is_repeated():
...
+            )
+        },
+        type="MOLLUSC_TYPE"
+    )
+    method = DummyMethod(output=OutputType)
+    v = samplegen.Validator(method)
+    v.validate_expression('$resp.cephalopods{"squid"}')
+
+
+def test_validate_expression_mapped_no_message():
```
vchudnov-g: is this `no_message` or `non_map`? This looks identical to the next test case.....

Since it's hard to eyeball these testcases, could you, in general, add a comment to the lines that is expected to make the call fail? (eg the `map_field` set to `False`)

tests/unit/samplegen/test_samplegen.py
```diff
@@ -1048,6 +1317,175 @@ def test_validate_expression_base_attr_is_repeated():
...
+        },
+        type="MOLLUSC_TYPE"
+    )
+    method = DummyMethod(output=OutputType)
+    v = samplegen.Validator(method)
+    with pytest.raises(samplegen.BadAttributeLookup):
+        v.validate_expression('$resp.cephalopods{"squid"}.mantle')
+
+
+def test_validate_expression_mapped_no_message():
```
vchudnov-g: This is shadowing the previous (repeated) definition....

tests/unit/samplegen/test_samplegen.py
```diff
@@ -1048,6 +1317,175 @@ def test_validate_expression_base_attr_is_repeated():
...
+            )
+        },
+        type="MOLLUSC_TYPE"
+    )
+    method = DummyMethod(output=OutputType)
+    v = samplegen.Validator(method)
+    v.validate_expression('$resp.cephalopods{"squid"}.mantle')
+
+
+def test_validate_expression_map_lookup_terminal_lookup():
```
vchudnov-g: Which test case checks the non-indexed terminal collection passing case and the non-indexed, non-terminal collection failing test case?

gapic/samplegen/samplegen.py
```diff
@@ -434,52 +449,82 @@ def validate_expression(self, exp: str) -> wrappers.Field:
...
+
+                value_field = message.fields.get("value")
+                if not value_field:
+                    raise BadAttributeLookup(
+                        f"Mapped attribute has no value field: {base}")
+
+                value_message = value_field.message
+                if not value_message:
+                    raise BadAttributeLookup(
+                        f"Mapped value field is not a message: {base}")
```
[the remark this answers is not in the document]
software-dov: This is informed guesswork on my part coming from https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto#L496 , which Hanzhen pointed me towards. The way I read the comment, `map` is transmuted by protoc into an internal message type with three fields, `key`, `value` and (implementation defined) `map_entry`; and then it itself is a repeated field in the higher level message. The `options.map_entry` is another sign.
Please let me know if I'm misinterpreting the docs, or if there's more edifying documentation elsewhere. Proto maps are frustratingly opaque and sparsely documented.

gapic/samplegen/samplegen.py
```diff
@@ -434,52 +449,82 @@ def validate_expression(self, exp: str) -> wrappers.Field:
...
+            base = expression[:first_dot] if first_dot > 0 else expression
+            match = chain_link_re.match(base)
             if not match:
                 raise BadAttributeLookup(
-                    f"Badly formatted attribute expression: {tok}")
+                    f"Badly formed attribute expression: {expression}")
+
+            name, idxed, mapped = (match.groupdict()["attr_name"],
+                                   bool(match.groupdict()["index"]),
+                                   bool(match.groupdict()["key"]))
```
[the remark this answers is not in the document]
software-dov: Yes, that's correct. The regex groups save the values of `index` or `key` purely for debugging, and because at some point in the future they may be useful.

gapic/samplegen/samplegen.py
software-dov: I think `s/body/loop` is an improvement.

tests/unit/samplegen/test_samplegen.py
software-dov: Good catch!

gapic/samplegen/samplegen.py
software-dov: I'll move the class definition closer to the intended use. I would argue that the hack is inevitable, but apparently the comment is doing a bad job of describing the situation.

The collection over which we wish to iterate is a repeated field. The iteration parameter is assigned to that field, one element at a time, but the parameter itself is not repeated. The data structure representing the iteration parameter needs to reflect that it is not repeated. The `wrappers.Field` class relies on the underlying `FieldDescriptorProto` for most of its attributes; `repeated` is the _only_ one we want to change.

gapic/samplegen/samplegen.py
```diff
@@ -434,52 +449,82 @@ def validate_expression(self, exp: str) -> wrappers.Field:
...
-                    "Cannot access attributes through repeated field: {}".format(attr_name))
-            if previous_was_indexed and not base.repeated:
+                    f"Collection lookup on non-repeated field: {base}")
+
+            # Can only ignore indexing or mapping in an indexed (or mapped) field
+            # if it is the terminal point in the expression.
+            if field.repeated and not (idxed or mapped) and first_dot != -1:
                 raise BadAttributeLookup(
-                    "Cannot index non-repeated attribute: {}".format(attr_name))
+                    (f"Accessing attributes from a non-indexed"
```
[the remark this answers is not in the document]
software-dov: Good point about mixing the proto info with the sample config.

tests/unit/samplegen/test_samplegen.py
software-dov: Good catch. This one is an unintended, name-shadowing duplicate.

tests/unit/samplegen/test_samplegen.py
software-dov: `test_validate_expression_repeated_attrs` verifies the non-indexed, non-terminal collection failure. I'll rename the testcase to make that more clear.

gapic/samplegen/samplegen.py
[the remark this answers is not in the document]
vchudnov-g: I don't have any deeper knowledge about this, but your interpretation seems consistent with the reference you included.

tests/unit/samplegen/test_samplegen.py
vchudnov-g: And non-indexed terminal collection?

gapic/samplegen/samplegen.py
vchudnov-g: Oh, I think I see: for the iteration variable, you need to use the same type `T` as that of the elements of the collection. Since you're using `FieldDescriptionProto` to get the type info, and since the concept of `collection<T>` in proto is manifested as a `repeated` field of type `T`, you can obtain `T` by getting the type of the collection minus the `repeated`. Does this capture it?

If you can adapt some wording similar to the above (still needs some massaging), it might make the explanation clearer.

software-dov: That's it exactly. I like your wording slightly better than what I'll have, will try tweaking it and see if I can get something good.

tests/unit/samplegen/test_samplegen.py
software-dov: A non-indexed terminal collection is explicitly allowed.
```
{"define": "molluscs=$resp.molluscs"}
{"loop": {"collection": "molluscs", "variable": "m", "body": [{"print": ["Mollusc: %s", "m"]}]}}
```

tests/unit/samplegen/test_samplegen.py
software-dov: Adding comments to the tests added in this PR. I'll annotate the others in a subsequent PR.

tests/unit/samplegen/test_samplegen.py
vchudnov-g: Right, and you have a test for that, correct?

software-dov: Yes, `test_validate_expression_base_attr_is_repeated`

PR #173, opened 2019-08-08
Add python3.5 to showcase unit and integration tests

software-dov:
This is the first part of a new regression against using python3.5
with the generated libraries.

Note: do not approve and merge this PR yet. The first commit is used for regression testing.


software-dov:
Abandoning; it's just too difficult to generate the library with one python version and run with another.

lukesneeringer:
Yeah, I dropped 3.5 testing for the same reason. It was just too difficult to get it working.

PR #182, opened 2019-08-15
Initial part of abstracting out python specifics from sample generation

software-dov:
Move manifest generation to its own file
Abstract out python-specific manfiest values
Provide a default response if sample spec doesn't contain one

Move some code into utility files
Minor cleanups

### DIFF
--- a/gapic/generator/generator.py
+++ b/gapic/generator/generator.py
@@ -140,7 +140,6 @@ class Generator:
                 spec["id"] = sample_id
                 id_to_samples[sample_id].append(spec)
 
-        # Interpolate the special variables in the sample_out_dir template.
         out_dir = "samples"
         fpath_to_spec_and_rendered = {}
         for samples in id_to_samples.values():
@@ -169,10 +168,9 @@ class Generator:
 
         # Only generate a manifest if we generated samples.
         if output_files:
-            manifest_fname, manifest_doc = manifest.generate_manifest(
+            manifest_fname, manifest_doc = manifest.generate(
                 ((fname, spec)
                  for fname, (spec, _) in fpath_to_spec_and_rendered.items()),
-                out_dir,
                 api_schema
             )
 
--- a/gapic/generator/generator.py
+++ b/gapic/generator/generator.py
@@ -19,9 +19,9 @@ import os
 from typing import (Any, DefaultDict, Dict, Mapping, List)
 from hashlib import sha256
 from collections import (OrderedDict, defaultdict)
-from gapic.samplegen_utils.utils import is_valid_sample_cfg
+from gapic.samplegen_utils.utils import (coerce_response_name, is_valid_sample_cfg)
 from gapic.samplegen_utils.types import InvalidConfig
-from gapic.samplegen import samplegen
+from gapic.samplegen import (manifest, samplegen)
 from gapic.generator import options
 from gapic.generator import formatter
 from gapic.schema import api
@@ -56,7 +56,7 @@ class Generator:
         self._env.filters['snake_case'] = utils.to_snake_case
         self._env.filters['sort_lines'] = utils.sort_lines
         self._env.filters['wrap'] = utils.wrap
-        self._env.filters['coerce_response_name'] = samplegen.coerce_response_name
+        self._env.filters['coerce_response_name'] = coerce_response_name
 
         self._sample_configs = opts.sample_configs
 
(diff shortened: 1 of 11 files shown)


gapic/generator/generator.py
```diff
@@ -167,7 +169,7 @@ def _generate_samples_and_manifest(
 
         # Only generate a manifest if we generated samples.
         if output_files:
-            manifest_fname, manifest_doc = samplegen.generate_manifest(
+            manifest_fname, manifest_doc = manifest.generate_manifest(
```
vchudnov-g: Minor Gophery style comment: consider avoiding stutter by having `manifest.generate(...)` instead

gapic/samplegen/manifest.py
```diff
@@ -0,0 +1,120 @@
...
+        manifest_time (int): Optional. An override for the timestamp in the name of the manifest filename.
+                             Primarily used for testing.
+
+    Returns:
+        Tuple[str, yaml.Doc]: The filename of the manifest and the manifest data as a dictionary.
+
+    """
+    # Use iter([]) instead of a generator expression due to a bug in pytest.
+    # See https://github.com/pytest-dev/pytest-cov/issues/310 for details.
+    base_path = next(
```
vchudnov-g: How about adding this logic to `yaml.Map.get(key, default)`? It would make this part cleaner and shorter.

gapic/samplegen/manifest.py
```diff
@@ -0,0 +1,120 @@
...
+                elements=[
+                    [  # type: ignore
+                        # Mypy doesn't correctly intuit the type of the
+                        # "region_tag" conditional expression.
+                        yaml.Alias(environment.anchor_name or ""),
+                        yaml.KeyVal("sample", sample["id"]),
+                        yaml.KeyVal(
+                            "path",
+                            "'{base_path}/%s'" % os.path.relpath(
+                                fpath, base_path)
```
vchudnov-g: Minor:

IIUC, our calling path always ensure `fpath` will always begin with `base_path`, so for our current use cases, we're covered.

Since this is its own module, it might make sense to verify that assumption? IIUC, when that assumption fails `relpath` returns a relative path relative to the `cwd`, which will not in general remain correct if we prefix the result with `{base_path}`.

gapic/generator/generator.py
vchudnov-g: How is the manifest time passed in?

gapic/samplegen/samplegen.py
```diff
@@ -160,6 +144,13 @@ def __init__(self, method: wrappers.Method):
             }
         )
 
+    @staticmethod
+    def preprocess_sample(api_schema, sample):
+        sample["package_name"] = api_schema.naming.warehouse_package_name
+        # If the sample author didn't define a response handling section,
```
vchudnov-g: nit: Turn this into a doc string.

gapic/generator/generator.py
software-dov: It isn't; it's an optional parameter with a default value of `None`, which means "use system time". The parameter is given a non-default value just for unit tests.

vchudnov-g: Oh, I see, the `None` value causes the system time. OK.

One stylistic note: In case this app ever decides to output more than one timestamped file at a time, it would be good to have a time value obtained once (globally or at module init), so that all timestamped outputs can use it and label artifacts as  coming from the same invocation.

software-dov: Okay. I did it this way because it made unit testing easier, but mocking `time` does the same thing. I'll keep that in mind if more than one manifest needs to be generated per run.

gapic/samplegen/samplegen.py
software-dov: Done.

gapic/samplegen/manifest.py
software-dov: Yes. That's a lot better. This is why we have abstract data types =P

gapic/generator/generator.py
software-dov: Done.

gapic/samplegen/manifest.py
software-dov: Good call. I will verify.

Aside: one of my least favorite things to deal with in all of programming is keeping track of relative path fragments and prefixes. In all of my homegrown code, all paths are absolute all the time, and relative calculations are handled on the fly (with caching).

software-dov: Verified via exception.

PR #198, opened 2019-08-29
Correctly find the sample template when run via protoc

software-dov:
Add a sample_template parameter to sample generation; this is just an internal parameter, not a CLI visible parameter, and it solves a problem I ran into when manually testing sample generation via the protoc plugin.

Adjust tests to pass the parameter

Quick and dirty test invoking the microgenerator via protoc plugin indicates that the template is now findable.

### DIFF
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -481,8 +481,10 @@ class Validator:
         num_prints = fmt_str.count("%s")
         if num_prints != len(body) - 1:
             raise types.MismatchedFormatSpecifier(
-                "Expected {} expresssions in format string but received {}".format(
-                    num_prints, len(body) - 1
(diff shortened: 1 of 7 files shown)


gapic/samplegen/samplegen.py
```diff
@@ -21,7 +21,7 @@
 import time
 
 from gapic.samplegen_utils import types
-from gapic.schema import (api, wrappers)
+from gapic.schema import (wrappers)
```
vchudnov-g: rm paren?

software-dov: Done.

software-dov:
Correct, this is not a CLI parameter. Updated the PR description to indicate that this solves a problem when invoking the generator as a protoc plugin.

PR #206, opened 2019-09-25
Fix map field generation

pragkent:
```proto
message Test {
    map<string, string> items = 1;
}
```
Currently, map field listed above will be generated as:
```python
items = proto.MapField(proto.STRING, proto.MESSAGE, number=1, message=ItemsEntry,)
```

Because `ItemsEntry` is not defined, this will cause a syntax error.


pragkent:
@googlebot I signed it!

Issue #194, opened 2019-08-27
Add resource name support in sample request construction


software-dov:
Add code and templates such that samples can use resource names in constructing requests.
Depends on #193

software-dov:
Implemented by #208

PR #208, opened 2019-10-03
Test and impl for resource-name requests (logic side) …

software-dov:
Contains logic and tests for validating and transforming resource-name based requests.

Also contains test and impl for the template rendering.

End-to-end testing is forthcoming.
Addresses #194

### DIFF
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -102,6 +102,7 @@ class TransformedRequest:
     base: str
     single: Optional[AttributeRequestSetup]
     body: Optional[List[AttributeRequestSetup]]
+    pattern: Optional[str] = None
 
 
 class Validator:
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -112,8 +112,8 @@ class TransformedRequest:
 
     @classmethod
     def build(cls, request_type: wrappers.MessageType, api_schema, base: str,
-              attrs: List[str], is_resource_request: bool):
-        """Build a TransformedRequest based on parsed inpu
+              attrs: List[AttributeRequestSetup], is_resource_request: bool):
+        """Build a TransformedRequest based on parsed input.
 
         Acts as a factory to hide complicated logic for resource based requests.
 
@@ -400,24 +400,24 @@ class Validator:
                                   RequestEntry] = defaultdict(RequestEntry)
 
         for r in request:
-            duplicate = dict(r)
-            val = duplicate.get("value")
+            r_dup = dict(r)
+            val = r_dup.get("value")
             if not val:
                 raise types.InvalidRequestSetup(
                     "Missing keyword in request entry: 'value'")
 
-            field = duplicate.get("field")
+            field = r_dup.get("field")
             if not field:
                 raise types.InvalidRequestSetup(
                     "Missing keyword in request entry: 'field'")
 
-            spurious_keywords = set(duplicate.keys()) - self.VALID_REQUEST_KWORDS
-            if spurious_keywords:
+            spurious_kwords = set(r_dup.keys()) - self.VALID_REQUEST_KWORDS
+            if spurious_kwords:
                 raise types.InvalidRequestSetup(
                     "Spurious keyword(s) in request entry: {}".format(
-                        ", ".join(f"'{kword}'" for kword in spurious_keywords)))
+                        ", ".join(f"'{kword}'" for kword in spurious_kwords)))
 
-            input_parameter = duplicate.get("input_parameter")
+            input_parameter = r_dup.get("input_parameter")
             if input_parameter:
                 self._handle_lvalue(input_parameter, wrappers.Field(
                     field_pb=descriptor_pb2.FieldDescriptorProto()))
@@ -428,7 +428,7 @@ class Validator:
             percent_idx = field.find('%')
             if percent_idx == -1:
                 base_param, attr = self._normal_request_setup(
-                    base_param_to_attrs, val, duplicate, field, base)
+                    base_param_to_attrs, val, r_dup, field, base)
 
                 request_entry = base_param_to_attrs.get(base_param)
                 if request_entry and request_entry.is_resource_request:
@@ -450,11 +450,11 @@ class Validator:
                         "Method request type {} has no attribute: '{}'".format(
                             self.request_type_, base_param))
 
-                duplicate["field"] = resource_attr
+                r_dup["field"] = resource_attr
                 request_entry = base_param_to_attrs[base_param]
                 request_entry.is_resource_request = True
                 request_entry.attrs.append(
-                    AttributeRequestSetup(**duplicate))  # type: ignore
+                    AttributeRequestSetup(**r_dup))  # type: ignore
 
         client_streaming_forms = {
             types.CallingForm.RequestStreamingClient,
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -212,15 +212,15 @@ class Validator:
     VAL_KWORD = "value"
     BODY_KWORD = "body"
 
+    # This regex matches each variable or attribute in the following example
+    # expression and indicates whether the lookup is indexed, mapped, or neither
+    #
+    # cephalopoda.coleoidea[0].suborder{"incirrina"}
     EXPRESSION_ATTR_RE = re.compile(
         r"""
         (?P<attr_name>\$?\w+)(?:\[(?P<index>\d+)\]|\{["'](?P<key>[^"']+)["']\})?$
         """.strip())
 
-    # This regex matches each variable or attribute in the following example
-    # expression and indicates whether the lookup is indexed, mapped, or neither
-    #
-    # cephalopoda.coleoidea[0].suborder{"incirrina"}
     VALID_REQUEST_KWORDS = frozenset(
         ("value", "field", "value_is_file", "input_parameter", "comment"))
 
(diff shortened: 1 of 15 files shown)


gapic/samplegen/samplegen.py
```diff
@@ -102,6 +104,81 @@ class TransformedRequest:
...
     body: Optional[List[AttributeRequestSetup]]
+    pattern: Optional[str] = None
+
+    # Resource patterns look something like
+    # kingdom/{kingdom}/phylum/{phylum}/class/{class}
+    RESOURCE_RE = re.compile(r"\{([^}/]+)\}")
+
+    @classmethod
+    def build(cls, request_type, api_schema, base, attrs, is_resource_request):
+        """Build a TransformedRequest based on parsed inpu
```
hzyi-google: input

gapic/samplegen/samplegen.py
```diff
@@ -102,6 +104,81 @@ class TransformedRequest:
...
+        if not attrs[0].field:
+            return cls(base=base, body=None, single=attrs[0])
+        elif not is_resource_request:
+            return cls(base=base, body=attrs, single=None)
+        else:
+            # This is the tricky one.
+            # We need to determine whether the field is describing a valid resource,
+            # and if so, what its corresponding message type is.
+            # Then we need to find the pattern with parameters
+            # that exatcly matches the attrs, if one exists.
```
hzyi-google: exactly

gapic/samplegen/samplegen.py
```diff
@@ -102,6 +104,81 @@ class TransformedRequest:
...
+            resource_typestr = (request_type.
+                                fields[base].
+                                options.
+                                Extensions[resource_pb2.resource_reference].
+                                type)
+
+            resource_message_descriptor = next(
+                (msg.options.Extensions[resource_pb2.resource]
+                 for msg in api_schema.messages.values()
+                 if msg.options.Extensions[resource_pb2.resource].type == resource_typestr),
```
hzyi-google: Note: please also check [file-level annotations](https://github.com/googleapis/googleapis/blob/d9576d95b44f64fb0e3da4760adfc4a24fa1faab/google/api/resource.proto#L35-L39). Some resource type might not have a message.

You should also have a way to instantiate the five common resource names: 
- cloudresourcemanager.googleapis.com/Project
- cloudresourcemanager.googleapis.com/Organization
- cloudresourcemanager.googleapis.com/Folder
- cloudbilling.googleapis.com/BillingAccount
- locations.googleapis.com/Location

Their patterns: https://github.com/googleapis/gapic-generator/blob/master/src/test/java/com/google/api/codegen/testsrc/common/common_resources.proto

hzyi-google: I imagine both of them, the common resource names especially, may take more time to implement. So feel free to defer them to follow-up PRs.

gapic/samplegen/samplegen.py
```diff
@@ -102,6 +104,81 @@ class TransformedRequest:
...
+                raise types.NoSuchResource(
+                    f"No message exists for resource: {resource_typestr}")
+
+            attr_names = [attr.field for attr in attrs]
+
+            # A single resource may be found under multiple paths and have many patterns.
+            # We want to find an _exact_ match, if one exists.
+            pattern = next((p
+                            for p in resource_message_descriptor.pattern
+                            if cls.RESOURCE_RE.findall(p) == attr_names), None)
```
hzyi-google: Would it be possible that `attr_names` has different order than `resource_message_descriptor.pattern`, and the search would find nothing?

software-dov: There's nothing in the samplegen spec about ordering requirements for resource-name based request construction. Allowing it would be tricky but doable; my concern is that there may be multiple valid paths that use the same set of parameters but in different order. I don't know if that's common or even allowed, but it was my concern.

gapic/samplegen/samplegen.py
software-dov: Thanks! That's absolutely something I did not think about.

gapic/samplegen/samplegen.py
software-dov: Doh!

gapic/samplegen/samplegen.py
hzyi-google: Oh are you concerned about there being two patterns like the following at the same time:
- `foos/{foo}/bars/{bar}`
- `bars/{bar}/foos/{foo}`

I would say the possibility that there is such a resource name is very unlikely, and allowing unordered input should prevail that. However I don't think a sample config with reversed order of resource name entities would be very likely either. So feel free to leave the implementation as it is until (if possible at all) someone complains :)

software-dov: Exactly; my feeling as well is to be strict until someone complains :P

gapic/samplegen/samplegen.py
```diff
@@ -102,6 +104,94 @@ class TransformedRequest:
...
+        Raises:
+            NoSuchResource: If the base parameter field for a resource-name
+                            request statement lists a resource_type for which
+                            there is no message with the same resource type.
+            NoSuchResourcePattern: If all the request setup statements for a
+                                   resource name parameter do not combine to
+                                   match a valid path pattern for that resource.
+        """
+
+        if not attrs[0].field:
```
vchudnov-g: not an expert on proto internals, so this might be a silly question: `attrs[0]` always exists?

gapic/samplegen/samplegen.py
```diff
@@ -102,6 +104,94 @@ class TransformedRequest:
...
+    # Resource patterns look something like
+    # kingdom/{kingdom}/phylum/{phylum}/class/{class}
+    RESOURCE_RE = re.compile(r"\{([^}/]+)\}")
+
+    @classmethod
+    def build(cls, request_type: wrappers.MessageType, api_schema, base: str,
+              attrs: List[AttributeRequestSetup], is_resource_request: bool):
+        """Build a TransformedRequest based on parsed input.
+
+        Acts as a factory to hide complicated logic for resource based requests.
```
vchudnov-g: nit: hyphenate "resource-based"

gapic/samplegen/samplegen.py
```diff
@@ -118,15 +208,26 @@ class Validator:
...
-        # The response ($resp) variable is special and guaranteed to exist.
+    CHAIN_LINK_RE = re.compile(
+        r"""
+        (?P<attr_name>\$?\w+)(?:\[(?P<index>\d+)\]|\{["'](?P<key>[^"']+)["']\})?$
+        """.strip())
+
+    VALID_REQUEST_KWORDS = frozenset(
+        ("value", "field", "value_is_file", "input_parameter", "comment"))
+
+    # BUG_dovs: temporary hack, may make the schema a required param.
```
vchudnov-g: Mark as a TODO instead or as well?

gapic/samplegen/samplegen.py
```diff
@@ -118,15 +208,26 @@ class Validator:
     VAL_KWORD = "value"
     BODY_KWORD = "body"
 
-    def __init__(self, method: wrappers.Method):
-        # The response ($resp) variable is special and guaranteed to exist.
+    CHAIN_LINK_RE = re.compile(
```
vchudnov-g: Why this name? Add a brief descriptive comment

software-dov: Think I'll change the name as well; the original impetus behind the name was that it was following the _chain_ of attributes within an expression, but it's not very helpful without that nugget of information.

software-dov: Going with EXPRESSION_ATTR_RE and adding a comment.

gapic/samplegen/samplegen.py
software-dov: Not a silly question, but also not a proto-related question: in `validate_and_transform_request`, a map between an attribute base and a list of AttributeRequestSetup objects gets created; during the construction of the map, every non-error lookup results in an append to the corresponding list. When the TransformedRequests are built, they are built from the key/value pairs in that map, so every attr list is guaranteed to have at least one element.
I'll add a comment to this effect.

gapic/samplegen/samplegen.py
software-dov: Marking as TODO.

gapic/samplegen/samplegen.py
vchudnov-g: Thanks; that's helpful! Could you put the comment right above the regex? It looks like it's describing `VALID_REQUEST_KWORDS` at the moment.

software-dov: Aaah, don't know how that happened.

gapic/samplegen/samplegen.py
software-dov: Filed issues #210 and #209

software-dov:
Open review bump

gapic/samplegen/samplegen.py
```diff
@@ -103,6 +105,98 @@ class TransformedRequest:
...
+        if not attrs[0].field:
+            return cls(base=base, body=None, single=attrs[0])
+        elif not is_resource_request:
+            return cls(base=base, body=attrs, single=None)
+        else:
+            # This is the tricky one.
+            # We need to determine whether the field is describing a valid resource,
+            # and if so, what its corresponding message type is.
+            # Then we need to find the pattern with parameters
+            # that exatcly matches the attrs, if one exists.
```
hzyi-google: typo: "exactly"

software-dov: Good eyesight!

software-dov:
Open review bump

Issue #213, opened 2019-10-10
Handle primitive types in request setup


software-dov:
The code currently doesn't handle primitive message fields.

software-dov:
Fixed by #214

PR #214, opened 2019-10-10
Add test and impl for request messages with primitive fields

software-dov:
Fields that are strings, bools, ints, or floats are permitted.

Mismatch between the type of the assigned value and type of the field
is NOT handled by samplegen yet.

Fix for #213

### DIFF
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -284,7 +284,8 @@ class Validator:
                     # Only valid if this is the last attribute in the chain.
                     break
                 else:
-                    raise TypeError
+                    raise TypeError(
+                        f"Could not handle attribute '{attr_name}' of type: {attr.type}")
 
             if i != len(attr_chain) - 1:
                 # We broke out of the loop after processing an enum or a primitive.
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -280,14 +280,17 @@ class Validator:
                     # and requires less munging of the assigned value
                     duplicate["value"] = f"'{val}'"
                     break
+                elif attr.is_primitive:
+                    # Only valid if this is the last attribute in the chain.
+                    break
                 else:
                     raise TypeError
 
             if i != len(attr_chain) - 1:
-                # We broke out of the loop after processing an enum.
+                # We broke out of the loop after processing an enum or a primitive.
                 extra_attrs = ".".join(attr_chain[i:])
                 raise types.InvalidEnumVariant(
-                    f"Attempted to reference attributes of enum value: '{extra_attrs}'")
+                    f"Attempted to reference attributes of enum value or primitive type: '{extra_attrs}'")
 
             if len(attr_chain) > 1:
                 duplicate["field"] = ".".join(attr_chain[1:])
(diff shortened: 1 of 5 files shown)


gapic/samplegen/samplegen.py
```diff
@@ -280,14 +280,17 @@ def validate_and_transform_request(self,
...
+                    # Only valid if this is the last attribute in the chain.
+                    break
                 else:
                     raise TypeError
 
             if i != len(attr_chain) - 1:
-                # We broke out of the loop after processing an enum.
+                # We broke out of the loop after processing an enum or a primitive.
                 extra_attrs = ".".join(attr_chain[i:])
                 raise types.InvalidEnumVariant(
```
hzyi-google: Let's include something about primitives in this error name?

software-dov: Good catch!

gapic/samplegen/samplegen.py
```diff
@@ -280,14 +280,17 @@ def validate_and_transform_request(self,
...
                     break
+                elif attr.is_primitive:
+                    # Only valid if this is the last attribute in the chain.
+                    break
                 else:
                     raise TypeError
 
             if i != len(attr_chain) - 1:
-                # We broke out of the loop after processing an enum.
+                # We broke out of the loop after processing an enum or a primitive.
```
hzyi-google: I know the other PR to handle resource names is still going, so just a remind here: when the primitive is a string, a `%` can still come after it to specify resource name entities. (It might not be a concern at all depending on how you have implemented resource name construction :) )

software-dov: All this code is lifted up into a method that is called only if the field being set is _not_ a resource name. I figured that this logic was too complicated and involved to mix with resource names, so I just partitioned it off.

tests/unit/samplegen/common_types.py
```diff
@@ -41,7 +41,12 @@
 DummyMessage.__new__.__defaults__ = (False,) * len(DummyMessage._fields)
 
 DummyField = namedtuple("DummyField",
-                        ["message", "enum", "repeated", "field_pb", "meta"])
+                        ["message",
+                         "enum",
+                         "repeated",
+                         "field_pb",
+                         "meta",
+                         "is_primitive"])
```
vchudnov-g: nit: All the other fields are nouns, but this one is a verb. I'd suggest changing it to `primitive` or `primitive_type`.

software-dov: This is one where my hands are mostly tied: it's a duck-type mask over the `wrappers.Field.is_primitive` property. The dummy type exists because in most cases constructing a real `Field` makes the test case bloated.

I could change the name of the real property, but that effects a lot of other code too.

gapic/samplegen/samplegen.py
```diff
@@ -280,14 +280,17 @@ def validate_and_transform_request(self,
                     # and requires less munging of the assigned value
                     duplicate["value"] = f"'{val}'"
                     break
+                elif attr.is_primitive:
+                    # Only valid if this is the last attribute in the chain.
+                    break
                 else:
                     raise TypeError
```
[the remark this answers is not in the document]
software-dov: The tricky bit here is that it's essentially a catch-all for anything wrong in the input or that I haven't thought of. Coming up with a general, helpful error message is going to take some thought, but I'll do my best.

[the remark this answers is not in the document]
software-dov: Came up with something that should hopefully be a lot more helpful.

Issue #217, opened 2019-10-10
lower case manifest file name


vchudnov-g:
Currently, the service name part of the filename is upper-camel-cased, eg `Showcase.v1beta1.python.20191010.223004.manifest.yaml`

software-dov:
Fixed by #220

Issue #216, opened 2019-10-10
quotes not handled correctly in input and output


vchudnov-g:
Default string values are not quoted in the request, and quotes within a string are not handled within a response (and possibly within a request, I haven't checked).

And a spacing issue.

With this config file against the showcase echo service:
```
---
type: com.google.api.codegen.SampleConfigProto
schema_version: 1.2.0
samples:
- service: google.showcase.v1beta1.Echo
  rpc: Echo
  description: A sample of a simple echo request and response
  request:
  - field: content
    comment: The content to be echoed back
    value: a resounding echo...echo...echo
    input_parameter: message
  response:
    - comment:
      - "We simply print the response"
    - print:
      - 'Heard back: "%s"'
      - $resp.content
```

The output (in part) is the following (minus the ##[X], which are my notes referenced below):
```
def sample_echo(message):
    """A sample of a simple echo request and response"""

    client = Echo(
        credentials=credentials.AnonymousCredentials(),
        transport="grpc",
    )

    response = client.echo(
                                                                      ##[1]
      content=a resounding echo...echo...echo)                        ##[2]

    # We simply print the response

    print("Heard back: "{}"".format(response.content))                ##[3]


# [END f39f489e]


def main():
    import argparse

    parser = argparse.ArgumentParser()

    parser.add_argument("--message",
                        type=str,
                        default=a resounding echo...echo...echo)      ##[4]
```

Note:
##[1] Extra blank line. We need to remove it.
##[2] & ##[4] Default value not quoted. Let's quote it. This is release blocker.
##[3] Double quotes `"` within a string not escaped. Let's escape it, just one level (ie no need to try to be sophisticated and do full string parsing to see whether the double quotes in the input are themselves escaped.)

software-dov:
Fixed by #220

PR #220, opened 2019-10-17
Fix for samplegen bugs discovered from the test harness

software-dov:
Includes fixes for #217 , manifest file name is snake case, and #216  , quoting and whitespace fixes

### DIFF
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -15,6 +15,7 @@
 import dataclasses
 import itertools
 import jinja2
+import json
 import keyword
 import os
 import re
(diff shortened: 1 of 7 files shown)


gapic/samplegen/samplegen.py
```diff
@@ -278,7 +279,7 @@ def validate_and_transform_request(self,
                     # Python code can set protobuf enums from strings.
                     # This is preferable to adding the necessary import statement
                     # and requires less munging of the assigned value
-                    duplicate["value"] = f"'{val}'"
+                    # duplicate["value"] = f"'{val}'"
```
hzyi-google: what's up with this change...?

software-dov: Ah, that's cruft that needs to be removed. The same thing is now handled for all string values with the 

```python
if isinstance(duplicate["value"], str):
```

software-dov:
Doh! Those two are close enough together and far enough away from the home row that my frozen fingers must have gotten confused.

PR #226, opened 2019-10-27
[fix] Better path handling in gapic.sh

alexander-fenster:
This PR fixes a two minor but annoying issues with paths handling in `gapic.sh`.

1. If `--path` is given, the check for `--in` existence should respect it.
2. `docker` requires paths for mount points to be absolute, so let's convert both `--in` and `--out` to absolute paths before running `docker`.

### DIFF
--- a/gapic.sh
+++ b/gapic.sh
@@ -64,8 +64,8 @@ if [ -z "$IMAGE" ] || [ -z "$IN" ] || [ -z "$OUT" ]; then
 fi
 
 # Ensure that the input directory exists (and is a directory).
-if ! [ -d $IN ]; then
-  >&2 echo "Directory does not exist: $IN"
+if ! [ -d "${PROTO_PATH}/$IN" ]; then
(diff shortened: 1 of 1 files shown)


gapic.sh
```diff
@@ -85,10 +85,19 @@ if [ "$(ls -A $OUT )"]; then
...
 fi
 
+# Convert IN and OUT to absolute paths for Docker
+CWD=`pwd`
+cd ${PROTO_PATH}/$IN
+ABS_IN=`pwd`
+cd $CWD
+cd $OUT
+ABS_OUT=`pwd`
+cd $CWD
```
software-dov: Outrageous nit: why not do
```bash
ABS_IN=$(realpath ${PROTO_PATH}/${IN})
ABS_OUT=$(realpath ${OUT})
```
?

alexander-fenster: There is no `realpath` (or `readlink -f`) on OS X (at least by default). The current script runs without any problems on OS X and I want to keep it that way.

software-dov: Aha! Good to know.

PR #234, opened 2019-10-31
[feat] Retries

lukesneeringer:
This PR implements reading a gRPC service config and populating the appropriate retry behavior.

### DIFF
--- a/gapic/schema/api.py
+++ b/gapic/schema/api.py
@@ -572,14 +572,14 @@ class _ProtoBuilder:
                         timeout = self._to_float(mc['timeout'])
                     r = mc.get('retryPolicy', {})
                     retry = wrappers.RetryInfo(
-                        max_attempts=r.get('maxAttempts', 5),
+                        max_attempts=r.get('maxAttempts', 0),
                         initial_backoff=self._to_float(
-                            r.get('initialBackoff', '1s'),
+                            r.get('initialBackoff', '0s'),
                         ),
                         max_backoff=self._to_float(
-                            r.get('maxBackoff', '60s'),
+                            r.get('maxBackoff', '0s'),
                         ),
-                        backoff_multiplier=r.get('backoffMultiplier', 1.5),
+                        backoff_multiplier=r.get('backoffMultiplier', 0.0),
                         retryable_exceptions=tuple([
                             exceptions.exception_class_for_grpc_status(
                                 getattr(grpc.StatusCode, code),
--- a/gapic/schema/api.py
+++ b/gapic/schema/api.py
@@ -23,8 +23,11 @@ import sys
 from itertools import chain
 from typing import Callable, Dict, FrozenSet, Mapping, Sequence, Set, Tuple
 
+from google.api_core import exceptions
 from google.longrunning import operations_pb2  # type: ignore
 from google.protobuf import descriptor_pb2
+
+import grpc
 
 from gapic.generator import options
 from gapic.schema import metadata
@@ -208,6 +211,7 @@ class API:
                 file_descriptor=fd,
                 file_to_generate=fd.package.startswith(package),
                 naming=naming,
+                opts=opts,
                 prior_protos=protos,
             ).proto
 
@@ -294,6 +298,7 @@ class _ProtoBuilder:
     def __init__(self, file_descriptor: descriptor_pb2.FileDescriptorProto,
                  file_to_generate: bool,
                  naming: api_naming.Naming,
+                 opts: options.Options = options.Options(),
                  prior_protos: Mapping[str, Proto] = None):
         self.proto_messages: Dict[str, wrappers.MessageType] = {}
         self.proto_enums: Dict[str, wrappers.EnumType] = {}
@@ -301,6 +306,7 @@ class _ProtoBuilder:
         self.file_descriptor = file_descriptor
         self.file_to_generate = file_to_generate
         self.prior_protos = prior_protos or {}
+        self.opts = opts
 
         # Iterate over the documentation and place it into a dictionary.
         #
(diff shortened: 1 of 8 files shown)


gapic/schema/api.py
```diff
@@ -526,24 +532,73 @@ def _get_methods(self,
...
+                        max_backoff=self._to_float(
+                            r.get('maxBackoff', '0s'),
+                        ),
+                        backoff_multiplier=r.get('backoffMultiplier', 0.0),
+                        retryable_exceptions=frozenset([
+                            exceptions.exception_class_for_grpc_status(
+                                getattr(grpc.StatusCode, code),
+                            )
+                            for code in r.get('retryableStatusCodes', [])
+                        ]),
```
software-dov: Nit: all the builtin dataclass ctors can just take generator comprehensions, no need for the square brackets.

tests/unit/schema/test_api.py
```diff
@@ -344,6 +346,23 @@ def test_messages_nested():
 def test_services():
     L = descriptor_pb2.SourceCodeInfo.Location
 
+    # Set up retry information.
+    def _n(method_name: str):
+        return {
+            'service': 'google.example.v2.FooService',
+            'method': method_name,
+        }
```
software-dov: Nit: extra newline please

tests/unit/schema/test_api.py
```diff
@@ -393,6 +424,21 @@ def test_services():
     assert method.input.meta.doc == 'This is the GetFooRequest message.'
     assert method.output.name == 'GetFooResponse'
     assert method.output.meta.doc == 'This is the GetFooResponse message.'
+    assert method.timeout - 30.0 < 5e-10
```
software-dov: Nit: why not lift the `- x < 5e-19` into an approx_eq function? Makes it more correct, too.

```python
def approx_eq(x, y, epsilon=5e-19):
   return abs(x - y) < epsilon
```

tests/unit/schema/test_api.py
```diff
@@ -393,6 +424,21 @@ def test_services():
...
+    assert not method.retry.max_attempts
+    assert not method.retry.initial_backoff
+    assert not method.retry.max_backoff
+    assert not method.retry.backoff_multiplier
+    assert not method.retry.retryable_exceptions
+    retry_method = service.methods['RetryableGetFoo']
```
software-dov: Nit: newline, please.

gapic/schema/api.py
```diff
@@ -526,24 +532,73 @@ def _get_methods(self,
...
+                    r = mc.get('retryPolicy', {})
+                    retry = wrappers.RetryInfo(
+                        max_attempts=r.get('maxAttempts', 0),
+                        initial_backoff=self._to_float(
+                            r.get('initialBackoff', '0s'),
+                        ),
+                        max_backoff=self._to_float(
+                            r.get('maxBackoff', '0s'),
+                        ),
+                        backoff_multiplier=r.get('backoffMultiplier', 0.0),
```
software-dov: Outside the scope of the review, but does this imply that the default values are provided by the default generated service config?

gapic/schema/api.py
```diff
@@ -526,24 +532,73 @@ def _get_methods(self,
...
+                    # Does this method config apply to us?
+                    # The gRPC service config uses a repeated `name` field
+                    # with a particular format, which we match against.
+                    selector = {
+                        'service': '{package}.{service_name}'.format(
+                            package='.'.join(service_address.package),
+                            service_name=service_address.name,
+                        ),
+                        'method': meth_pb.name,
+                    }
```
software-dov: This structure should be lifted up; if you stick with the current control flow, it's a loop invariant, and if you change it, it's a data dependency for the `next` expression.

gapic/schema/api.py
```diff
@@ -526,24 +532,73 @@ def _get_methods(self,
...
+            timeout = None
+            if self.opts.retry:
+                # This object should be a dictionary that conforms to the
+                # gRPC service config proto:
+                #   Repo: https://github.com/grpc/grpc-proto/
+                #   Filename: grpc/service_config/service_config.proto
+                #
+                # We only care about a small piece, so we are just leaving
+                # it as a dictionary and parsing accordingly.
+                for mc in self.opts.retry.get('methodConfig', []):
```
software-dov: This confused me a bit at first; the structure of the loop implies that there might be more than one match.

This is a little more clear in terms of data and control flow IMO, plus the comment:
```python
# Search for the method config for this method, if one exists.
mc = next((mc
           for mc in self.opts.retry.get('methodConfig', [])
           if selector in mc.get('name')), None)
if mc:
   ...
```

gapic/schema/api.py
```diff
@@ -659,3 +714,9 @@ def _load_service(self,
             service_pb=service,
         )
         return self.proto_services[address.proto]
+
+    def _to_float(self, s: str) -> float:
+        """Convert a protobuf duration string (e.g. `"30s"`) to float."""
+        if s.endswith('n'):
+            return int(s[:-1]) / 1e9
+        return float(s[:-1])
```
software-dov: Outrageous personal preference nit:
```python
return (int(s[:-1]) / 1e9) if s.endswith('n') else float(s[:-1])
```

gapic/schema/api.py
lukesneeringer: The template declines to use zero values. In theory, all of these fields are required, but I do not trust that to actually happen.

tests/unit/schema/test_api.py
lukesneeringer: Replacing with `pytest.approx`. I meant to look up the syntax and forgot.

gapic/schema/api.py
lukesneeringer: Done.

tests/unit/schema/test_api.py
lukesneeringer: Done, and added comments. (I have a hard rule that a comment must follow any blank line within a function.)

gapic/schema/api.py
lukesneeringer: Done, and added a better comment.

gapic/schema/api.py
lukesneeringer: Done, but this changed coverage slightly, so now the test has three methods instead of two. This is better anyway.

gapic/schema/api.py
lukesneeringer: Done.

tests/unit/schema/test_api.py
software-dov: Nice rule! Think I'll steal it.

PR #239, opened 2019-11-06
Fixes for reading service config

software-dov:
Adds fixes and a unit test for reading a service config from file.
Includes minor rendering fixes.

System tests to verify that the config percolates to method calls are in progress.

### DIFF
--- a/gapic/generator/options.py
+++ b/gapic/generator/options.py
@@ -92,12 +92,18 @@ class Options:
                 os.path.join(os.path.dirname(__file__), '..', 'templates'),
             )
 
+        retry_cfg = None
+        retry_path = opts.pop(cls.RETRY_OPT, None)
+        if retry_path:
+            with open(retry_path[-1]) as f:
+              retry_cfg = json.load(f)
+
         # Build the options instance.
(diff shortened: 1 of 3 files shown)


software-dov:
Open review bump

gapic/generator/options.py
```diff
@@ -92,12 +92,19 @@ def build(cls, opt_string: str) -> 'Options':
                 os.path.join(os.path.dirname(__file__), '..', 'templates'),
             )
 
+        retry_cfg = None
+        retry_paths = opts.pop(cls.RETRY_OPT, None)
```
hzyi-google: Can we just make `cls.RETRY_OPT` point to a single file?

software-dov: The way that the option string is parsed makes it tricky to do that. It could be done, but it would require special casing `cls.RETRY_OPT` or making a general solution that other options do not pick by default.

hzyi-google: That sounds reasonable.

software-dov: I would lean towards keeping the parsing simple and general. If there are ever more options that should only take a single value, generalizing the parsing can be revisited.

Issue #241, opened 2019-11-13
Revert client method interface and flattening to be isomorphic to the monolithic surface


software-dov:
The monolithic generator flattens client methods like so: required fields become positional parameters, optional fields become optional parameters with a default value of `None`.

This interface will be preserved to preserve backwards compatibility with the generated surfaces.

software-dov:
Fixed by #245

PR #245, opened 2019-11-14
Adopt legacy method flattening semantics

software-dov:
Tests and implementation for #241
The python microgenerator transition consensus is to minimize breaking changes for the duration of the transition and revisit any breaking UI/UX improvements after the fact. As a result, the legacy flattening interface must be restored.

Inspection of the monolith parameter transformer and template reveal that the legacy interface only flattens the fields of the request object one level deep; required fields are used first as positional parameters, followed by optional fields used as optional parameters.

Includes changes to the client template, client unit test template, and new features and associated tests in the generator logic.

Note: please ignore most of the formatting changes. They are due to an overly aggressive emacs configuration that is difficult to change.

The relevant source lines in the generator are https://github.com/googleapis/gapic-generator/blob/bcedba65bf930d3e35530fe5360f1c6f24d27abc/src/main/resources/com/google/api/codegen/py/main.snip#L321 and https://github.com/googleapis/gapic-generator/blob/bcedba65bf930d3e35530fe5360f1c6f24d27abc/src/main/java/com/google/api/codegen/transformer/py/PythonApiMethodParamTransformer.java#L43 .

### DIFF
--- a/gapic/schema/wrappers.py
+++ b/gapic/schema/wrappers.py
@@ -594,8 +594,7 @@ class Method:
         # We found no repeated fields. Return None.
         return None
 
-    @utils.cached_property
-    def ref_types(self) -> Sequence[Union[MessageType, EnumType]]:
+    def _ref_types(self, use_legacy: bool = False) -> Sequence[Union[MessageType, EnumType]]:
         """Return types referenced by this method."""
         # Begin with the input (request) and output (response) messages.
         answer = [self.input]
@@ -607,7 +606,8 @@ class Method:
         #
         # This entails adding the module for any field on the signature
         # unless the field is a primitive.
-        for field in self.flattened_fields.values():
+        flattening = self.legacy_flattened_fields if use_legacy else self.flattened_fields
+        for field in flattening.values():
             if field.message or field.enum:
                 answer.append(field.type)
 
@@ -619,6 +619,14 @@ class Method:
 
         # Done; return the answer.
         return tuple(answer)
+
+    @utils.cached_property
+    def ref_types(self) -> Sequence[Union[MessageType, EnumType]]:
+        return self._ref_types()
+
+    @utils.cached_property
+    def ref_types_legacy(self) -> Sequence[Union[MessageType, EnumType]]:
+        return self._ref_types(use_legacy=True)
 
     @property
     def void(self) -> bool:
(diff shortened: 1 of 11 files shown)


gapic/schema/wrappers.py
```diff
@@ -540,6 +543,15 @@ def flattened_fields(self) -> Mapping[str, Field]:
         # Done; return the flattened fields
         return answer
 
+    @utils.cached_property
+    def legacy_flattened_fields(self) -> Mapping[str, Field]:
+        """Return the legacy flattening interface: top level fields only,
+        required fields first"""
+        required, optional = utils.partition(lambda f: f.required,
```
hzyi-google: Is this _the only_ method generated from the rpc (or, does Python ignore `method_signature` annotation`)?

Also, does this mean changing a field from required to optional may cause breaking change?

software-dov: Python has no built in concept of function overloads. It ignores all `method_signature` annotations.

Yes, a field changing from required to optional may cause a breaking change in the client surface. To the best of my understanding, this is true for the monolith as well and was one of the reasons @lukesneeringer changed the method interface semantics.

hzyi-google: Okay cool. I asked this mainly because of a thread in the chat (I'll cc you). I think the code LGTM if the goal is to preserve the same behavior the monolithic does. But the potential breaking change may be something worth considering for the next major version.

software-dov: The plan is to transition to the microgenerator by making as few breaking changes as possible. Once the transition is complete we can revisit interface improvements.

lukesneeringer:
I am nervous about this change. The legacy flattening semantics violate several provisions of [the spec](https://aip.dev/4232). In particular, they make it so that adding a `google.api.method_signature` when there was not one before becomes a breaking change. **This is a really big deal.**

I understand the desire to make the transition easier, but we really _really_ need to hold the principle of "adding configuration is non-breaking" as something that we do not violate.

software-dov:
@lukesneeringer I completely agree. This is a change that I was fighting against, but other people convinced me that smooth and speedy adoption of the microgenerator was the more pressing concern. Once that is completed I will be Putting the Foot Down about a better interface.

Issue #227, opened 2019-10-29
Add support for client_options


busunkim96:
Client constructors should accept `client_options`. See https://googleapis.dev/python/google-api-core/1.14.3/client_options.html 

Currently `api_endpoint` is the only supported option, but this will grow in the future.

software-dov:
Implemented by #248

PR #248, opened 2019-11-18
Generated test and impl for client options

software-dov:
In order to preserve the interface of legacy client construction,
client constructors take a
'google.api_core.client_options.ClientOptions' parameter.

Currently, this option only customizes the api endpoint, but other
customizations may be enabled.

Includes generated unit tests.

Implementation for #227

### DIFF
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
@@ -99,12 +99,12 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
                 automatically.
             client_options (ClientOptions): Custom options for the client.
         """
-        # Save or instantiate the transport.
-        # Ordinarily, we provide the transport, but allowing a custom transport
-        # instance provides an extensibility point for unusual situations.
         if isinstance(client_options, dict):
             client_options = ClientOptions.from_dict(client_options)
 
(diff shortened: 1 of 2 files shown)


gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -79,15 +82,13 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
 
 
     def __init__(self, *,
-            host: str{% if service.host %} = '{{ service.host }}'{% endif %},
```
lukesneeringer: This change is a bit scary. We may have situations where the `host` may not be specified (for instance, APIs with host-based regionalization). I made the original change to make overriding `host` not feel like a gymnastics exercise.

Also, the previous code had `host` as a required argument if we did not have a sensible default, but this change loses that.

software-dov: Again, preserving the legacy interface is preventing us from adopting better semantics.
@busunkim96 indicated that more configuration may be added to `client_options` in the future, and having the data structure augmented is arguably better than bloating the constructor signature.
I didn't know about APIs being able to not set a host. The new change should still catch that, no? If host is a required parameter, static analysis can detect its absence, but the transport will still raise if it isn't given a host. It's a minor UI wart, but the user is till prevented from constructing a client sans host.

PR #250, opened 2019-11-23
First pass template for a keyword cst transformer utility

software-dov:
Manual testing on samples insert keywords correctly, preserve comments, and (mostly) maintain formatting. 

TODO:
* Be better at preserving whitespace and formatting
* Remove the space between introduced keywords, the =, and the value
* Add a dry-run option
* Add some tests for the scripts
* Package them properly in the generated setup.py


software-dov:
Open review bump

vchudnov-g:
Could you amend the PR description to provide context as to why you're doing this? What's the problem you're solving?

software-dov:
@vchudnov-g This PR adds an auto-conversion script to ease the transition to the microgenerator method flattening semantics. Client calls to API methods are rewritten to pass flattened parameters by keyword.

This _specific_ PR will be closed because the proposed microgenerator flattening semantics have changed again; once the original flattening semantics are back, I'll make a new PR that generates a conversion script per API, which will probably look very similar to this one.

Issue #242, opened 2019-11-13
Add helper functions for resource path construction


software-dov:
The monolith currently generates helper methods to allow users to construct paths for methods that consume references.

E.g. if a request takes a `name` parameter, and that name refers to a resource that lives in the path
"kingdom/{kingdom}/phylum/{phylum}", the client generates a method that looks something like

```python
@staticmethod
def phylum_path(kingdom, phylum):
    return "kingdom/{kingdom}/phylum/{phylum}".format(kingdom=kingdom, phylum=phylum)
```

The microgenerator clients need corresponding resource helper methods for backwards compatibility.

software-dov:
Implemented by #253

Issue #210, opened 2019-10-09
Support the five common resource names in generated samples


software-dov:
https://github.com/googleapis/gapic-generator/blob/master/src/test/java/com/google/api/codegen/testsrc/common/common_resources.proto

PR #253, opened 2019-11-28
Tests and impl for resource path helper methods

software-dov:
Includes generated unit tests, plumbing code unit tests, and system
tests for determining whether a message is a resource, providing an
accessor for one and only one path if it is a resource, and a service
accessor for all message fields that are resources (only one level
deep at the moment).

Impl for #242

### DIFF
--- a/gapic/schema/wrappers.py
+++ b/gapic/schema/wrappers.py
@@ -37,6 +37,7 @@ from typing import (cast, Dict, FrozenSet, List, Mapping, Optional,
 from google.api import annotations_pb2  # type: ignore
 from google.api import client_pb2
 from google.api import field_behavior_pb2
+from google.api import resource_pb2
 from google.api_core import exceptions  # type: ignore
 from google.protobuf import descriptor_pb2
 
@@ -211,6 +212,9 @@ class MessageType:
     def __getattr__(self, name):
         return getattr(self.message_pb, name)
 
+    def __hash__(self):
+        return hash(self.name)
+
     @utils.cached_property
     def field_types(self) -> Sequence[Union['MessageType', 'EnumType']]:
         """Return all composite fields used in this proto's messages."""
@@ -229,6 +233,20 @@ class MessageType:
     def ident(self) -> metadata.Address:
         """Return the identifier data to be used in templates."""
         return self.meta.address
+
+    @property
+    def resource_path(self) -> Optional[str]:
+        """If this message describes a resource, return the path to the resource.
+        If there are multiple paths, returns the first one."""
+        return next(
+            iter(self.options.Extensions[resource_pb2.resource].pattern),
+            None
+        )
+
+    @property
+    def resource_path_args(self) -> Sequence[str]:
+        path_arg_re = re.compile(r'\{([a-zA-Z0-9_-]+)\}')
+        return path_arg_re.findall(self.resource_path or '')
 
     def get_field(self, *field_path: str,
                   collisions: FrozenSet[str] = frozenset()) -> Field:
@@ -730,6 +748,17 @@ class Service:
 
         # Done; return the answer.
         return frozenset(answer)
+
+    @utils.cached_property
+    def resource_messages(self) -> FrozenSet[MessageType]:
+        """Returns all the resource message types used in all
+        request fields in the service."""
+        return frozenset(
+            field.message
+            for method in self.methods.values()
+            for field in method.input.fields.values()
+            if field.message and field.message.resource_path
+        )
 
     def with_context(self, *, collisions: FrozenSet[str]) -> 'Service':
         """Return a derivative of this service with the provided context.
--- a/gapic/schema/wrappers.py
+++ b/gapic/schema/wrappers.py
@@ -244,6 +244,11 @@ class MessageType:
         )
 
     @property
+    def resource_type(self) -> Optional[str]:
+        resource = self.options.Extensions[resource_pb2.resource]
+        return resource.type[resource.type.find('/') + 1:] if resource else None
+
+    @property
     def resource_path_args(self) -> Sequence[str]:
         path_arg_re = re.compile(r'\{([a-zA-Z0-9_-]+)\}')
         return path_arg_re.findall(self.resource_path or '')
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
@@ -83,8 +83,8 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
 
     {% for message in service.resource_messages -%}
     @staticmethod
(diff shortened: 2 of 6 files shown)


software-dov:
Open review bump

tests/unit/schema/wrappers/test_message.py
```diff
@@ -124,6 +125,18 @@ def test_get_field_nonterminal_repeated_error():
         assert outer.get_field('inner', 'one') == inner_fields[1]
 
 
+def test_resource_path():
+    options = descriptor_pb2.MessageOptions()
+    patterns = options.Extensions[resource_pb2.resource].pattern
+    patterns.append("kingdom/{kingdom}/phylum/{phylum}/class/{klass}")
```
hzyi-google: nit: let's change the static parts to plurals to be consistent with (most) real APIs.

gapic/schema/wrappers.py
```diff
@@ -230,6 +234,20 @@ def ident(self) -> metadata.Address:
         """Return the identifier data to be used in templates."""
         return self.meta.address
 
+    @property
+    def resource_path(self) -> Optional[str]:
+        """If this message describes a resource, return the path to the resource.
+        If there are multiple paths, returns the first one."""
```
hzyi-google: How about other patterns? Do we not generate helpers for them?

gapic/schema/wrappers.py
```diff
@@ -731,6 +749,17 @@ def names(self) -> FrozenSet[str]:
...
 
+    @utils.cached_property
+    def resource_messages(self) -> FrozenSet[MessageType]:
+        """Returns all the resource message types used in all
+        request fields in the service."""
+        return frozenset(
+            field.message
+            for method in self.methods.values()
+            for field in method.input.fields.values()
+            if field.message and field.message.resource_path
```
hzyi-google: You might have missed the file-level annotations (`google.api.resource_definition`).

Optional: consider filter out resources defined but not referenced. The monolith does that; If your goal is to maintain backward-compatibility it might be worth keeping the behavior. OTOH, the logic seems too implicit, I am also fine if you drop it until we found something broken :p

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -81,6 +81,13 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
     from_service_account_json = from_service_account_file
 
 
+    {% for message in service.resource_messages -%}
+    @staticmethod
+    def {{ message.name|snake_case }}_path({% for arg in message.resource_path_args %}{{ arg }}: str,{% endfor %}) -> str:
```
hzyi-google: I think using `(google.api.resource).type` to generate the helper method name is better - It is designed to be the identifier of the resources. And that works for file-level `resource_definition` annotation too.

software-dov:
They are being added back in the name of backwards compatibility and user friendliness :package:

tests/unit/schema/wrappers/test_message.py
software-dov: Done.

gapic/schema/wrappers.py
software-dov: What other patterns does the client know or care about? A cursory perusal of the monolithic templates leads me to the (tentative) conclusion that the helper methods are for describing paths in request fields.

gapic/schema/wrappers.py
software-dov: Good catch! I did miss file-level annotations.
Where should those helper methods live? If they belong to the API but not to any particular service, there isn't an obvious client module where they should belong. Should they just be part of the base module? That would reduce duplication between multiple services that reference the same resource, but on the other hand the whole point of adding these methods back is backwards compatibility; if I'm reading the monolithic template correctly, all that's currently created are per-client-class helper methods.

gapic/schema/wrappers.py
hzyi-google: This falls into the multi-pattern resource name black hole...

Gapic yaml has an entity name for every resource pattern, and the entity name is used to generate the name of the helper method.

In annotations however, we only have an identifier for every resource. For single-pattern resources we are fine because the resource is the pattern. However, for multi-pattern resources the generators don't know what the resource name helpers for each pattern should be named because we don't have that information.

Some design work might be needed for multi-pattern resources in Python. I was under the impression of discontinuing the resource name helpers in the micro generator so I kinda skipped the design of that in Python. It might be reasonable to take the same approach as Ruby and Node do, described in https://docs.google.com/document/d/1nuh6yrPubl1XMWH2WQHtmOwk_s-0MZEgSoFD6RbMrxg/edit#heading=h.a3gbsu9p63o

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
software-dov: Done

gapic/schema/wrappers.py
software-dov: Ahh, multipattern resources. Got it.

The consensus was to take a breaking change and not support multipattern resources. It's just easier not to :P

gapic/schema/wrappers.py
hzyi-google: Yeah that's a good point. I would say check where the resource is referenced, but a quick search in googleapis doesn't give me any API that does so, and it's really hard to reason without one.

Now I realized that the five common resources should also be supported. In this case, the helper methods should be generated if they are referenced in any messages that are either the input, output or fields of input or output of any method in a service. Given this requirement, I think treating the file-level annotations the same way is probably the best solution.

software-dov: Okay, so file-level resources just get added to a client if the client refers to that resource in any way. I can live with that (and so will the users).
I think I may be missing something, though: the Python `google.api.resource_pb2` module doesn't have a `resource_definition` member.

hzyi-google: Oh, you probably need a new version of Python common protos. It's probably released manually? @lukesneeringer

gapic/schema/wrappers.py
hzyi-google: ... well okay then.

gapic/schema/wrappers.py
software-dov: Are you okay with postponing the five common resources and file level resources? The first is already being tracked by #210, and I can file an issue for the second.

gapic/schema/wrappers.py
software-dov: Dynamic languages make the issue worse because there isn't a good mechanism to have method overloading. Without method overloading, we would need some kind of name disambiguation, and moving away from the gapic config makes that more difficult. Given the small number of multi-pattern resources, the consensus was "screw it."

gapic/schema/wrappers.py
hzyi-google: SGTM

hzyi-google:
LGTM

Issue #169, opened 2019-08-07
Unused imports 


busunkim96:
Flake raises warnings about unused imports.

See https://github.com/googleapis/google-cloud-python-private/pull/504
```
google/cloud/workflows_v1alpha1/services/workflows/pagers.py:2:1: F401 'copy' imported but unused
google/cloud/workflows_v1alpha1/services/workflows/client.py:3:1: F401 'typing.Mapping' imported but unused
google/cloud/workflows_v1alpha1/services/workflows/client.py:3:1: F401 'typing.Optional' imported but unused
google/cloud/workflows_v1alpha1/services/workflows/transports/grpc.py:2:1: F401 'typing.Sequence' imported but unused
google/cloud/workflows_v1alpha1/services/workflows/transports/grpc.py:2:1: F401 'typing.Tuple' imported but unused
tests/unit/workflows_v1alpha1/test_workflows.py:10:1: F401 'google.api_core.operation' imported but unused
tests/unit/workflows_v1alpha1/test_workflows.py:18:1: F401 'google.protobuf.empty_pb2 as empty' imported but unused
```

lukesneeringer:
I added a PR (#177) for the easy ones. The two in the unit tests are actually somewhat difficult, but still worth trying to bop out.

software-dov:
For the two in `test_workflows.py`, I don't see the corresponding imports in the `test_$service.py.j2`. Am I misreading the template? Has the issue been fixed and there's now just version shear?

lukesneeringer:
It is rendered by the last loop, I believe.

software-dov:
Last bit fixed by #254

PR #254, opened 2019-12-02
Fix for #169: unused imports

software-dov:
Fix the unused import of api_core.operation and protobuf.empty_pb2 in
generated unit tests.

The fix is hacky, gross, and specific, but presumably it should generalize.


software-dov:
Open review bump

PR #256, opened 2019-12-06
Re-add original flattening semantics

software-dov:
Re-add the original method flattening semantics while lifting the restriction that flattened parameters be primitive types.

### DIFF
--- a/gapic/schema/wrappers.py
+++ b/gapic/schema/wrappers.py
@@ -545,26 +545,14 @@ class Method:
     @utils.cached_property
     def flattened_fields(self) -> Mapping[str, Field]:
         """Return the signature defined for this method."""
-        answer: Dict[str, Field] = collections.OrderedDict()
         signatures = self.options.Extensions[client_pb2.method_signature]
 
-        # Iterate over each signature and add the appropriate fields.
-        for sig in signatures:
-            # Get all of the individual fields.
-            fields = collections.OrderedDict([
-                (f.strip(), self.input.get_field(*f.strip().split('.')))
-                for f in sig.split(',')
-            ])
+        answer: Dict[str, Field] = collections.OrderedDict(
+            (f.strip(), self.input.get_field(*f.strip().split('.')))
+            for sig in signatures
+            for f in sig.split(',')
+        )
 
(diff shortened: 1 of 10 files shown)


gapic/schema/wrappers.py
```diff
@@ -545,26 +545,14 @@ def field_headers(self) -> Sequence[str]:
...
-            ])
-
-            # Sanity check: If any fields contain a message, we ignore the
-            # entire signature.
-            if any([i.message for i in fields.values()]):
-                continue
-
-            # Add the fields to the answer.
-            answer.update(fields)
+        answer: Dict[str, Field] = collections.OrderedDict(
```
hzyi-google: Having a hard time understanding this. If there are multiple `method_signature`s, will `answer` be a flat map of all the fields in all signatures...?

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -164,13 +161,20 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
         {%- endif %}
         """
         # Create or coerce a protobuf request object.
-        {% if method.legacy_flattened_fields -%}
+        {% if method.flattened_fields -%}
+        # Sanity check: If we got a request object, we should *not* have
+        # gotten any keyword arguments that map to the request.
```
hzyi-google: Do you intend to render this comment in generated gapic client?

software-dov: Yes, this comment is intended to be rendered in the generated client logic.

gapic/schema/wrappers.py
software-dov: Yes. Because python doesn't let us overload methods, we have to combine the API method signatures in the GAPIC client method signature somehow.
It's then the job of the documentation and samples to describe which combinations of flattened parameters are valid and the job of the backend service to generate errors when invalid combinations get passed.

software-dov: Cleared up AFK.

PR #258, opened 2019-12-10
Add template for a flattening conversion script

software-dov:
The generated method flattening semantics change between the the
original monolithic GAPIC generator and this GAPIC generator (aka the
Python Microgenerator). User code that calls methods from GAPICs
generated by the monolith will need to be converted to use the new
flattening semantics.

This change describes a template for an automated per-service
conversion script.

Cases the conversion script will NOT handle correctly include
* Dynamic method dispatch
* Method calls via an alias or as a free function
* Star arg exapnsion in a method call (*args or **kwargs)

Plain, normal, vanilla calls are the target for conversion.

### DIFF
--- a/gapic/templates/scripts/fixup_%service_keywords.py.j2
+++ b/gapic/templates/scripts/fixup_%service_keywords.py.j2
@@ -1,0 +1,125 @@
+{% extends '_base.py.j2' %}
+{% block content %}
+import argparse
+import os
+import libcst as cst
+from typing import (Any, Callable, Dict, List, Sequence, Tuple)
+
+
+def partition(
+    predicate: Callable[[Any], bool],
+    iterator: Sequence[Any]
+) -> Tuple[List[Any], List[Any]]:
+    """A stable, out-of-place partition."""
+    results = ([], [])
+
+    for i in iterator:
+        results[int(predicate(i))].append(i)
+
+    # Returns trueList, falseList
+    return results[1], results[0]
+
+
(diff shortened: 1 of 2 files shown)


gapic/templates/scripts/fixup_%service_keywords.py.j2
```diff
@@ -0,0 +1,125 @@
...
+class {{ service.client_name }}CallTransformer(cst.CSTTransformer):
+    CTRL_PARAMS: Tuple[str] = ('retry', 'timeout', 'metadata')
+
+    METHOD_TO_PARAMS: Dict[str, Tuple[str]] = {
+        {% for method in service.methods.values() -%}
+          '{{ method.name|snake_case }}': ({% for field in method.legacy_flattened_fields.values() %}'{{ field.name }}', {% endfor %}),
+        {% endfor -%}
+    }
+
+    def leave_Call(self, original: cst.Call, updated: cst.Call) -> cst.CSTNode:
```
hzyi-google: The casing is a bit weird. Did you mean `leaveCall` or `leave_call`?

software-dov: I mean leave_Call. It's the naming from libCST, and this is a method override as a child class of cst.CSTVisitor.

https://libcst.readthedocs.io/en/latest/visitors.html#visit-and-leave-helper-functions
https://libcst.readthedocs.io/en/latest/tutorial.html#Build-Visitor-or-Transformer

Issue #266, opened 2020-01-09
Provide automatic conversion of type and enum imports


busunkim96:
It would be nice if the conversion could be done automatically or if enums and types could continue to be available at the previous location `google.cloud.texttospeech.types` and `google.cloud.texttospeech.enums`.

Monocode
```py
from google.cloud import texttospeech
 
voice = texttospeech.types.VoiceSelectionParams(
        language_code='en-US',
        ssml_gender=texttospeech.enums.SsmlVoiceGender.NEUTRAL)
```

Microcode
```py
from google.cloud import texttospeech

voice = texttospeech.VoiceSelectionParams(
        language_code='en-US',
        ssml_gender=texttospeech.SsmlVoiceGender.NEUTRAL)
```

[This PR](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2698/files) has more examples.

lukesneeringer:
I thought I made the enums available on `texttospeech.SsmvlVoiceGender.NEUTRAL`?

busunkim96:
@lukesneeringer Whoops, I copied-pasted the same code twice. :sweat_smile: It's fixed now.

In the microgenerator the enums and types are available under `texttospeech` which is much nicer than having to do `texttospeech.enums` and `texttospeech.types`.  But when folks are switching over they'll have to manually change the import paths. 

![image](https://user-images.githubusercontent.com/8822365/72190493-8105eb00-33b4-11ea-8620-977d9398966c.png)


I was wondering if it would be technically feasible to either:

1. Provide a script to fix the imports automatically.
2. Continue to provide enums and types at `texttospeech.enums` and `texttospeech.types`.

software-dov:
I'm not sure I fully understand the description: is it that
```python
from google.cloud import texttospeech

ssml_gender = texttospeech.enums.SsmlVoiceGender.NEUTRAL
```
is no longer valid because there is no enum submodule in `texttospeech`?

busunkim96:
@software-dov Yep, that's it.

software-dov:
Thanks for clarifying. I'll get right on that.

software-dov:
Technically a bugfix in that it's a gratuitous delta in the exposed surface.

software-dov:
Fixed by #273

PR #273, opened 2020-01-25
Attempt at providing 'types' and 'enums' submodule view

software-dov:
Tentative fix for #266 
This is almost certainly not quite right, but it's a starting point.
Testing is WIP.


busunkim96:
@software-dov Once this is merged would you be able to cut a release? I think this was the last of the issues I filed after converting the TTS quickstart. Thank you. :)

software-dov:
Sure, just need to fight with unit-test coverage real quick.

PR #282, opened 2020-02-05
add test cases for interceptor

arithmetic1728:


tests/system/conftest.py
```diff
@@ -34,3 +35,81 @@ def identity():
...
+                                grpc.StreamStreamClientInterceptor):
+
+    def __init__(self, interceptor_function):
+        self._fn = interceptor_function
+
+    def intercept_unary_unary(self, continuation, client_call_details, request):
+        new_details, new_request_iterator, postprocess = self._fn(
+            client_call_details, iter((request,)), False, False)
+        response = continuation(new_details, next(new_request_iterator))
+        return postprocess(response) if postprocess else response
```
software-dov: Style nit: it may make the code a little clearer to lift this evaluation into its own line, i.e.
```python
response = postprocess(response) if postprocess else response
return response
```
Judgement call on your part, feel free to ignore.

tests/system/conftest.py
```diff
@@ -34,3 +35,81 @@ def identity():
...
+
```
software-dov: I think mypy is going to complain unless there are two spaces between methods.

tests/system/conftest.py
```diff
@@ -34,3 +35,81 @@ def identity():
...
+    def intercept_stream_stream(self, continuation, client_call_details,
+                                request_iterator):
+        new_details, new_request_iterator, postprocess = self._fn(
+            client_call_details, request_iterator, True, True)
+        response_it = continuation(new_details, new_request_iterator)
+        return postprocess(response_it) if postprocess else response_it
+
+
+class _ClientCallDetails(
```
software-dov: If I understand what you're trying to do here, it may be easier to read if it's a [dataclass](https://docs.python.org/3.7/library/dataclasses.html) instead of inheriting from namedtuple.

tests/system/conftest.py
```diff
@@ -34,3 +35,81 @@ def identity():
...
+            '_ClientCallDetails',
+            ('method', 'timeout', 'metadata', 'credentials')),
+        grpc.ClientCallDetails):
+    pass
+
+
+# metadata interceptor to add a key value pair to client metadata
+def metadata_interceptor(key, value):
+
+    def intercept_call(client_call_details, request_iterator, request_streaming,
```
software-dov: This is really, really cool, but I'm having a little trouble understanding the flow of types. Can you add type annotations to the signature and return value of intercept_call?

tests/system/test_grpc_interceptor_streams.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def test_unary_stream(intercepted_echo):
+    content = 'The hail in Wales falls mainly on the snails.'
+    responses = intercepted_echo.expand({
+        'content': content,
+    })
+
+    # Consume the response and ensure it matches what we expect.
+    # with pytest.raises(exceptions.NotFound) as exc:
```
software-dov: Is this cruft from debugging?

tests/system/test_grpc_interceptor_streams.py
```diff
@@ -0,0 +1,50 @@
...
+
+def test_stream_stream(intercepted_echo):
+    requests = []
+    requests.append(showcase.EchoRequest(content="hello"))
+    requests.append(showcase.EchoRequest(content="world!"))
+    responses = intercepted_echo.chat(iter(requests))
+
+    contents = []
+    for response in responses:
+        contents.append(response.content)
```
software-dov: Nit: can just do
```python
contents = list(responses)
```

tests/system/test_grpc_interceptor_streams.py
```diff
@@ -0,0 +1,50 @@
+# Copyright 2018 Google LLC
```
busunkim96:
```suggestion
# Copyright 2020 Google LLC
```

software-dov: Nice catch :P

tests/system/test_grpc_interceptor_streams.py
arithmetic1728: here contents is a list of response.content not response

tests/system/test_grpc_interceptor_streams.py
arithmetic1728: removed

arithmetic1728:
Simplified client interceptor code, now it should be much more readable.

tests/system/test_grpc_interceptor_streams.py
software-dov: Okay

tests/system/conftest.py
```diff
@@ -34,3 +35,54 @@ def identity():
...
+    def __init__(self, key, value):
+        self._key = key
+        self._value = value
+
+    def _add_metadata(self, client_call_details):
+        if client_call_details.metadata is not None:
+            client_call_details.metadata.append((self._key, self._value,))
+
+    def intercept_unary_unary(self, continuation, client_call_details, request):
+        self._add_metadata(client_call_details)
```
software-dov: Unless I'm missing something, the code for all the interceptor calls is now identical. Maybe abstract that into a helper method and just alias that.

tests/system/test_grpc_interceptor_streams.py
software-dov: The python nitpicker in me still wants to turn this into a comprehension =P
```python
contents = [response.content for response in responses]
```

tests/system/conftest.py
arithmetic1728: there are slice differences in the ```continuation``` call arg and return value, request vs request_iterator, response/response_iterator. I think it might be more clear to write them separately.

PR #291, opened 2020-02-10
Change the behavior of the fixup script

software-dov:
Only script parameters are a single input dir and a single output dir
Fixups are no longer in-place: they are written to the corresponding
path in the output dir. The files in the input dir are unmodified.

Fixup script works on all services in an api concurrently instead of
one script per service. It has been renamed accordingly.

Hand testing checks out.

### DIFF
--- a/gapic/templates/scripts/fixup_keywords.py.j2
+++ b/gapic/templates/scripts/fixup_keywords.py.j2
@@ -3,7 +3,8 @@
 import argparse
 import os
 import libcst as cst
-from itertools import chain
+import pathlib
+import sys
 from typing import (Any, Callable, Dict, List, Sequence, Tuple)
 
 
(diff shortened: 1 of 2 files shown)


BenRKarl:
Just so I understand, since I haven't been close to the fixup scripts thus far. This script consumes a code base, i.e. a set of `.py` files, looks for instances where our `google-ads-python` library is used, and modifies the usage, where appropriate, to correspond to the microgenerator interface?

gapic/templates/scripts/fixup_keywords.py.j2
```diff
@@ -21,14 +22,18 @@ def partition(
...
 
 
-class {{ service.client_name }}CallTransformer(cst.CSTTransformer):
+class {{ api.naming.module_name }}CallTransformer(cst.CSTTransformer):
     CTRL_PARAMS: Tuple[str] = ('retry', 'timeout', 'metadata')
-
+    {% with all_methods = [] -%}
+    {% for service in api.services.values() %}{% for method in service.methods.values() -%}
+    {% do all_methods.append(method) -%}
+    {% endfor %}{% endfor -%}
```
BenRKarl: Is there a reason these lines aren't nested, for example:

```
{% with all_methods = [] -%}
    {% for service in api.services.values() %}
        {% for method in service.methods.values() -%}
            {% do all_methods.append(method) -%}
        {% endfor %}
    {% endfor -%}
    ...
{% endwith %}
```

software-dov: Whitespace control. All of the strangeness in the template is, ultimately, about whitespace control. To the best of my knowledge, [Jinja](https://palletsprojects.com/p/jinja/) is primarily intended to be a template engine for HTML/CSS, which are whitespace insensitive, and which in any case means that the viewer won't see any weird formatting stuff. We have to jump through a few hoops and engage in some tortuous indirection in order to make it generate reasonable, human-viewable Python, and even then the result is sometimes frustrating.

software-dov:
That's correct; it uses a rough match, based on method names, to change the method invocation to a microgenerator-friendly expression. The details are based on the microgenerator method interface and monolith/microgen flattening delta, the details of which are too much to describe in comments here.

PR #327, opened 2020-02-27
Add author and license to setup.py template

busunkim96:
To give nice metadata to pip and PyPI

### DIFF
--- a/gapic/templates/setup.py.j2
+++ b/gapic/templates/setup.py.j2
@@ -7,8 +7,8 @@ import setuptools  # type: ignore
 setuptools.setup(
     name='{{ api.naming.warehouse_package_name }}',
     version='0.0.1',
-    author="Google LLC",
-    license="Apache 2.0",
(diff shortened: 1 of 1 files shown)


gapic/templates/setup.py.j2
```diff
@@ -7,6 +7,8 @@ import setuptools  # type: ignore
 setuptools.setup(
     name='{{ api.naming.warehouse_package_name }}',
     version='0.0.1',
+    author='Google LLC',
+    license='Apache 2.0',
```
software-dov: Not hugely serious, but if we are committed to the generator being used beyond our APIs, should this not be a standalone template that is included and can be overridden more easily? That's what we do for the license header in the generated source files.

busunkim96: Ah good points. I didn't realize that's how the license header was getting pulled in. Will close this PR and poke around a bit more.

Issue #318, opened 2020-02-26
LRO metadata and response types that aren't imported cause generator crashes


software-dov:
Example:
```proto
// bivalve.proto
// MolluscMetadata lives in mollusc.proto which is not imported
rpc IdentifyBivalve(IdentifyBivalveRequest) returns (google.longrunning.Operation) {
  option (google.longrunning.operation_info) = {
    response_type: "Mollusc",
    metadata_type: "google.molluscs.v1.MolluscMetadata",
  };
} 
```

Causes the following backtrace:
```
 File "/usr/local/google/home/dovs/code/gapic-generator-python/gapic/schema/api.py", line 546, in _get_methods
    metadata_type=self.api_messages[service_address.resolve(
  File "/usr/local/google/home/dovs/.pyenv/versions/3.8.1/lib/python3.8/collections/__init__.py", line 891, in __getitem__
    return self.__missing__(key)            # support subclasses that define __missing__
  File "/usr/local/google/home/dovs/.pyenv/versions/3.8.1/lib/python3.8/collections/__init__.py", line 883, in __missing__
    raise KeyError(key)
KeyError: 'google.molluscs.v1.MolluscsMetadata'
--python_gapic_out: protoc-gen-python_gapic: Plugin failed with status code 1.
```

There are a few possible solutions:
1. Require the corresponding import statement in the proto file.
2. Scour the entire API for LRO response/metadata types instead of just the current view.

The downside of 1 is that it will lead to more "imported but not used warnings". It is tricky to implement 2 because of the way that the generated surface references types when wrapping LROs.

software-dov:
A common case is fixed by #330

software-dov:
This is technically prohibited by [AIP-151](https://aip.dev/151). Closing as kinda fixed, kinda not.

PR #330, opened 2020-02-29
Unimported LRO type annotation fix

software-dov:
Partial fix for #318 
If a LRO returning method has a type annotation for a type that is not imported but is in the API, that is now supported. The general case where the annotated type lives in an arbitrary proto is not addressed.
Tested against the `functions` API, which now builds and passes generated unit tests.

### DIFF
--- a/gapic/schema/api.py
+++ b/gapic/schema/api.py
@@ -56,11 +56,13 @@ class Proto:
         return getattr(self.file_pb2, name)
 
     @classmethod
-    def build(cls, file_descriptor: descriptor_pb2.FileDescriptorProto,
-              file_to_generate: bool, naming: api_naming.Naming,
-              opts: options.Options = options.Options(),
-              prior_protos: Mapping[str, 'Proto'] = None,
-              ) -> 'Proto':
+    def build(
+        cls, file_descriptor: descriptor_pb2.FileDescriptorProto,
+        file_to_generate: bool, naming: api_naming.Naming,
+        opts: options.Options = options.Options(),
+        prior_protos: Mapping[str, 'Proto'] = None,
+        load_services: bool = True
+    ) -> 'Proto':
         """Build and return a Proto instance.
 
         Args:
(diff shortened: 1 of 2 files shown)


gapic/schema/api.py
```diff
@@ -221,16 +236,43 @@ def disambiguate_keyword_fname(
...
+        # don't import the proto file defining the real response or metadata
+        # type into the proto file that defines an LRO.
+        # We just load all the APIs types first and then
+        # load the services and methods with the full scope of types.
+        pre_protos: Dict[str, Proto] = dict(prior_protos or {})
         for fd in file_descriptors:
-            fd.name = disambiguate_keyword_fname(fd.name, protos)
-            protos[fd.name] = _ProtoBuilder(
+            fd.name = disambiguate_keyword_fname(fd.name, pre_protos)
+            pre_protos[fd.name] = Proto.build(
```
hzyi-google: Is it possible to refactor `Proto.build` so that it loads all messages and enums first, and then build methods and services? So instead of calling `build` twice, you could have build call part of itself to "prebuild". It seems cleaner. I don't know the implementation details so this might be impossible or very hard in which case ignore me :)

software-dov: There are a couple of tricky issues to work around:
* `Proto` is a frozen datatype, so once it's been created it is immutable.
* We need some kind of two stage solution so that we load _all_ the proto messages, and then methods can see all message and enum types.

I would certainly be willing to rework the current implementation so that messages and enums are loaded and wrapped in rich structures, then services and methods are loaded and wrapped in rich structures using the messages and enums, then `Proto` is created by injecting messages, enums, services, and methods. This would require a larger change, but would be cleaner and more aesthetically pleasing. I could go either way.

software-dov:
Deferring merge for other reviewers and to permit myself some time for experimental refactoring.

gapic/schema/api.py
hzyi-google: Sounds like a lot of work :( You don't have to block on this comment. Feel free to merge if the fix is pressing and the Python developers sign off.

software-dov: I definitely want to rework the internal implementation, but that is going to take some time. I'll file a cleanup issue around it but commit the current impl as is, pending approval.

software-dov: Tracked by #331

PR #341, opened 2020-03-12
Turn naming.Naming into an abstract class

software-dov:
The new style client libraries have a directory/module structure of
the form:

apiName:
\- \_\_init\_\_.py <alias to the most recent version>
apiName_version:
\- \_\_init\_\_.py <load all the types and submodules>

Certain client libraries need to preserve a legacy interface with a
directory structure of the form

apiName:
\- version

This change abstracts out the Naming class to make make the change feasible.

### DIFF
--- a/gapic/schema/naming.py
+++ b/gapic/schema/naming.py
@@ -32,7 +32,7 @@ class Naming(abc.ABC):
     This class contains the naming nomenclature used for this API
     within templates.
 
-    An concree child of this object is made available to every template
+    An concrete child of this object is made available to every template
     (as ``api.naming``).
     """
     name: str = ''
--- a/gapic/schema/naming.py
+++ b/gapic/schema/naming.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import abc
 import dataclasses
 import os
 import re
@@ -22,15 +23,16 @@ from google.protobuf import descriptor_pb2
 from gapic import utils
 from gapic.generator import options
 
-
-@dataclasses.dataclass(frozen=True)
-class Naming:
+# See https://github.com/python/mypy/issues/5374 for details on the mypy false
+# positive.
+@dataclasses.dataclass(frozen=True)  # type: ignore
+class Naming(abc.ABC):
     """Naming data for an API.
 
     This class contains the naming nomenclature used for this API
     within templates.
 
-    An instance of this object is made available to every template
+    An concree child of this object is made available to every template
     (as ``api.naming``).
     """
     name: str = ''
@@ -43,11 +45,11 @@ class Naming:
         if not self.product_name:
             self.__dict__['product_name'] = self.name
 
-    @classmethod
-    def build(cls,
-              *file_descriptors: descriptor_pb2.FileDescriptorProto,
-              opts: options.Options = options.Options(),
-              ) -> 'Naming':
+    @staticmethod
+    def build(
+        *file_descriptors: descriptor_pb2.FileDescriptorProto,
+        opts: options.Options = options.Options(),
+    ) -> 'Naming':
         """Return a full Naming instance based on these file descriptors.
 
         This is pieced together from the proto package names as well as the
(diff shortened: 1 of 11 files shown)


gapic/schema/naming.py
```diff
@@ -22,15 +23,16 @@
...
+# positive.
+@dataclasses.dataclass(frozen=True)  # type: ignore
+class Naming(abc.ABC):
     """Naming data for an API.
 
     This class contains the naming nomenclature used for this API
     within templates.
 
-    An instance of this object is made available to every template
+    An concree child of this object is made available to every template
```
lukesneeringer: concrete

lukesneeringer:
> Certain client libraries need to preserve a legacy interface with a
> directory structure of the form

Which ones?

gapic/schema/naming.py
software-dov: Nice catch

software-dov:
Ads needs the legacy structure.

lukesneeringer:
> Ads needs the legacy structure.

Wait, what? They do? That makes no sense; they have generated with the monolith, which does not output that structure.

This also fails the five APIs rule. :-( Can this be done either with post-processing or custom templates?

software-dov:
I believe they sent the monolith output through postprocessing to achieve structure like this.
This change is a requirement for imports within the custom templates to make sense. Addresses and names are not template agnostic.

The alternative is to use custom templates and then fix up the imports via post-processing.

ccing @BenRKarl for his perspective and to clarify any erroneous statements I've made.

BenRKarl:
@software-dov @lukesneeringer Confirming that we do a good bit of post-processing on the monolithic output. I'll need to check with my team on what does/does not get handled by the post processor though. In any case I believe the goal for us is to minimize post-processing as much as possible, if not entirely.

bobhancock:
The monolithic generator outputs this directory structure:
└── python
    └── googleads-v3
        ├── docs
        │   └── gapic
        │       └── v3
        ├── google
        │   └── ads
        │       └── googleads_v3
        │           ├── gapic
        │           │   └── services
        │           │       └── transports
        │           └── proto
        │               ├── common
        │               ├── enums
        │               ├── errors
        │               ├── resources
        │               └── services
        └── tests
            └── unit
                └── gapic
                    └── v3

Any post processing we do is to change strings inside the files. No post processing of the directory structure.

software-dov:
Then I'm afraid I've been misunderstanding the required directory structure.

software-dov:
Wait a second; the directory structure in [google-ads-python](https://github.com/googleads/google-ads-python/tree/master/google/ads/google_ads) does imply a different directory structure. Is this not the ads client library?

BenRKarl:
@software-dov you're looking at the correct library. I think Bob is correct in that we don't modify the dir structure of the monolithic output, but we also don't use the entire output as-is. We just copy the contents of a portion to the client library, like so:

```
$ cp /monolithic_output/google/ads/google_ads/<version> /google-ads-python/google/ads/google_ads/
```

bobhancock:
We do not modify the directory structure. The directory tree above is what is produced by the monolithic generator.

software-dov:
I've sent @BenRKarl the most recent generated GAPIC with the custom template and the 'old-style' naming for perusal. If it is indeed the correct surface, i.e. one which the handwritten layer expects, then we can resume the conversation about how to deal with it.

My ranked option choices are:
1) Do nothing, force the Ads to modify the handwritten layer to accommodate the new structure, kick the can.
    Previous meetings have come to the conclusion that this is unworkable.
2) Implement a fix purely through custom templates.
    I don't think this is possible; the `Naming` class is tied too tightly to the module structure of the default templates, which means that import paths throughout the surface and generated unit tests would be incorrect.
3) Make a one-shot custom hack to `schema.wrappers.Naming` toggled via a special flag/option.
    This is the 'default' choice, as it describes the code in this PR.
4) Add a plugins system so that custom `Naming` and `Import` subclasses can be defined without adding one-shot hacks to core datatypes and logic.
    I think that long term this may end up being more correct, especially if we find other use cases for custom templates that deviate from the default directory structure. At the moment, though, it seems like a much more invasive and difficult change than we need, and it will delay the implementation. I'm also not sure how to integrate a plugins system with Bazel.
5) Fix up import paths via post-processing.
    This is the bad-old status quo.

What do other people think?

software-dov:
Just finished skimming through the postprocessing script courtesy of @bobhancock. The script _does_ move directories around in a manner consistent with the naming convention option in this PR and the custom template.
To that end, open review/conversation bump.

Issue #356, opened 2020-04-02
Consolidate Ads Templates in Main Repo


BenRKarl:
Right now custom templates for Ads live in a separate fork owned by @software-dov. After some discussion we decided that those templates should reside in the main repo on `master`

software-dov:
Implemented by #357

PR #357, opened 2020-04-02
Dual templates

software-dov:
Add an alternative set of template as a first-class supported target with corresponding tests and CI targets.


software-dov:
Implementation for #356

PR #359, opened 2020-04-03
add mTLS support to generator

arithmetic1728:
For #360 

Currently python-api-core doesn't have `client_cert_source` in `ClientOptions`, so the code in this PR uses the unreleased python-api-core code in this branch:
https://github.com/googleapis/python-api-core/tree/sijun/cert

PR #365, opened 2020-04-09
feat: Provide AsyncIO support for generated code

lidizheng:
Let's see how the CI react to the change.
The test might not pass until changes in https://github.com/googleapis/python-api-core/pull/22 got released.

### DIFF
--- a/gapic/schema/wrappers.py
+++ b/gapic/schema/wrappers.py
@@ -566,6 +566,13 @@ class Method:
 
     @utils.cached_property
     def client_output(self):
+        return self._client_output(enable_asyncio=False)
+
+    @utils.cached_property
+    def client_output_async(self):
+        return self._client_output(enable_asyncio=True)
+
+    def _client_output(self, enable_asyncio: bool):
         """Return the output from the client layer.
 
         This takes into account transformations made by the outer GAPIC
@@ -584,8 +591,8 @@ class Method:
         if self.lro:
             return PythonType(meta=metadata.Metadata(
                 address=metadata.Address(
-                    name='Operation',
-                    module='operation',
+                    name='AsyncOperation' if enable_asyncio else 'Operation',
+                    module='operation_async' if enable_asyncio else 'operation',
                     package=('google', 'api_core'),
                     collisions=self.lro.response_type.ident.collisions,
                 ),
@@ -603,7 +610,7 @@ class Method:
         if self.paged_result_field:
             return PythonType(meta=metadata.Metadata(
                 address=metadata.Address(
-                    name=f'{self.name}Pager',
+                    name=f'{self.name}AsyncPager' if enable_asyncio else f'{self.name}Pager',
                     package=self.ident.api_naming.module_namespace + (self.ident.api_naming.versioned_module_name,) + self.ident.subpackage + (
                         'services',
                         utils.to_snake_case(self.ident.parent[-1]),
@@ -744,6 +751,8 @@ class Method:
         if not self.void:
             answer.append(self.client_output)
             answer.extend(self.client_output.field_types)
+            answer.append(self.client_output_async)
+            answer.extend(self.client_output_async.field_types)
 
         # If this method has LRO, it is possible (albeit unlikely) that
         # the LRO messages reside in a different module.
@@ -802,12 +811,21 @@ class Service:
         return self.name + "Client"
 
     @property
+    def async_client_name(self) -> str:
+        """Returns the name of the generated AsyncIO client class"""
+        return self.name + "AsyncClient"
+
+    @property
     def transport_name(self):
         return self.name + "Transport"
 
     @property
     def grpc_transport_name(self):
         return self.name + "GrpcTransport"
+
+    @property
+    def grpc_asyncio_transport_name(self):
+        return self.name + "GrpcAsyncIOTransport"
 
     @property
     def has_lro(self) -> bool:
@@ -856,7 +874,7 @@ class Service:
         used for imports.
         """
         # Put together a set of the service and method names.
-        answer = {self.name, self.client_name}
+        answer = {self.name, self.client_name, self.async_client_name}
         answer.update(
             utils.to_snake_case(i.name) for i in self.methods.values()
         )
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
@@ -2,9 +2,9 @@
 
 {% block content %}
 from collections import OrderedDict
-import os
+import functools
 import re
-from typing import Callable, Dict, {% if service.any_server_streaming %}Iterable, {% endif %}{% if service.any_client_streaming %}Iterator, {% endif %}Sequence, Tuple, Type, Union
+from typing import Dict, {% if service.any_server_streaming %}AsyncIterable, {% endif %}{% if service.any_client_streaming %}AsyncIterator, {% endif %}Sequence, Tuple, Type, Union
 import pkg_resources
 
 import google.api_core.client_options as ClientOptions # type: ignore
@@ -12,8 +12,6 @@ from google.api_core import exceptions                 # type: ignore
 from google.api_core import gapic_v1                   # type: ignore
 from google.api_core import retry as retries           # type: ignore
 from google.auth import credentials                    # type: ignore
-from google.auth.transport import mtls                 # type: ignore
-from google.auth.exceptions import MutualTLSChannelError  # type: ignore
 from google.oauth2 import service_account              # type: ignore
 
 {% filter sort_lines -%}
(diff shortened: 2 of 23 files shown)


gapic/schema/wrappers.py
```diff
@@ -593,6 +593,63 @@ def client_output(self):
         # Return the usual output.
         return self.output
 
+    @utils.cached_property
+    def client_output_async(self):
+        """Return the output from the client layer.
```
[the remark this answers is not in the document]
lidizheng: Good idea! I will create a function to hide common logic. In next pass, I will try to improve readability using similar pattern.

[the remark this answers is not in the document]
software-dov: Thanks. If it turns out to be tricky then no need to twist the logic too much.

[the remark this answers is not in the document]
lidizheng: I didn't find a way to unify this specific method without regression.
1. The `client_output` is cached, so it can't accept parameters;
2. The output of `client_output` is frozen, so we can't modify the content in it (we can create a new one but it is worse than what we have now).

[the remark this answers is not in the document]
software-dov: Keep `client_output` and `client_output_async` as the cached layer, then move the guts of the logic into an internal method that takes a parameter. The internal method should _not_ be cached and should not be a property.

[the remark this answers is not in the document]
lidizheng: Created the common internal method in 4ecf34dd1e0303210521d7f315a03eb2d5324015.

lidizheng:
This PR may need a smarter way to reuse existing logic. The newly generated class should not be the burden for incoming important features like https://github.com/googleapis/gapic-generator-python/pull/359.

lidizheng:
To improve the maintainability of the microgenerator project, we might want to reduce the duplicated logic. Otherwise, new features will have to apply same changes in multiple places, which is quite error-prone. I want to list the options we have to compare:

1. Composition: all AsyncIO class contains an instance of existing class;
2. Helper functions: promote duplicated logic between the existing class and AsyncIO class to external functions;
3. Inheritance: create a common base class for both existing class and AsyncIO class (https://github.com/googleapis/gapic-generator-python/pull/365/commits/13fba42e16631b08e9187b4e605e325ce8b4d5ca);
4. Duplication: duplicate the logic (https://github.com/googleapis/gapic-generator-python/pull/365/commits/a51bcd3810b424ce2cacdacad054a15b895c0eef).

Options 1, 2, 3 solve the maintainability issue, but have different drawbacks. I can change to either one, if you see fit. WDYT? @software-dov @busunkim96 

For option 1, proxying methods from one class to another requires quite some boilerplates. For example, the existing Client class has not only instance method, but also class variables and class methods. The code looks like `class_method_a = _comp_class.method_a`, `method_b = _comp_instance.method_b`. Metaclass pattern makes this option even worse which requires its own composition logic. So, the implementation might look hacky.

For option 2, the issue of helper functions is that the duplicated logic usually contains modifying member variables. Passing `self` out, and allow external function to update it is less ideal. Also, helper functions doesn't solve the member variable initialization issue. The member variable still needs to be defined explicitly.

For option 3, inheritance increases code complexity, and challenges the scope of the interface abstraction. For example, it is unclear that should MTLS feature apply to all transport or only gRPC transport. It also makes it harder for users to understand the logic, since they need to read multiple class instead of one.

software-dov:
My preferences would be 2, then 1, then 3, then 4. @arithmetic1728 has confirmed that mTLS logic should eventually apply to http transports, and it presumably will also apply to asynchronous clients. I could be persuaded (and please do, if experiments indicate it!) that composition is a less gross hack than helper functions.
With helpers we need to be careful not to wind up duplicating the downsides of inheritance: if a function takes a Client OR an AsyncClient, with explicit type checking and separate logic paths, then that's arguably worse. If we tack on too many accessor methods to facilitate the helper functions, we've just split the implementation between 2 and 3.

I think, in light of the above, I'm leaning more towards composition as a general strategy.

lidizheng:
@software-dov I have experiment with the composition pattern to unify logic between **Client and AsyncClient**, PTAL at commit https://github.com/googleapis/gapic-generator-python/pull/365/commits/3cc730c80165953850ab1848aaf38bee470d2b46.

The composition pattern doesn't work cleanly with **Transport classes** due to function signature difference. AsyncTransport stubs returns `Awaitable[Response]`(coroutine) while gRPC Transport returns `Response` type directly. Also, the channel object has different type (`grpc.Channel` vs. `aio.Channel`). The options are:

1. Have a common parent class to make types more generic (see https://github.com/googleapis/gapic-generator-python/pull/365/commits/3cc730c80165953850ab1848aaf38bee470d2b46);
2. Similar to option 1, but instead of using inheritance we can use composition to weave the new class and the two public transport classes;
3. Loosen the type restriction on Transport classes, so it enables GrpcTransport to composite AsyncTransport class;
4. Let the logic being copied twice (see https://github.com/googleapis/gapic-generator-python/pull/365/commits/e17602057ae87815593f3aa4fb810dac0577b641).

@software-dov WDYT?

plamut:
@lidizheng If I understand correctly, we want type compatibility between the transport classes `grpc.Channel` and `aio.Channel`? Would it help if we create an abstract Channel base class and [register](https://docs.python.org/3/library/abc.html#abc.ABCMeta.register) the two existing Channel classes as its "virtual subclasses"?

lidizheng:
For channels, if we loosen the type check, it is equivalent to option 3. Abstract Channel base class works. Another option is type it with `Union[grpc.Channel, aio.Channel]`. It stops linter from complaining, but it might be error-prone if user pass-in an AsyncIO channel into a normal gRPC Transport.

software-dov:
Possibly bad idea: is there any way we can rewrite the synchronous Transport in terms of the async and just force explicit `await`s?

I'm a little lost reading through the proposed diffs and the existing `aio` and regular grpc code. If I understand correctly, the _main_ problem is the return type/semantics of the transport stubs.
If we can come up with a reasonable solution for that, it looks like everything else is a workable problem.

lidizheng:
> I'm a little lost reading through the proposed diffs and the existing aio and regular grpc code. If I understand correctly, the main problem is the return type/semantics of the transport stubs.
If we can come up with a reasonable solution for that, it looks like everything else is a workable problem.

Yes, the composition pattern works except the types. The complain can be suppressed by `typing.cast` if you don't think it is hacky.

> Possibly bad idea: is there any way we can rewrite the synchronous Transport in terms of the async and just force explicit awaits?

When I'm writing gRPC AsyncIO, I found even though AsyncIO is the official asynchronous library. People still appreciate the ability to use normal synchronous Python.

arithmetic1728:
> @arithmetic1728 Thanks for the quick response. The test I mentioned is [this](https://github.com/googleapis/gapic-generator-python/blob/a48ff4dc3706891b39338016deb052645238d3e2/gapic/templates/tests/unit/%25name_%25version/%25sub/test_%25service.py.j2#L105).
> 
> ```
>     os.environ["GOOGLE_API_USE_MTLS"] = "never"
>     with mock.patch('{{ (api.naming.module_namespace + (api.naming.versioned_module_name,) + service.meta.address.subpackage)|join(".") }}.services.{{ service.name|snake_case }}.transports.{{ service.name }}GrpcTransport.__init__') as grpc_transport:
>         grpc_transport.return_value = None
>         client = {{ service.client_name }}()
>         grpc_transport.assert_called_once_with(
>             api_mtls_endpoint=client.DEFAULT_ENDPOINT,
>             client_cert_source=None,
>             credentials=None,
>             host=client.DEFAULT_ENDPOINT,
>         )
> ```
> 
> This is the test case that expects `api_mtls_endpoint` under "never".

Yes, please see my previous comments. 

Let's schedule a meeting to discuss this, it might be easier that way.

lidizheng:
@arithmetic1728  Thanks! See you offline (online).

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -170,7 +172,7 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
         if client_options is None:
             client_options = ClientOptions.ClientOptions()
 
-        if transport is None and client_options.api_endpoint is None:
```
arithmetic1728, line 173: Since the condition changed (now it becomes: `client_options` is used when transport is not a {{ service.name }}Transport object), please update the `client_options` document above to reflect this change.

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
```diff
@@ -0,0 +1,266 @@
...
+        Args:
+            credentials (Optional[google.auth.credentials.Credentials]): The
+                authorization credentials to attach to requests. These
+                credentials identify the application to the service; if none
+                are specified, the client will attempt to ascertain the
+                credentials from the environment.
+            transport (Union[str, ~.{{ service.name }}Transport]): The
+                transport to use. If set to None, a transport is chosen
+                automatically.
+            client_options (ClientOptions): Custom options for the client.
```
arithmetic1728, line 63: please update the document string for `client_options`, it seems out of sync.

arithmetic1728:
After discussing with Lidi, I think the mTLS change looks good to me. LGTM for the mTLS change.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
lidizheng: The docstring updated to:
```
            client_options (ClientOptions): Custom options for the client. It
                won't take effect if a ``transport`` instance is provided.
```

lidizheng:
@software-dov @busunkim96 PTALAA.

tests/system/conftest.py
```diff
@@ -81,6 +96,55 @@ def echo(use_mtls):
...
+    )
+    return IdentityClient(transport=transport)
+
+
+@pytest.fixture
+def async_identity(use_mtls):
+    if use_mtls:
+        with mock.patch("grpc.ssl_channel_credentials", autospec=True) as mock_ssl_cred:
+            mock_ssl_cred.return_value = ssl_credentials
+            client = IdentityAsyncClient(
```
software-dov, line 133: The mainline code has generalized these somewhat, heads up.

lidizheng:
@software-dov Thanks for the review! I updated `conftest.py` to use the generalized factory function. About alternative templates, I wonder if Jinja2's [include](https://jinja.palletsprojects.com/en/2.11.x/templates/#include) command helps? It can import text files (including another template file).

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc_asyncio.py.j2
```diff
@@ -0,0 +1,207 @@
...
+                       **kwargs) -> aio.Channel:
+        """Create and return a gRPC AsyncIO channel object.
+        Args:
+            address (Optionsl[str]): The host for the channel to use.
+            credentials (Optional[~.Credentials]): The
+                authorization credentials to attach to requests. These
+                credentials identify this application to the service. If
+                none are specified, the client will attempt to ascertain
+                the credentials from the environment.
+            scopes (Sequence[str]): A optional list of scopes needed for this
```
busunkim96, line 56:
```suggestion
            scopes (Optional[Sequence[str]]): A optional list of scopes needed for this
```

tests/system/test_grpc_unary.py
```diff
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 import pytest
+import asyncio
+import logging
```
busunkim96, line 17: Is the `logging` import being used?

tests/system/test_grpc_streams.py
```diff
@@ -71,3 +75,145 @@ def test_stream_stream_passing_dict(echo):
...
+
+@pytest.mark.asyncio
+async def test_async_unary_stream_reader(async_echo):
+    logging.debug('test_async_unary_stream_reader %s %s',
+                  threading.current_thread(), asyncio.get_event_loop())
+    content = 'The hail in Wales falls mainly on the snails.'
+    call = await async_echo.expand({
+        'content': content,
+    }, metadata=metadata)
+    logging.debug('test_async_unary_stream_reader 2')
```
busunkim96, line 88: Are these additional numbered debug statements left in this function intentionally?

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
```diff
@@ -116,15 +118,19 @@ class {{ service.name }}GrpcTransport({{ service.name }}Transport):
                 credentials identify this application to the service. If
                 none are specified, the client will attempt to ascertain
                 the credentials from the environment.
+            scopes (Sequence[str]): A optional list of scopes needed for this
```
busunkim96, line 121:
```suggestion
            scopes (Optional[Sequence[str]]): A optional list of scopes needed for this
```

lidizheng:
@busunkim96 Thanks for reviewing. I'm shamed that I forgot to remove the debug logging. All comments addressed. Do I need to squash all commits?

busunkim96:
@software-dov What merge style is preferred in this repo?

Issue #360, opened 2020-04-07
add mTLS support to generator


arithmetic1728:

arithmetic1728:
We need the `ClientOptions` PR in python-api-core to add `client_cert_source` property: https://github.com/googleapis/python-api-core/pull/17

PR #374, opened 2020-04-09
fix: fix mTLS logic

arithmetic1728:
new logic for client_options:
1. api_endpoint: yes, client_cert_source: yes => use as it is (same as before)
2. api_endpoint: yes, client_cert_source: no => no adc (previously try to load adc SSL cred)
3. api_endpoint: whatever, client_cert_source: yes => use cert from client_cert_source, use provided api_endpoint or default mTLS endpoint (same as before)
4. api_endpoint: no, client_cert_source: no => do nothing (same as before)

#360

### DIFF
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
@@ -91,7 +91,6 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
     DEFAULT_MTLS_ENDPOINT = _get_default_mtls_endpoint.__func__(  # type: ignore
         DEFAULT_ENDPOINT
     )
-    DEFAULT_OPTIONS = ClientOptions.ClientOptions(api_endpoint=DEFAULT_ENDPOINT)
 
     @classmethod
     def from_service_account_file(cls, filename: str, *args, **kwargs):
@@ -126,7 +125,7 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
     def __init__(self, *,
             credentials: credentials.Credentials = None,
             transport: Union[str, {{ service.name }}Transport] = None,
-            client_options: ClientOptions = DEFAULT_OPTIONS,
+            client_options: ClientOptions = None,
             ) -> None:
         """Instantiate the {{ (service.client_name|snake_case).replace('_', ' ') }}.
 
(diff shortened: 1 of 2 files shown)


gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -170,24 +164,27 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
-            # endpoint, otherwise use the default mTLS endpoint.
-            option_endpoint = client_options.api_endpoint
-            api_mtls_endpoint = self.DEFAULT_MTLS_ENDPOINT if option_endpoint == self.DEFAULT_ENDPOINT else option_endpoint
+            # We have a non-empty ClientOptions, trigger mTLS. If the user
+            # doesn't provide endpoint, use the default mTLS endpoint.
+            if client_options.api_endpoint:
+                api_endpoint = api_mtls_endpoint = client_options.api_endpoint
+            else:
+                api_endpoint = self.DEFAULT_ENDPOINT
+                api_mtls_endpoint = self.DEFAULT_MTLS_ENDPOINT
```
software-dov, line 183: Judgement call on your part, but what do you think about
```python
opt_endpoint = client_options.api_endpoint
endpoint, mtls_endpoint = (opt_endpint, opt_endpoint) if opt_endpoint else (self.DEFAULT_ENDPOINT, self.DEFAULT_MTLS_ENDPOINT)
```

I generally prefer conditional assignment in python, when it isn't too complicated, because it eliminates undefined reference errors; I also find it's often easier to reason about how values flow through code than logic branches.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -170,24 +164,27 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
 
             self._transport = {{ service.name }}GrpcTransport(
```
software-dov, line 185: This may be addressed in other documents somewhere, but does mTLS require gRPC, or could it be used over http?

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
arithmetic1728: logic is changed in the new code. Sure, I will use conditional assignment when simple and possible.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
arithmetic1728: it is for http and grpc. Will need to add http mTLS support when microgenerator supports http.

PR #394, opened 2020-04-15
fix: add system tests and cert/key

arithmetic1728:
1. added testing cert/key for mtls system test (expired after 1000 years, CN=localhost). cert can also be used as root ca cert since it is self signed.
2. we will reuse all the system tests we have now. Introduced a `--mtls` flag to change all the channels to mtls channels, and added separated nox sessions.
3. since the generated client doesn't take root ca cert (only takes cert and key), we need to mock `grpc.grpc.ssl_channel_credentials`, assert it is called with the correct cert/key, but return the channel created with root ca cert, cert and key.


tests/system/conftest.py
```diff
@@ -13,76 +13,140 @@
...
     # The interceptor adds 'showcase-trailer' client metadata. Showcase server
     # echos any metadata with key 'showcase-trailer', so the same metadata
     # should appear as trailing metadata in the response.
-    interceptor = MetadataClientInterceptor('showcase-trailer', 'intercepted')
-    channel = grpc.insecure_channel('localhost:7469')
+    interceptor = MetadataClientInterceptor("showcase-trailer", "intercepted")
+    if use_mtls:
+        channel = grpc.secure_channel("localhost:7469", ssl_credentials)
+    else:
+        channel = grpc.insecure_channel("localhost:7469")
```
software-dov, line 147: Personal preference nit: what about
```python
host = "localhost:7469"
channel = grpc.secure_channel(host, ssl_credentials) if use_mtls else grpc.insecure_channel(host)
```
?

arithmetic1728: done

Issue #400, opened 2020-04-22
Fetching @gapic_generator_python_pip_deps takes 5+ minutes


aohren:
Breaking this out as a separate issue from #361 (where I first mentioned this). 

When trying to build the `py_gapic_library` Bazel rule, the `Fetching @gapic_generator_python_pip_deps` step takes a little >5 mins from a clean build every time. 

Reproducible by doing  `bazel clean --expunge` before building.

hkdevandla:
I guess @vam-google is looking into this. Vadym, please update with your findings.

Issue #390, opened 2020-04-14
No way to specify `python_gapic_opt` flag with `py_gapic_library` rule


aohren:
We need to replicate the following protoc invocation from Bazel rule `py_gapic_library`. In particular, there's no current way to set `--python_gapic_opt` to the desired value below:

```
  protoc google/ads/googleads/${VERSION}/**/*.proto \
    --proto_path="${COMMON_PROTOS_PATH}" \
    --proto_path=. \
    --python_gapic_out=${OUTPUT_DIR} \
    --plugin=${GAPIC_GENERATOR_PATH}/gapic/cli/generate.py=python_gapic \
    --python_gapic_opt="old-naming,lazy-import,python-gapic-name=${GAPIC_NAME},python-gapic-templates=${GAPIC_GENERATOR_PATH}/gapic/ads-template"
```

where one var is custom:

`GAPIC_NAME` is `googleads`.

And the other vars are what you'd expect:

`COMMON_PROTOS_PATH` is api-common-protos dependencies
`GAPIC_GENERATOR_PATH` is root of gapic-generator-python

aohren:
Output from a `bazel build --verbose_failures` run:

https://drive.google.com/a/google.com/file/d/1oKBkNbG6_wjwlNvS8F34CHQFBmLIpM8N/view?usp=sharing

BenRKarl:
@aohren there's an updated value for `python-gapic-templates` now, it should point to the `/gapic/ads-templates` directory on the master branch of this repository.

aohren:
@BenRKarl thanks for that. I've edited the original issue in-place to reflect the updated protoc arguments.

vam-google:
Should be fixed in #411. Closing.

software-dov:
Should be fixed by #402

PR #402, opened 2020-04-25
feat: Make GAPIC Bazel rules production ready

vam-google:
Shoud fix https://github.com/googleapis/gapic-generator-python/issues/400 and https://github.com/googleapis/gapic-generator-python/issues/390, plus a bunch of other not-yet-opened issues.

This includes:
1) Fix long initial load time (5+ min). This was caused by python_rules buildling `grpcio` dependency from sources in one core (which was super slow). Switched to using bazel-native `"@com_github_grpc_grpc//src/python/grpcio/grpc:grpcio"` target instead, which is not only much faster, but is also already used in googleapis, so there is no additional cost for reusing it in microgenerator rules.

2) Properly handle `pandoc` dependency (platform-sepcific version of pandoc is properly pulled by bazel itself using toolchains).

3) Add simplistic version of the `py_gapic_assembly_pkg` rule, to make output of microgenerator compatible with `GAPICBazel` class in synthtool.

4) Add `plugin_args` argument for python_gapic_library rule to pass custom argumetns to the plugin (similar to PHP rules).

5) Add compatibility with  `python3.6` runtime (otherwise `python3.7` is minimum because of dependency on `dataclasses` module). Python 3.6 compatibility can be enabled by adding `--define=gapic_gen_python=3.6` command line argument to `bazel build` command. 

6) Add support for Python runtimes installed with `pyenv`. To tell bazel using Python3 installed via pyenv add `--extra_toolchains=@gapic_generator_python//:pyenv3_toolchain` argument to `bazel build` command.


vam-google:
@software-dov @aohren PTAL

rules_python_gapic/py_gapic_pkg.bzl
```diff
@@ -0,0 +1,70 @@
...
+# limitations under the License.
+
+load("@com_google_api_codegen//rules_gapic:gapic_pkg.bzl", "construct_package_dir_paths")
+
+def _py_gapic_src_pkg_impl(ctx):
+    srcjar_srcs = []
+    for dep in ctx.attr.deps:
+        for f in dep.files.to_list():
+            if f.extension in ("srcjar", "jar", "zip"):
+                srcjar_srcs.append(f)
```
software-dov, line 22: Does skylark allow list comprehensions? Personal preference, feel free to ignore, but I think
```python
srcjar_srcs = [
    dep_file
    for dep in ctx.attr.deps
    for dep_file in dep.files.to_list()
    if dep_file.extension in ("srcjar", "jar", "zip")
]
```
is more readable because of the indentation.

WORKSPACE
```diff
@@ -22,14 +24,43 @@ pip_repositories()
...
 
 pip_install()
 
 load("@com_google_protobuf//:protobuf_deps.bzl", "protobuf_deps")
 
 protobuf_deps()
+
+#
+# Import grpc as a native bazel dependency. This avoids duplication and also
+# spedds up loading phase a lot (otherwise python_rules will be building grpcio
```
software-dov, line 46: Nit: speeds

software-dov:
Any updates or increased bandwidth?

vam-google:
@software-dov Still trying to make CI work with gapic-generator-python (python3.7 requirement issue). Before that issue is solved this PR can't be used (can't integrate it in googleapis).

WORKSPACE
```diff
@@ -22,14 +24,43 @@ pip_repositories()
...
+# from sources in a single-core speed, which takes around 5 minutes on a regular
+# workstation)
```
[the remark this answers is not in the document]
vam-google: I dont know how exactly it does it, but based on time it takes and the CPU load on the build machine I assumed it must be single core

rules_python_gapic/py_gapic_pkg.bzl
vam-google: It is a reduced verison of more complicated for loops copied from the other rules. Starlak support list comprehension. For sake of style conssstency with the other rules lets keep it like that.

vam-google:
@software-dov PTAL. I finally made it wokr from kokoro  under pyenv 3.6.1 python (the available on a kokoro instance)

BUILD.bazel
```diff
@@ -42,6 +66,8 @@ py_binary(
...
         requirement("pypandoc"),
         requirement("PyYAML"),
-    ],
-    python_version = "PY3",
-)
+    ] + select({
+        ":gapic_gen_python_3_6": [requirement("dataclasses")],
+        "//conditions:default": [],
+    }),
+)
```
software-dov, line 73: Nit: no newline

BUILD.bazel
```diff
@@ -23,14 +24,37 @@ pandoc_toolchain(
...
+
+config_setting(
+    name = "gapic_gen_python_3_6",
+    values = {"define": "gapic_gen_python=3.6"},
+)
+
+py_runtime(
+    name = "pyenv3_runtime",
+    interpreter = ":pyenv3wrapper.sh",
```
software-dov, line 37: Just to clarify, we're not spending O(minutes) building the correct python version, right?

pyenv3wrapper.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+HOME_DIR=$(getent passwd "$(whoami)" | cut -d: -f6)
```
software-dov, line 3: Isn't this already provided via `$HOME`?

BUILD.bazel
vam-google: Yes, we are not building anything here, simply using the preinstalled one.

pyenv3wrapper.sh
vam-google: Bazel commands, to make them isolated from the environment and thus reporducible, are executed under `env -` environment, i.e. in clean environment, where none of the env variables are set, including `$HOME` `$PATH` or anything else. That is why to find home directory we need to rely only on the core built-in commands (like `whoami`) and some trickery to get home dir path from that.

software-dov: Okay, good to know. TIL

PR #409, opened 2020-05-05
Template directories passed to options are turned into absolute paths

software-dov:


gapic/generator/options.py
```diff
@@ -99,7 +99,7 @@ def build(cls, opt_string: str) -> 'Options':
             os.path.join(os.path.dirname(__file__), '..', 'templates'),
         )
         templates = [
-            (default_path if path == default_token else path)
+            (default_path if path == default_token else os.path.abspath(path))
```
vam-google, line 102: I'm afraid it will not work as es expected in bazel environment, as it will resolve the relative path against the current working directory. In bazel it will be the directory in which the proto package is, while it should be relative to the location of the generator itself.
Can we change such that the path is resolved against `__FILE__`?

software-dov: I've changed it to something that I think works better in light of the restricted environment

vam-google: Thanks, that should work (theoretically)!

PR #420, opened 2020-05-20
feat: add GOOGLE_API_USE_MTLS support

arithmetic1728:
Implement the logic in go/python-apiary-mtls, section 2.2 (replaced `client_encrypted_cert_source` with `client_cert_source` because microgenerator doesn't use encrypted client key)

For the following logic, part 2 is implemented in Client constructor, part 1 is implemented in GrpcTransport constructor.
```
# 1. First determines the client cert to use.
client_cert_source_to_use = 
    client_options.client_cert_source or ADC_cert_source or None

# 2. Then determine the api endpoint to use.
if client_options.api_endpoint is not None:
    api_endpoint_to_use = client_options.api_endpoint
else:
    if GOOGLE_API_USE_MTLS == ‘Always’:
        client_options.api_endpoint = mtls_endpoint
    elif GOOGLE_API_USE_MTLS == ‘Never’:
        client_options.api_endpoint = regular_endpoint
    else: 
        if client_cert_source_to_use is None:
            client_options.api_endpoint = regular_endpoint
        else:
            client_options.api_endpoint = mtls_endpoint
```

### DIFF
--- a/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
@@ -2,6 +2,7 @@
 
 {% block content %}
 from collections import OrderedDict
+import os
 import re
 from typing import Callable, Dict, {% if service.any_server_streaming %}Iterable, {% endif %}{% if service.any_client_streaming %}Iterator, {% endif %}Sequence, Tuple, Type, Union
 import pkg_resources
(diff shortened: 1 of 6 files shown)


gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
```diff
@@ -144,21 +147,47 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+            google.auth.exceptions.MutualTLSChannelError: If mutual TLS transport
                 creation failed for any reason.
         """
         if isinstance(client_options, dict):
             client_options = ClientOptions.from_dict(client_options)
+        if client_options is None:
+            client_options = ClientOptions.ClientOptions()
+
+        if transport is None and client_options.api_endpoint is None:
+            use_mtls_env = os.getenv("GOOGLE_API_USE_MTLS", "Never")
```
software-dov, line 174: Should the value of `GOOGLE_API_USE_MTLS` be case insensitive? Is there a spec, or have any other generators implemented MTLS?

arithmetic1728: I think it should be case sensitive, please see go/mtls-endpoint-test.

PR #427, opened 2020-05-28
Integration tests for nonslash resource paths

software-dov:
Add integration tests based on showcase's non-slash separated resource paths.

Note: I was having editor formatting problems and eventually resorted to using 'apply black on save'. This is to blame for all the spurious string and formatting changes.

Note: Even though the changes in this commit imply support for proto3 optional fields, such support is not yet implemented. The 'support' is a workaround to allow the most recent showcase spec to compile.


gapic/generator/generator.py
```diff
@@ -92,31 +90,30 @@ def get_response(
...
+            self._generate_samples_and_manifest(
+                api_schema, self._env.get_template(sample_templates[0]),
+            )
+        )
 
         # Return the CodeGeneratorResponse output.
-        return CodeGeneratorResponse(file=[i for i in output_files.values()])
+        res = CodeGeneratorResponse(
+            file=[i for i in output_files.values()])  # type: ignore
+        res.supported_features |= CodeGeneratorResponse.Feature.FEATURE_PROTO3_OPTIONAL  # type: ignore
```
busunkim96, line 112: Just want to check that this is intentional (since the other changes in this file seem to be formatting)

tests/system/test_resource_crud.py
```diff
@@ -50,3 +48,27 @@ def test_path_methods(identity):
     actual = identity.user_path("bdfl")
 
     assert expected == actual
+
+
+def test_nonslash_resource(messaging):
+    expected = "users/bdfl/profile/blurbs/legacy/apocalyptic~city"
```
busunkim96, line 54: Out of curiosity, was does the `~` mean in a path?

software-dov: It's used in certain ads resources for reasons I don't understand. In this instance right here it's just intended to be not a slash.

gapic/generator/generator.py
software-dov: The formatting got fed through a log chipper due to (I think) a horrible interaction between internal Google formatting logic and custom opensource style formatting logic.
I've resolved the aggressive autoformating issue in emacs; this should hopefully be the last review where it is such an issue.

PR #443, opened 2020-06-04
fix: add field headers for other http verbs

### LINKED ISSUE #401: Generated clients are missing routing headers
https://aip.dev/client-libraries/4222

It looks like the generator is not adding variables declared in the `google.api.http` annotation to routing headers.

https://github.com/googleapis/googleapis/blob/52701da10fec2a5f9796e8d12518c0fe574488fe/google/cloud/servicedirectory/v1beta1/registration_service.proto#L60-L65

See https://github.com/GoogleCloudPlatform/python-docs-samples/pull/3456/files#r412305714 for the modifications that need to be made to currently use the library.

Possibly related to #380.

Metadata should be added [somewhere here](https://github.com/googleapis/python-service-directory/blob/25e0fcfb32d100be0cba3799d62543569dd2d2c6/google/cloud/servicedirectory_v1beta1/services/registration_service/client.py#L359-L378) (see [this method](https://github.com/googleapis/python-service-directory/blob/25e0fcfb32d100be0cba3799d62543569dd2d2c6/google/cloud/servicedirectory_v1beta1/services/registration_service/client.py#L432-L459) for comparison))

busunkim96:
Closes #401

### DIFF
--- a/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
@@ -117,7 +117,7 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
     from_service_account_json = from_service_account_file
 
 
-    {% for message in service.resource_messages -%}
+    {% for message in service.resource_messages|sort(attribute="resource_type") -%}
     @staticmethod
     def {{ message.resource_type|snake_case }}_path({% for arg in message.resource_path_args %}{{ arg }}: str,{% endfor %}) -> str:
         """Return a fully-qualified {{ message.resource_type|snake_case }} string."""
--- a/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
@@ -333,7 +333,9 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
         metadata = tuple(metadata) + (
             gapic_v1.routing_header.to_grpc_metadata((
                 {%- for field_header in method.field_headers %}
+                {%- if not method.client_streaming %}
                 ('{{ field_header }}', request.{{ field_header }}),
+                {%- endif %}
                 {%- endfor %}
             )),
         )
--- a/gapic/schema/wrappers.py
+++ b/gapic/schema/wrappers.py
@@ -628,20 +628,16 @@ class Method:
 
         pattern = re.compile(r'\{([a-z][\w\d_.]+)=')
 
-        if http.get:
-            return tuple(pattern.findall(http.get))
-        elif http.put:
-            return tuple(pattern.findall(http.put))
-        elif http.post:
-            return tuple(pattern.findall(http.post))
-        elif http.delete:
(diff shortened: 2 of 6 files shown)


gapic/schema/wrappers.py
```diff
@@ -625,8 +625,22 @@ def client_output(self):
...
+        elif http.put:
+            return tuple(pattern.findall(http.put))
+        elif http.post:
+            return tuple(pattern.findall(http.post))
+        elif http.delete:
+            return tuple(pattern.findall(http.delete))
+        elif http.patch:
+            return tuple(pattern.findall(http.patch))
+        elif http.custom.path:
+            return tuple(pattern.findall(http.custom.path))
```
software-dov, line 642: This screams "a pattern we can generalize."
How about
```python
potential_verbs = [
    http.get,
    http.put,
    http.post,
    http.delete,
    http.patch,
    http.custom.path,
]
return next((tuple(pattern.findall(verb)) for verb in potential verbs if verb), ())
```
?

tests/unit/schema/wrappers/test_method.py
```diff
@@ -204,12 +204,44 @@ def test_method_field_headers_none():
...
+def test_method_field_headers_present_delete():
+    http_rule = http_pb2.HttpRule(delete='/v1/{parent=projects/*}/topics')
+    method = make_method('DoSomething', http_rule=http_rule)
+    assert method.field_headers == ('parent',)
+
+
+def test_method_field_headers_present_patch():
+    http_rule = http_pb2.HttpRule(patch='/v1/{parent=projects/*}/topics')
+    method = make_method('DoSomething', http_rule=http_rule)
+    assert method.field_headers == ('parent',)
```
software-dov, line 234: I'm thinking we can also generalize these to something like
```python
verbs = [
    'get',
    'put',
    'post',
    'delete',
    'patch',
]

for v in verbs:
    rule = http_pb2.HttpRule(**{verb: '/v1/{parent=projects/*}/topics'})
    method = make_method('DoSomething', http_rule=rule)
    assert method.field_headers == ('parent',)
```

in a single test method.

software-dov:
Ahh, it looks like the generated unit tests aren't getting 100% coverage.
It should be easy to add the new verbs to the generated unit test or, if you're okay with my suggested implementation change, the problem goes away because of the simplified control flow.

gapic/schema/wrappers.py
busunkim96: Wasn't sure how to wrap this up nicely, thank you!

software-dov: I'm the champion of abusing `next` :)

gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
```diff
@@ -333,7 +333,9 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
         metadata = tuple(metadata) + (
             gapic_v1.routing_header.to_grpc_metadata((
                 {%- for field_header in method.field_headers %}
+                {%- if not method.client_streaming %}
```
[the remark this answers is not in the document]
software-dov: I was under the impression that headers for streaming methods don't make a lot of sense because they're not tightly bound to http semantics. If that's not correct, then I think we would need to ask someone who knows better.

Issue #410, opened 2020-05-07
Library unit test failing with struct_pb2.Value


dizcology:
Generate library with the following service proto:
```
// put this file in google/cloud/myservice/v1/

syntax = "proto3";

package myservice.v1;

import "google/protobuf/struct.proto";
import "google/api/client.proto";

service MyService {
  option (google.api.default_host) = "my.example.com";

  rpc MyMethod(MethodRequest) returns (MethodResponse) {
  	option (google.api.method_signature) = inputs,parameters";
  }
}

message MethodRequest {
  repeated google.protobuf.Value inputs = 1;
  google.protobuf.Value parameters = 2;
}

message MethodResponse {
  repeated google.protobuf.Value outputs = 1;
}
```

The generated unit test fails with:
```
# Merge in the value being set.
        if pb_value is not None:
>           self._pb.MergeFrom(self._meta.pb(**{key: pb_value}))
E           TypeError: Value must be iterable
```

Variant: commenting out the repeated `inputs` field, the generated unit test fails with a different error:
```
_, args, _ = call.mock_calls[0]
>           assert args[0].parameters == struct.Value(null_value=struct.NullValue.NULL_VALUE)
E           AssertionError: assert None == null_value: NULL_VALUE\n
E            +  where None = parameters {\n  null_value: NULL_VALUE\n}\n.parameters
E            +  and   null_value: NULL_VALUE\n = <class 'google.protobuf.struct_pb2.Value'>(null_value=0)
E            +    where <class 'google.protobuf.struct_pb2.Value'> = struct.Value
E            +    and   0 = <google.protobuf.internal.enum_type_wrapper.EnumTypeWrapper object at 0x10ac9ce50>.NULL_VALUE
E            +      where <google.protobuf.internal.enum_type_wrapper.EnumTypeWrapper object at 0x10ac9ce50> = struct.NullValue
```

software-dov:
Filed #413 and #414 to handle specific issues, closing this one in favor of the new two.

PR #445, opened 2020-06-05
fix: handle repeated fields in method signatures [gg-python]

### LINKED ISSUE #413: Method signature elements that are repeated fields are not handled properly
Minimal reproducible code:

```proto
service MyService {
  rpc MyMethod(MethodRequest) returns (MethodResponse) {
  	option (google.api.method_signature) = "inputs,parameters";
  }
}

message MethodRequest {
  repeated google.protobuf.Value inputs = 1;
}

message MethodResponse {
  google.protobuf.Value output = 1;
}
```

In the portion of the client method that constructs the request, the following lines occur 

```python
 if inputs is not None:
            request.inputs = inputs
```

This causes protobuf to raise an error because `request.intputs` is a repeated field.
The solution is to do the following instead:
```python
  if inputs is not None:
          request.inputs.extend(inputs)
```

miraleung:
Fixes #413. Tested this locally with a proto like the one in #410 (although this doesn't solve the other, non-repeated error in that issue). There don't seem to be applicable unit tests, but please let me know if there are and I can add more.


software-dov:
Change looks good, but I would be more comfortable approving the change if there were a repeated, flattened field in showcase to test against.

software-dov (approved):

Issue #432, opened 2020-06-02
Clients should support credentials_file via client_options


busunkim96:
Clients should allow credentials_file to be specified through client_options. This `credentials_file` can be a service account json file or an authorized user file.

Existing methods on the client `from_service_account_json` and `from_service_account_file` will be kept. https://googleapis.dev/python/texttospeech/latest/gapic/v1/api.html#google.cloud.texttospeech_v1.TextToSpeechClient.from_service_account_file 

```py
from google.cloud import texttospeech

client = texttospeech.TextToSpeechClient(
    client_options={"credentials_file": "path/to/credentials.json"}
)
```
Internal link: go/extensible-client-options-exit

busunkim96:
There's a function in the auth library the reads credentials from a file that we can use for this

https://github.com/googleapis/google-auth-library-python/blob/master/google/auth/_default.py#L72

PR #461, opened 2020-06-16
feat: add `credentials_file` and `scopes` via `client_options`

### LINKED ISSUE #433: Clients should support scope override via client_options
Clients should allow `scopes` to be set via `client_options`.

The scopes provided should override the scopes of the OAuth access token.

```py
from google.cloud import texttospeech

client = texttospeech.textToSpeechClient(
    client_options={"scopes": ["..", "..."]}
)

```
Internal link: go/extensible-client-options-exit

Many credentials types prohibit modifying the scopes on the original credential, so we will use `with_scopes`.

busunkim96:
Closes #433 and #432 

Depends on:
- [x] https://github.com/googleapis/python-api-core/pull/50
- [x] https://github.com/googleapis/google-auth-library-python/pull/530

TODO:

- [x] Add unit tests
- [x] Update Async Client (after #365 is merged)
- [ ] Copy to ads templates


busunkim96:
@software-dov Any idea what's up with the showcase tests?

PR #475, opened 2020-06-19
fix: only require dataclases if python<3.7

busunkim96:
The [dataclasses package](https://pypi.org/project/dataclasses/) now only supports 3.6, which creates some weirdness with the requirements.txt requiring `0.7.0`.


requirements.txt
```diff
@@ -1,9 +1,9 @@
 click==7.1.2
-google-api-core==1.20.1
+google-api-core==1.21.0
```
busunkim96, line 2: Not sure why this is showing up in the diff, but this is already `1.21.0` in master.

https://github.com/googleapis/gapic-generator-python/blob/ed244b31bd885c372a65df8e14c5759e6b5a16a4/requirements.txt#L2

requirements.txt
```diff
@@ -1,9 +1,9 @@
...
 googleapis-common-protos==1.52.0
 jinja2==2.11.2
 MarkupSafe==1.1.1
 protobuf==3.12.2
 pypandoc==1.5
 PyYAML==5.3.1
-dataclasses==0.7
\ No newline at end of file
+dataclasses==0.7; python_version < '3.7'
```
vam-google, line 9: This should be 0.6 instead, otherwise it will fail bazel build.

busunkim96: 🤦‍♀️ Fixed!

requirements.txt
```diff
@@ -6,4 +6,4 @@ MarkupSafe==1.1.1
 protobuf==3.12.2
 pypandoc==1.5
 PyYAML==5.3.1
-dataclasses==0.7
\ No newline at end of file
+dataclasses==0.6; python_version < '3.7'
```
vam-google, line 9: LGTM, assuming this syntax is understood by bazel python rules properly (the python_version < '3.7' part). Note, requirements.txt is currently used only by bazel.

alexander-fenster: apparently it does not :(  (or something else is wrong)

Issue #484, opened 2020-06-25
oneof proto not generating as oneof


crwilcox:
```diff
--- a/google/cloud/firestore_v1/types/document.py
+++ b/google/cloud/firestore_v1/types/document.py
@@ -128,19 +128,19 @@ class Value(proto.Message):
             A map value.
     """

-    null_value = proto.Field(proto.ENUM, number=11, enum=struct.NullValue,)
-    boolean_value = proto.Field(proto.BOOL, number=1)
-    integer_value = proto.Field(proto.INT64, number=2)
-    double_value = proto.Field(proto.DOUBLE, number=3)
+    null_value = proto.Field(proto.ENUM, number=11, enum=struct.NullValue, oneof="value_type")
+    boolean_value = proto.Field(proto.BOOL, number=1, oneof="value_type")
+    integer_value = proto.Field(proto.INT64, number=2, oneof="value_type")
+    double_value = proto.Field(proto.DOUBLE, number=3, oneof="value_type")
     timestamp_value = proto.Field(
-        proto.MESSAGE, number=10, message=timestamp.Timestamp,
+        proto.MESSAGE, number=10, message=timestamp.Timestamp, oneof="value_type"
     )
-    string_value = proto.Field(proto.STRING, number=17)
-    bytes_value = proto.Field(proto.BYTES, number=18)
-    reference_value = proto.Field(proto.STRING, number=5)
-    geo_point_value = proto.Field(proto.MESSAGE, number=8, message=latlng.LatLng,)
-    array_value = proto.Field(proto.MESSAGE, number=9, message="ArrayValue",)
-    map_value = proto.Field(proto.MESSAGE, number=6, message="MapValue",)
+    string_value = proto.Field(proto.STRING, number=17, oneof="value_type")
+    bytes_value = proto.Field(proto.BYTES, number=18, oneof="value_type")
+    reference_value = proto.Field(proto.STRING, number=5, oneof="value_type")
+    geo_point_value = proto.Field(proto.MESSAGE, number=8, message=latlng.LatLng, oneof="value_type")
+    array_value = proto.Field(proto.MESSAGE, number=9, message="ArrayValue", oneof="value_type")
+    map_value = proto.Field(proto.MESSAGE, number=6, message="MapValue", oneof="value_type")
```

[...]

PR #485, opened 2020-06-25
fix: add oneof fields to generated protoplus init

crwilcox:
Fixes: https://github.com/googleapis/gapic-generator-python/issues/484


crwilcox:
It seems perhaps the data for oneofs may not be captured?

The template ought to be right provided oneof makes it to the field.

https://github.com/googleapis/gapic-generator-python/blob/920e419d4fb827f4cb154d0a38ffab83c19a0675/gapic/schema/api.py#L471

https://github.com/googleapis/gapic-generator-python/blob/920e419d4fb827f4cb154d0a38ffab83c19a0675/gapic/schema/api.py#L782

BenRKarl:
@crwilcox is the issue here that oneof fields aren't present at all in your generated classes? If so I think we have the same problem for Ads, if a change is made here can we also change the ads templates? See [here](https://github.com/googleapis/gapic-generator-python/blob/master/gapic/ads-templates/%25namespace/%25name/%25version/%25sub/types/_message.py.j2#L46)

crwilcox:
@BenRKarl For sure. Thanks for pointing that out. Still need to figure out piping the oneof data down to the field I think.

gapic/schema/api.py
```diff
@@ -804,6 +844,7 @@ def _load_message(self,
                 address=address,
                 documentation=self.docs.get(path, self.EMPTY),
             ),
+            oneofs=oneofs,
```
crwilcox, line 847: @software-dov  I didn't use this anywhere, but oneofs are part of a message so I thought having it here may serve useful

gapic/schema/api.py
```diff
@@ -586,6 +615,10 @@ def _get_fields(self,
         # `_load_message` method.
         answer: Dict[str, wrappers.Field] = collections.OrderedDict()
         for field_pb, i in zip(field_pbs, range(0, sys.maxsize)):
+            oneof_name = None
+            if getattr(field_pb, 'oneof_index', -1) >= 0 and oneofs:
```
lukesneeringer, line 619: Oh goodness, protobuf usability at its finest. 🤣

gapic/schema/api.py
```diff
@@ -556,14 +556,43 @@ def _load_children(self,
...
+            path (Tuple[int]): The source location path thus far, as
+                understood by ``SourceCodeInfo.Location``.
+
+        Returns:
+            Mapping[str, ~.wrappers.Oneof]: A ordered mapping of
+                :class:`~.wrappers.Oneof` objects.
+        """
+        # Iterate over the oneofs and collect them into a dictionary.
+        answer: Dict[str, wrappers.Oneof] = collections.OrderedDict()
+        for oneof_pb, i in zip(oneof_pbs, range(0, sys.maxsize)):
```
software-dov, line 579: Style nit: prefer
```python
answer = collections.OrderedDict(
    (oneof_pb.name, wrappers.Oneof(oneof_pb=oneof_pb))
    for i, oneof_pb in enumerate(oneof_pbs)
)
```

gapic/schema/api.py
```diff
@@ -556,14 +556,43 @@ def _load_children(self,
...
+                oneof_pb=oneof_pb,
+            )
+
+        # Done; return the answer.
+        return answer
+
     def _get_fields(self,
                     field_pbs: Sequence[descriptor_pb2.FieldDescriptorProto],
                     address: metadata.Address, path: Tuple[int, ...],
+                    oneofs=None
```
software-dov, line 590: Type annotations nit :P

gapic/schema/api.py
software-dov: 🤮

gapic/schema/api.py
software-dov: Can't hurt.

PR #486, opened 2020-06-25
docs: add test run info to contributing and check protoc version

busunkim96:


noxfile.py
```diff
@@ -47,6 +51,27 @@ def unit(session):
     )
 
 
+def check_protoc(session):
+    """Make sure the Protocol Compiler (protoc) is installed"""
+    session.install("packaging")
```
busunkim96, line 56: I just realized this won't work for making sure `packaging` is available in the noxfile. 😅  Will modify

PR #505, opened 2020-07-09
fix: tweak oneof detection

software-dov:
Oneof detection and assignment to fields is tricky.
This patch fixes detection of oneof fields,
fixes uses in generated clients
and tweaks generated tests to use them correctly.


gapic/schema/api.py
```diff
@@ -614,8 +615,18 @@ def _get_fields(self,
         # first) and this will be None. This case is addressed in the
         # `_load_message` method.
         answer: Dict[str, wrappers.Field] = collections.OrderedDict()
+
+        def oneof_p(field_pb):
+            # This is the _only_ way I have found to determine whether
```
lukesneeringer, line 620: Sanity check with haberman@.

software-dov: Resolved offline and updated

PR #548, opened 2020-07-23
fix: use context manager for mtls env var

busunkim96:
git seems to be producing a hard to read diff, but the only change I've made is to use `with mock.patch.dict(...)` to modify the environment variable `GOOGLE_API_USE_MTLS`. 

I deleted `del os.environ["GOOGLE_API_USE_MTLS"]` in #496 without realizing there were more tests above,  and this group of tests started failing for me locally after the latest release.


arithmetic1728:
@busunkim96 Sorry this is essentially caused by a problem of the showcase test. I just noticed `client.DEFAULT_ENDPOINT` and `client_DEFAULT_MTLS_ENDPOINT` happen to be the same (`localhost`) for showcase unit tests. This is the reason why for your PR #496 the showcase unit tests can pass, but any generated client unit test fail. Just sent a PR #551 to address this issue.

PR #551, opened 2020-07-23
chore: distinguish default/mtls endpoint for tests

arithmetic1728:
`showcase` has default endpoint `localhost:7469`, so mtls endpoint is the same. When we interchange `client.DEFAULT_ENDPOINT` and `client.DEFAULT_MTLS_ENDPOINT` by mistake, the unit tests still pass. This causes problem, because generated client lib always has `*.googleapis.com` as the default endpoint, so mtls endpoint is always different (`*.mtls.googleapis.com`), so client lib unit test always fail. This PR replaces the default endpoint in case it is `localhost` (as it does in `showcase` unit tests) so we can distinguish the endpoints in showcase unit test.


gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -77,6 +84,8 @@ def test_{{ service.client_name|snake_case }}_get_transport_class():
     assert transport == transports.{{ service.name }}GrpcTransport
 
 
+@mock.patch.object({{ service.client_name }}, "DEFAULT_ENDPOINT", modify_default_endpoint({{ service.client_name }}))
+@mock.patch.object({{ service.async_client_name }}, "DEFAULT_ENDPOINT", modify_default_endpoint({{ service.async_client_name }}))
```
software-dov, line 88: I don't think the async client has been ported to the ads templates.

arithmetic1728: removed

PR #587, opened 2020-09-03
fix: rendering mock values for recursive messages no longer crashes

software-dov:
Protobuf allows recursive message types, i.e. messages whose fields
are of the same type as the message itself.

```proto
message Foo {
    Foo foo = 1; // Degenerate case
}
```

A real world example is bigquery.v2.data:RowFilter

These recursive types cause a problem when trying to render 
mock values for unit tests because there's no inherent limit on 
when to stop rendering nested values.
The solution in this commit is an artifical cap on the depth of
recursion in rendering mock values.

### DIFF
--- a/gapic/schema/wrappers.py
+++ b/gapic/schema/wrappers.py
@@ -56,6 +56,8 @@ class Field:
     )
     oneof: Optional[str] = None
 
+    MAX_MOCK_DEPTH: int = 20    # Arbitrary cap set via heuristic rule of thumb.
+
     def __getattr__(self, name):
         return getattr(self.field_pb, name)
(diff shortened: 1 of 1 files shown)


busunkim96:
Does it make sense to add a message like this to https://github.com/googleapis/gapic-showcase?

arithmetic1728 (approved):

software-dov:
It definitely sounds like a good idea to add a recursive showcase message type. I'll file an issue there.

software-dov:
Showcase issue tracked by https://github.com/googleapis/gapic-showcase/issues/449

PR #641, opened 2020-10-08
fix: fix types on server and bidi streaming callables

dpcollins-google:
The return type is an Awaitable that produces an AsyncIterator.


lidizheng (approved):
This PR fixes a type annotation error I made. For bidi streaming calls, the asyncio application needs to explicitly invoke `await` to obtain the `AsyncIterable` responses (see [example](https://github.com/googleapis/gapic-generator-python/blob/517118cd3e924b97be8140e58426c0a723633d78/tests/system/test_grpc_streams.py#L183)). This `await` is used to establish connection to peer, so the retry can happen immediately instead of being postponed till the first message read. LGTM.

software-dov (approved):

PR #645, opened 2020-10-09
fix: expose transport property for clients

### LINKED ISSUE #640: Expose operations_client
To use the operations client that you have to go through `client._transport`. It feels weird to recommend using a variable that looks like it's intended for internal use only. 

Published AutoML sample: https://github.com/googleapis/python-automl/blob/master/samples/snippets/get_operation_status.py
```py
def get_operation_status(operation_full_id):
    """Get operation status."""
    # [START automl_get_operation_status]
    from google.cloud import automl

    # TODO(developer): Uncomment and set the following variables
    # operation_full_id = \
    #     "projects/[projectId]/locations/us-central1/operations/[operationId]"

    client = automl.AutoMlClient()
    # Get the latest state of a long-running operation.
    response = client._transport.operations_client.get_operation(
        operation_full_id
    )

    print("Name: {}".format(response.name))
    print("Operation details:")
    print(response)
    # [END automl_get_operation_status]
```

It looks like the old generator left`transport` public. Was that changed intentionally? https://github.com/googleapis/gapic-generator/blob/17fbb3b38a20d13e69879d845a045dc1f8448a3d/src/test/java/com/google/api/codegen/gapic/testdata/py/python_multiple_services.baseline#L999

If we want to keep `transport` private, could we make `operations_client` public somewhere else?

software-dov:
Sometimes it's useful to get a reference to the transport for a client object.

Closes #640

### DIFF
--- a/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
@@ -118,6 +118,15 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
 
     from_service_account_json = from_service_account_file
 
+    @property
+    def transport(self) -> {{ service.name }}Transport:
+        """Return the transport used by the client instance.
+
+        Returns:
+            {{ service.name }}Transport: The transport used by the client instance.
+        """
+        return self._transport
+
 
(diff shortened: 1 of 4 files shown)


busunkim96 (approved):
Thank you!

arithmetic1728 (approved):

PR #665, opened 2020-10-16
test: use built-in mock module for tests

tswast:
Since Python 3.6+ is required, there is no need to use the mock backport.

### DIFF
--- a/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
+++ b/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
@@ -2,7 +2,7 @@
 
 {% block content %}
 import os
-import mock
+from unittest import mock
 
 import grpc
 from grpc.experimental import aio


busunkim96 (commented):
It looks like this is breaks the async tests

```
[gw0] linux -- Python 3.7.9 /root/project/.nox/showcase_unit-3-7/bin/python

transport = 'grpc_asyncio'

    @pytest.mark.asyncio
    async def test_connect_async(transport: str = 'grpc_asyncio'):
        client = MessagingAsyncClient(
            credentials=credentials.AnonymousCredentials(),
            transport=transport,
        )
    
        # Everything is optional in proto3 as far as the runtime is concerned,
        # and we are mocking out the actual API, so just send an empty request.
        request = messaging.ConnectRequest()
    
        requests = [request]
    
        with mock.patch.object(
                type(client._client.transport.connect),
                '__call__') as call:
            # Designate an appropriate return value for the call.
            call.return_value = mock.Mock(aio.StreamStreamCall, autospec=True)
>           call.return_value.read = mock.AsyncMock(side_effect=[messaging.StreamBlurbsResponse()])
E           AttributeError: module 'unittest.mock' has no attribute 'AsyncMock'

tests/unit/gapic/showcase_v1beta1/test_messaging.py:3076: AttributeError
```

busunkim96:
@software-dov It doesn't look like codeowners kicked in properly. Does `actools-python` exist? https://github.com/googleapis/gapic-generator-python/blob/964544ea5ecf9788cbe1caba8f76c0320d1ed6ba/.github/CODEOWNERS#L7

software-dov (approved):

software-dov:
Noah created it yesterday. I'll ask him if there's any lag or anything.

tswast:
Ah, bummer. I guess we do still need the backfill library. AsyncMock is "New in version 3.8"

Closing

PR #684, opened 2020-10-23
perf: collisions don't contain reserved names by default

software-dov:
The 'collisions' set does NOT contain RESERVED_NAMES; they are
combined at runtime when needed.
    
For a large real-world, this results in an order of magnitude
reduction in memory usage.
I'm not joking: Google Ads v5 uses 2.45 GB peak before this change
and 223 MB after.

Also contains changes to add `__slots__` attributes to Metadata and Address. These are ancillary, optional, and open to negotiation.
In the above scenario, they reduce memory usage from 223 MB to 177 MB. If other people feel that the reduction in readability does not warrant the reduction in memory usage I am absolutely open to dropping that particular commit.

Includes other minor memory usage optimizations that collectively shave about 5 MB.

### DIFF
--- a/gapic/schema/metadata.py
+++ b/gapic/schema/metadata.py
@@ -49,9 +49,6 @@ class Address:
     )
     collisions: FrozenSet[str] = dataclasses.field(default_factory=frozenset)
 
-    def __post_init__(self):
-        super().__setattr__("collisions", self.collisions | RESERVED_NAMES)
-
     def __eq__(self, other) -> bool:
         return all([getattr(self, i) == getattr(other, i) for i
                     in ('name', 'module', 'module_path', 'package', 'parent')])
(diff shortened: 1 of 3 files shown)


gapic/schema/metadata.py
```diff
@@ -114,7 +148,9 @@ def module_alias(self) -> str:
         while still providing names that are fundamentally readable
         to users (albeit looking auto-generated).
         """
-        if self.module in self.collisions:
+        # Don't keep RESERVED_NAMES in self.collisions,
+        # just combine the two when necessary.
```
busunkim96, line 152: Perhaps add a note that this is for memory usage?

busunkim96 (approved):
> Also contains changes to add __slots__ attributes to Metadata and Address. These are ancillary, optional, and open to negotiation.
> In the above scenario, they reduce memory usage from 223 MB to 177 MB. If other people feel that the reduction in readability does not warrant the reduction in memory usage I am absolutely open to dropping that particular commit.

I would personally prefer preserving readability.

PR #702, opened 2020-11-20
feat: add proper handling of query/path/body parameters for rest transport

### LINKED ISSUE #267: Re-add HTTP transport and make it the fallback
Some users may not want to use gRPC. Http is a sufficient fallback for certain APIs, so this transport must be explicitly supported.
This support used to exist but was removed due to testing limitations.
See https://github.com/googleapis/gapic-generator-python/pull/114 for additional details.

yon-mg:
Implementation and tests assume basic case of grpc transcoding for handling of all query/path parameters and body field.

These changes are dependent upon a PR in googleapis/proto-plus-python repo:
https://github.com/googleapis/proto-plus-python/pull/164


gapic/schema/wrappers.py
```diff
@@ -767,6 +767,23 @@ def http_opt(self) -> Optional[Dict[str, str]]:
...
+        return [x[1:-1] for x in re.findall(pattern, self.http_opt['url'])]
+
+    @property
+    def query_params(self) -> Set[str]:
+        if self.http_opt is None:
+            return set()
+        body = []
+        if 'body' in self.http_opt.keys():
+            body.append(self.http_opt['body'])
+        all_params = self.input.fields.keys()
```
software-dov, line 784: We can move the variable setups around a little to make it a little easier to read.
```python
params = set(self.path_params)
body = self.http_opt.get('body')
if body:
    params.add(body)

return set(self.input.fields) ^ params
```
Operations on dicts by default just use the keyspace, e.g. `assert 6 in {6: "six"}` and `assert set({6: "six"}) == {6}`.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -135,28 +135,42 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
...
+
+        potentialParams = {
+            {%- for field in method.query_params %}
+            '{{ field|camel_case }}': request.{{ field }},
+            {%- endfor %}
+        }
+        potentialParams = {k: v for k, v in potentialParams.items() if v}
+        for i, (param_name, param_value) in enumerate(potentialParams.items()):
+            q = '?' if i == 0 else '&'
+            url += q + param_name + '=' + str(param_value).replace(' ', '+')
```
software-dov, line 169: Nit: prefer format strings to concatenating using +. It's easier to eyeball parse for longer or more convoluted strings.
```python
url += "{q}{name}={value}".format(q=q, name=param_name, value=param_value.replace(' ', '+'))
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -135,28 +135,42 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
...
+
         {% if not method.void %}response = {% endif %}self._session.{{ method.http_opt['verb'] }}(
-            url,
-            {%- if 'body' in method.http_opt.keys() %}
+            url
+            {%- if 'body' in method.http_opt.keys() %},
```
software-dov, line 173: No need to use `keys` here, `if 'body' in method.http_opt` is more idiomatic.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -135,28 +135,42 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
...
             {{ field }}=request.{{ field }},
             {%- endfor %}
         )
+
+        potentialParams = {
+            {%- for field in method.query_params %}
+            '{{ field|camel_case }}': request.{{ field }},
+            {%- endfor %}
+        }
+        potentialParams = {k: v for k, v in potentialParams.items() if v}
```
software-dov, line 166: Very minor premature optimization nit: since we're immediately iterating over and then discarding this dictionary, we can turn it into a generator instead and prevent looping over the same data multiple times.
```python
potential_params = ((k, v) for k, v in potentialParams.items() if v) # The parentheses make this a generator expression.
for i, (param_name, param_value) in enumerate(potentialParams):
```

[This](https://wiki.python.org/moin/Generators) is a good rundown on generators and generator expressions. Dave Beazley also has a really fun youtube talk on generators.

gapic/utils/case.py
```diff
@@ -45,3 +45,17 @@ def to_snake_case(s: str) -> str:
...
+    This is provided to templates as the ``camel_case`` filter.
+
+    Args:
+        s (str): The input string, provided in snake case.
+    
+    Returns:
+        str: The string in camel case with the first letter unchanged.
+    '''
+    items = s.split('_')
+    return items[0] + "".join([x.capitalize() for x in items[1:]])
```
software-dov: Nit: no need to make a list as the argument to join. We could replace the square brackets with parentheses to make it a generator expression, but there's a minor syntactic optimization where if a generator expression is the sole function argument you can remove the parens.
```python
join(x.capitalize() for x in items[1:])
```

gapic/generator/generator.py
```diff
@@ -288,6 +289,8 @@ def _render_template(
                         opts=opts,
                     )
                 )
+                #for method in service.methods.values():
+                    #breakpoint()
```
software-dov, line 293: Just a reminder to clean this up.

software-dov (commented):
Looks good so far, just took an early readthrough.

gapic/utils/case.py
```diff
@@ -45,3 +45,17 @@ def to_snake_case(s: str) -> str:
 
     # Done; return the camel-cased string.
     return s.lower()
+
+def to_camel_case(s: str) -> str:
```
vam-google, line 49: Are there any built-in/standard python functions which do the same? Please prefer using standard ones to custom, if there are any.

gapic/schema/wrappers.py
```diff
@@ -767,6 +767,23 @@ def http_opt(self) -> Optional[Dict[str, str]]:
         # TODO(yon-mg): enums for http verbs?
         return answer
 
+    @property
+    def path_params(self) -> Sequence[str]:
+        if self.http_opt is None:
+            return []
+        pattern = r'\{\w+\}'
```
vam-google, line 774: Please add a comment that this handles grpc encoding in its simples case

vam-google (commented):
Overall looks good, please try to get LGTM when Dov comes back and push this

gapic/utils/case.py
software-dov: I'm fairly certain there is no standard library function for to camel-case.

gapic/schema/wrappers.py
```diff
@@ -767,6 +767,29 @@ def http_opt(self) -> Optional[Dict[str, str]]:
...
+        # TODO(yon-mg): fully implement grpc transcoding (currently only handles basic case)
+        if self.http_opt is None:
+            return []
+        pattern = r'\{\w+\}'
+
+        return [x[1:-1] for x in re.findall(pattern, self.http_opt['url'])]
+
+    @property
+    def query_params(self) -> Set[str]:
+        """Return query parameters for API call as determined by http annotation and grpc transcoding"""
```
vchudnov-g, line 782: !! Q: can you know all the query parameters ahead of time? If a user sets a field nested inside a message that is not listed in the path template nor in the body part of the annotation, it should be reported as a query param, but presumably in dotted notation (see my other comment about dotted notation being used in practice for the body even though disallowed in theory).

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -133,30 +133,44 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
         {%- endif %}
         """
 
-        {%- if 'body' in method.http_opt.keys() %}
+        {%- if 'body' in method.http_opt %}
         # Jsonify the input
-        data = {{ method.output.ident }}.to_json(
-            {%- if method.http_opt['body'] == '*' %}
+        {%- if method.http_opt['body'] != '*' %}
+        data = {{ method.input.fields[method.http_opt['body']].type.ident }}.to_json(
```
vchudnov-g, line 139: nit: maybe have the generated code call this variable `body` rather than `data` so the code is more self-explanatory.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -133,30 +133,44 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
...
+
+        potentialParams = {
+            {%- for field in method.query_params %}
+            '{{ field|camel_case }}': request.{{ field }},
+            {%- endfor %}
+        }
+        potentialParams = ((k, v) for k, v in potentialParams.items() if v)
+        for i, (param_name, param_value) in enumerate(potentialParams):
+            q = '?' if i == 0 else '&'
+            url += "{q}{name}={value}".format(q=q, name=param_name, value=param_value.replace(' ', '+'))
```
vchudnov-g, line 169: !! Does this ensure values are URL-encoded? If not, at least add a TODO.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -133,30 +133,44 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
...
             host=self._host,
-            {%- for field in method.input.fields.keys() %}
+            {%- for field in method.path_params %}
             {{ field }}=request.{{ field }},
             {%- endfor %}
         )
+
+        potentialParams = {
+            {%- for field in method.query_params %}
+            '{{ field|camel_case }}': request.{{ field }},
```
vchudnov-g, line 163: Sounds like this should work for nested fields (dot notation). Correct?

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -133,30 +133,44 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
...
+            {%- endfor %}
+        }
+        potentialParams = ((k, v) for k, v in potentialParams.items() if v)
```
vchudnov-g, line 166: It would probably be clearer to s/`potentialParams`/`queryParams`/

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -133,30 +133,44 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
...
         # Send the request
         url = 'https://{host}{{ method.http_opt['url'] }}'.format(
             host=self._host,
-            {%- for field in method.input.fields.keys() %}
+            {%- for field in method.path_params %}
             {{ field }}=request.{{ field }},
             {%- endfor %}
         )
+
+        potentialParams = {
```
vchudnov-g, line 161: It would probably be clearer to s/`potentialParams`/`queryParams`/g

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
vchudnov-g, line 166: This is fine, but another approach is to `potentialParams=["{k}={v}".format(k=k, v=v) for k, v in ...]` here and then `url += "?{}".format("&".join(potentialParams))` below

vchudnov-g, line 166: !! re: `if v`
Is the Python truthiness of `v` what we want here? I don't know, but I wonder given that we want to transmit only explicitly-set values, even if they match the defaults, which are falsy.

For any data type in the Python data model:
- if the unset value is null, then we can test for null to determine whether to send it. Simply checking for falsiness would not make this distinction.
- if the unset value is non-null, is there some other way to distinguish between values that were explicitly set and those that weren't?

We should talk about this, and at the very least insert a TODO for now.

gapic/utils/case.py
```diff
@@ -45,3 +45,19 @@ def to_snake_case(s: str) -> str:
 
     # Done; return the camel-cased string.
     return s.lower()
+
+
+def to_camel_case(s: str) -> str:
+    '''Convert any string to camel case.
```
vchudnov-g, line 51: nit: In this and the previous, pre-existing function, we're not touching spaces, right? Might be worth mentioning that.

gapic/utils/case.py
```diff
@@ -45,3 +45,19 @@ def to_snake_case(s: str) -> str:
...
+    This is provided to templates as the ``camel_case`` filter.
+
+    Args:
+        s (str): The input string, provided in any sane case system
+
+    Returns:
+        str: The string in lower camel case.
+    '''
+
+    items = re.split(r'[_-]', to_snake_case(s))
```
vchudnov-g, line 62: I'm surprised by the hyphen `-`. It almost seems like that should be handled by `to_snake_case`, though that's outside the scope of this PR. Does it come up in practice, that an input to this function has hyphens/kebab-case?

Opinion, @software-dov ?

gapic/schema/wrappers.py
```diff
@@ -767,6 +767,29 @@ def http_opt(self) -> Optional[Dict[str, str]]:
...
+        """Return the path parameters found in the http annotation url"""
+        # TODO(yon-mg): fully implement grpc transcoding (currently only handles basic case)
+        if self.http_opt is None:
+            return []
+        pattern = r'\{\w+\}'
+
+        return [x[1:-1] for x in re.findall(pattern, self.http_opt['url'])]
+
+    @property
+    def query_params(self) -> Set[str]:
```
vchudnov-g, line 781: Just curious: why can't  `query_params`  and `path_params` return the same type, rather than a `Set` and a `Sequence`?

vchudnov-g (changes requested):
I flagged my more important comments with `!!`

It looks like this generator does not have golden files or baselines, correct? Too bad; it would be good to see a sample of the generated GAPIC showing off these features.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -135,28 +135,42 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
 
         {%- if 'body' in method.http_opt.keys() %}
         # Jsonify the input
-        data = {{ method.output.ident }}.to_json(
-            {%- if method.http_opt['body'] == '*' %}
+        {%- if method.http_opt['body'] != '*' %}
+        data = {{ method.input.fields[method.http_opt['body']].type.ident }}.to_json(
```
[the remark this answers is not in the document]
yon-mg: It does not handle it but I have added a todo in the appropriate place in `wrappers.Method`. Pretty much anything dealing with full handling of grpc transcoding (or special cases outside grpc transcoding) is not yet taken care of in this PR.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
yon-mg: Short answer is yes and your second bullet point applies. Afaik from debugging it, there isn't actually a way to tell if a value was unset or set to default value since when the `<InputType>Request` object is made and passed to this generated method as a parameter, its values seem set to default ones. I also believe this is intended. For example, if I made a request but set a query param value to an empty string, simply checking if a value was set would allow that empty string to be included as part of the path_url and may result in a `'Bad Request'` response.

gapic/schema/wrappers.py
```diff
@@ -767,6 +767,23 @@ def http_opt(self) -> Optional[Dict[str, str]]:
...
+        if self.http_opt is None:
+            return set()
+        body = []
+        if 'body' in self.http_opt.keys():
+            body.append(self.http_opt['body'])
+        all_params = self.input.fields.keys()
+        return set(all_params) ^ set(body + list(self.path_params))
```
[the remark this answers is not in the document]
yon-mg: I agree you're right and I'll change it but there shouldn't be a difference between difference and symmetric difference here since all `body + path_params` should implicitly be included in all_params. I suppose that's why it hasn't failed for me but I'm not sure why I didn't just use `-` in the first place.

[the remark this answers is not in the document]
yon-mg: Not really here no but I have a todo for that as well as including in the pr description.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
yon-mg: Good point! Will do that

gapic/utils/case.py
yon-mg: It actually might handle it: I didn't check. As far as use, it doesn't get used but I wouldn't want a future user to attempt o try it and have it fail.

gapic/schema/wrappers.py
yon-mg: I suppose they could. Just thought it's more appropriate since ordering seems important for `path_params` but not for `query_params`.

gapic/utils/case.py
software-dov: My vote: write a test for `to_snake_case`'s behavior, add the functionality if the test fails, just split on '_'.

gapic/schema/wrappers.py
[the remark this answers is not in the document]
software-dov: I have no strong feeling for whether we want difference or XOR; Victor and the spec should take priority there. +1 for adding a test that explicitly checks for the one that we actually want and NOT the other one.

Side note: IMHO the set operators are not very clear, and especially if doing more complex set operations it's more reader friendly to use the named methods.

gapic/schema/wrappers.py
```diff
@@ -767,6 +767,29 @@ def http_opt(self) -> Optional[Dict[str, str]]:
...
 
+    @property
+    def path_params(self) -> Sequence[str]:
+        """Return the path parameters found in the http annotation url"""
+        # TODO(yon-mg): fully implement grpc transcoding (currently only handles basic case)
+        if self.http_opt is None:
+            return []
+        pattern = r'\{\w+\}'
+
+        return [x[1:-1] for x in re.findall(pattern, self.http_opt['url'])]
```
software-dov, line 778: We can move more of the logic into the regex by using capturing.
```python
param_re = r"\{(\w+)\}"
return re.findall(param_re, self.http_opt['url']) if self.http_opt else []
```

gapic/utils/case.py
vchudnov-g: If `to_snake_case` does not handle it, could adding the functionality break existing clients? is there an easy way to check? (This is why I was suggesting "outside the scope of this PR")

yon-mg: I doubt it would break anything since it's additive but this is why I didn't change `to_snake_case`'s behavior

gapic/schema/wrappers.py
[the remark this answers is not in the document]
software-dov: Replying to Victor's ancestral comment: the generated unit tests and 100% coverage thereof is intended to fill a similar goal as golden tests without making people tear their hair out from frustration. Occasionally it is helpful to diff surfaces, though; I like to either stash changes or do sub-branches, invoke the generator on showcase to save the resulting GAPIC, then ediff the relevant files in the two outputs. It can be a little fiddly, but it's *_far_* easier to do as a tactical move than to have golden tests.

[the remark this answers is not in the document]
vchudnov-g: For reviews, it's _very_ helpful to be able to diff the generated surfaces. That's the main thing I miss about golden files. The frustration, I can do without. ;-)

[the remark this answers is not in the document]
vchudnov-g: For this particular comment, as Yon and I discussed yesterday: `body + list(self.path_params)` should be a subset of `all_params`, so difference and xor should behave the same. However, difference expresses the intent better and it makes it more likely we can catch or prevent errors if this assumption is violated due to a bug (now or in the future).

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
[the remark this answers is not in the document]
vchudnov-g: OK. As discussed, we should err on the side of duplicate TODOs rather than too few. If we address one, we're likely to see the others and address/delete them.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -135,28 +135,42 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
...
+            request.{{ method.http_opt['body'] }},
+            including_default_value_fields=False
+        )
+        {%- else %}
+        data = {{ method.input.ident }}.to_json(
```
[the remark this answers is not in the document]
vchudnov-g: As we clarified yesterday: params in the path may be (but don't have to be) repeated in the body if we're using `"*"`, though we should never have a `body: "foo"` if `foo` is a path param. And query params are whatever's left over from what's not included in the path and body. So we just need to ensure no body params wind up as query params.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
vchudnov-g: OK, this ties back to the default values design I still have open.....

On empty strings, maybe  one of the [ideas from the web](https://www.google.com/search?q=empty+strings+as+query+params&oq=empty+strings+as+query+params&aqs=chrome..69i57j69i64l3.6119j1j7&sourceid=chrome&ie=UTF-8) will work with OP APIs. But we probably don't need to spend much time investigating until we decide on a the approach to default values. It certainly shouldn't block this PR.

But yes, let's add a TODO that likely something will need to be done here once we decide what we're doing for default values.

gapic/utils/case.py
vchudnov-g: But the generator has already produced libraries that were published, so if one of those had a hyphen somewhere, it made it through `to_snake_case` and changing that implementation could lead to an incompatible surface.

gapic/schema/wrappers.py
vchudnov-g: As discussed. That seems reasonable, but a complication is that repeated fields map to repeated query params. In that case, a sequence may be best. (If order doesn't matter, a multiset might be enough, but we can't assume that order doesn't matter for repeated fields)

gapic/schema/wrappers.py
```diff
@@ -767,6 +767,29 @@ def http_opt(self) -> Optional[Dict[str, str]]:
         # TODO(yon-mg): enums for http verbs?
         return answer
 
+    @property
+    def path_params(self) -> Sequence[str]:
+        """Return the path parameters found in the http annotation url"""
```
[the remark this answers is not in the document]
yon-mg: Well it doesn't do this yet so I'll update the comment when it does.

gapic/schema/wrappers.py
yon-mg: This is an oversight yes. Due to these nested fields I don't think it makes sense to identify all query params ahead of time, though it is possible! I am moving this logic out of the generator entirely and have left appropriate todos.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
yon-mg: So the short answer is yes but I will add tests to confirm that.

gapic/schema/wrappers.py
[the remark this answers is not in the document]
vchudnov-g: WDYM? You're returning the `{foo}` which is a field path, right?

gapic/schema/wrappers.py
vchudnov-g: good idea

gapic/schema/wrappers.py
vchudnov-g: OK. I see you added TODOs to come back to this later. sgtm.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
vchudnov-g: OK, great. If it's not trivial to URL encode, you can leave a TODO and do it in a follow-up PR, which is also fine. That would let us start to use/test what you have here for well formed data where URL encoding isn't necessary.

vchudnov-g: (BTW, do you know why the `&` shows up in red on GitHub?)

vchudnov-g (commented):
Looking good. Two of my previous `!!` comments are still open. The truthiness one we should mark with a TODO, since we can't fully resolve it right now. The URL-encoding one, if it's not trivial to address now, we should mark with a TODO and address in a follow-up PR.

gapic/schema/wrappers.py
yon-mg: You are right set is not appropriate even if order doesn't mater but I should mention though that once query param logic is moved out, this won't matter anymore.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
yon-mg: Not sure. Seems to be a bug w/ ampersand messing with syntax highlighting? Certainly no real error AFAIK.

gapic/schema/wrappers.py
vchudnov-g: True!

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
vchudnov-g: Thanks for the TODO!

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -168,13 +168,16 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
...
         #               not required for GCE
-        potentialParams = {
+        query_params = {
             {%- for field in method.query_params %}
             '{{ field|camel_case }}': request.{{ field }},
             {%- endfor %}
         }
-        potentialParams = ((k, v) for k, v in potentialParams.items() if v)
-        for i, (param_name, param_value) in enumerate(potentialParams):
+        # TODO(yon-mg): further discussion needed whether 'python truthiness' is appropriate here
```
vchudnov-g, line 176: Thanks for the TODO on truthiness!

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -168,13 +168,16 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
...
+        #               discards default values
+        # TODO(yon-mg): add test for proper url encoded strings
+        query_params = ((k, v) for k, v in query_params.items() if v)
+        for i, (param_name, param_value) in enumerate(query_params):
```
vchudnov-g, line 180: nit: Do consider the string.Join suggestion from earlier (but it's certainly not a blocker)

vchudnov-g (approved):
Looks good! Thanks for doing this!

For clarity when looking back at this commit, maybe you can expand in the merge commit description what the "basic case" is: path params in simple form, no body, no query param.

software-dov (approved):

gapic/schema/wrappers.py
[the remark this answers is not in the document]
software-dov: Excellent. Glad to hear that the issue has been resolved.

PR #708, opened 2020-12-11
build: use rules_python 0.1.0

busunkim96:
https://github.com/bazelbuild/rules_python/releases/tag/0.1.0


vam-google (approved):
LGTM assuming that when used from `googleapis` this still works (i.e. the `pip_install` works as excected).

software-dov (approved):

PR #716, opened 2020-12-21
fix: updating testing, rest-only generation, & minor bug-fixes

### LINKED ISSUE #267: Re-add HTTP transport and make it the fallback

yon-mg:
Updating unit tests on generated client rest transport. Most changes made to unit tests are with compute api in mind and are not yet comprehensive enough for showcase.

Fixed issues with rest-only transport generation. Again, changes made with compute api in mind especially since compute api doesn't have grpc endpoint.

Fixed bug with attribute keyword conflicts.

The goal with this PR is to be able to generate a minimum usable GCE client library. Since rest transport is still under work, it is not generated by default and must be set with the transport flag `--python_gapic_opt="transport=grpc+rest"`.


gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -81,7 +86,12 @@ def test_{{ service.client_name|snake_case }}_from_service_account_info():
         {% if service.host %}assert client.transport._host == '{{ service.host }}{% if ":" not in service.host %}:443{% endif %}'{% endif %}
 
 
-@pytest.mark.parametrize("client_class", [{{ service.client_name }}, {{ service.async_client_name }}])
+@pytest.mark.parametrize("client_class", [
+    {{ service.client_name }},
+    {%- if 'grpc' in opts.transport %}
```
vam-google, line 91: This construct is used everywhere in the template. Since it is so commonly used, please consider simplifying it by introducing some helper method, such that it look something like:
```
{%- if grpc %}
stuff
{%- endif %}

File: gapic/utils/checks.py [hunk 2]
```diff
@@ -0,0 +1,42 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re
+
+
+def is_str(expr: str) -> bool:
```
vam-google, line 18: If this file is used only for testing, please put it where the testing code is. Also please make sure that we don't do anything like this (guessting field type by its value, and using regexps for it), because it  is not reliable enough and is quite slow (the regexp part).

vam-google:
@software-dov PTAL. This is the last PR which blocks us from generating the first Pyghon Alpha client for GCE!

gapic/utils/checks.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2018 Google LLC
```
vam-google, line 1: Please use `field.type` instead `field.mock_value` in your test templates. By doing that you may avoid guessing work here, as we discussed over GVC. Also, please update copyright to 2020 here and in all your new files.

gapic/templates/noxfile.py.j2
```diff
@@ -6,7 +6,7 @@ import os
 import nox  # type: ignore
 
 
-@nox.session(python=['3.6', '3.7'])
+@nox.session(python=['3.6', '3.7', '3.8'])
```
busunkim96, line 9: Could you also add 3.9?
```suggestion
@nox.session(python=['3.6', '3.7', '3.8', '3.9'])
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -180,10 +186,8 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
         # TODO(yon-mg): further discussion needed whether 'python truthiness' is appropriate here
         #               discards default values
         # TODO(yon-mg): add test for proper url encoded strings
-        query_params = ((k, v) for k, v in query_params.items() if v)
-        for i, (param_name, param_value) in enumerate(query_params):
-            q = '?' if i == 0 else '&'
-            url += "{q}{name}={value}".format(q=q, name=param_name, value=param_value.replace(' ', '+'))
+        query_params = ['{k}={v}'.format(k=k, v=v) for k, v in query_params.items() if v]
+        url += '?{}'.format('&'.join(query_params)).replace(' ', '+')
```
busunkim96, line 190: It might be easier to use `urllib.parse.urlencode` here (It will create the `k=v` structure and glue the pairs together with `&` characters).

https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode
https://docs.python.org/3/library/urllib.request.html#urllib-examples

busunkim96 (commented):
Looks good to me (I am not an expert on the generator so deferring to other folks for approvals). I'm excited to try this out! 🚀

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
yon-mg: Not sure how I missed this. Thanks!

software-dov (approved):

gapic/templates/noxfile.py.j2
yon-mg: I thought the 3.9 requirement was optional for now since it's newish.

busunkim96: I've already added 3.9 to the noxfile template the cloud client libraries use. (https://github.com/googleapis/synthtool/blob/d5fc0bcf9ea9789c5b0e3154a9e3b29e5cea6116/synthtool/gcp/common.py#L221) It's my understanding that while there are a few packages missing wheels the generated unit tests should all be runnable with 3.9.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
yon-mg: To clarify, I'll leave this out for now so I don't accidentally break anything but I'll update in the subsequent PR

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -991,9 +1031,148 @@ def test_{{ method.name|snake_case }}_raw_page_lro():
...
+        response_value._content = json_return_value.encode('UTF-8')
+        req.return_value = response_value
+        {% if method.client_streaming %}
+        response = client.{{ method.name|snake_case }}(iter(requests))
+        {% else %}
+        response = client.{{ method.name|snake_case }}(request)
+        {% endif %}
+
+    {% if "next_page_token" in method.output.fields.values()|map(attribute='name') and not method.paged_result_field %}
+    {# Cheeser assertion to force code coverage for bad paginated methods #}
```
vchudnov-g, line 1077: What does "cheeser" mean in this context?

tests/unit/generator/test_generator.py
```diff
@@ -125,31 +125,49 @@ def test_get_response_ignores_unwanted_transports():
...
             )
-
-            assert len(cgr.file) == 4
+            assert len(cgr.file) == 5
             assert {i.name for i in cgr.file} == {
                 "foo/some_service/transports/river.py",
                 "foo/some_service/transports/car.py",
                 "foo/some_service/transports/__init__.py",
                 "foo/some_service/transports/base.py",
+                # Only generate async client with grpc transport
```
vchudnov-g, line 156: nit: This comment looks as thought it applies to the next line, but is a bit confusing. Suggest moving it below the next line and tweaking it to be something like: `# no async_client.py, as that's only for grpc transport`

vchudnov-g (commented):
Did a quick once-over. Looks good! Thanks for doing this.

Two very minor comments.

Issue #730, opened 2021-01-11
cannot parse error response for REST


arithmetic1728:
If response code is not 200, then `response.content` will be an error message. The following code cannot parse this error message, and an exception will be thrown.

https://github.com/googleapis/gapic-generator-python/blob/56c31de4a9f661e3d69b52e19c9a28dddfe9d7dc/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/transports/rest.py.j2#L202

Example `response.content`:
`
{\n"error": {\n    "code": 403,\n    "message": "Request is prohibited by organization\'s policy. vpcServiceControlsUniqueIdentifier: a69c658379e4178a",\n    "errors": [\n      {\n        "message": "Request is prohibited by organization\'s policy. vpcServiceControlsUniqueIdentifier: a69c658379e4178a",\n        "domain": "global",\n        "reason": "forbidden"\n      }\n    ],\n    "status": "PERMISSION_DENIED"\n  }\n}
`

PR #741, opened 2021-01-21
fix: fix rest transport unit test template

arithmetic1728:
Follow up fix of googleapis/gapic-generator-python#730

When regenerating python-compute, I noticed some unit test failure due to missing status code in the mock response. This PR adds status code to the mock response in the template.


busunkim96 (approved):
Are the generated unit tests for the REST transport not getting run in CI?

arithmetic1728:
> Are the generated unit tests for the REST transport not getting run in CI?

Looking at the testing log, it seems CI only generates and runs grpc unit tests. Filed an issue for this: https://github.com/googleapis/gapic-generator-python/issues/744

PR #750, opened 2021-01-25
fix: update paging implementation to handle unconventional pagination

yon-mg:
Pagination for GCE is unconventional in that it may use `max_results` instead of `page_results` and may return a `map` paginated resource instead of a `repeated` resource. Pagination otherwise does not change and even GCE may use conventional pagination for some rpcs.

Before this PR, pagination has only been implemented for the conventional case. This fix resolves that and adds the relevant testing to ensure coverage.

On an unrelated note, this PR now also includes a minor update to mypy cli flags due to the recent mypy update, https://mypy-lang.blogspot.com/2021/01/mypy-0800-released.html.

### DIFF
--- a/gapic/schema/wrappers.py
+++ b/gapic/schema/wrappers.py
@@ -866,16 +866,17 @@ class Method:
         """Return the response pagination field if the method is paginated."""
         # If the request field lacks any of the expected pagination fields,
         # then the method is not paginated.
-        for page_field in ((self.input, str, 'page_token'),
+        for page_field_token in ((self.input, str, 'page_token'),
                            (self.output, str, 'next_page_token')):
-            field = page_field[0].fields.get(page_field[2], None)
-            if not field or field.type != page_field[1]:
+            field = page_field_token[0].fields.get(page_field_token[2], None)
+            if not field or field.type != page_field_token[1]:
                 return None
-        page_fields = [self.input.fields.get('max_results', None),
-                       self.input.fields.get('page_size', None)]
-        page_field = next(
+
+        page_fields = (self.input.fields.get('max_results', None),
+                       self.input.fields.get('page_size', None))
+        page_field_size = next(
             (field for field in page_fields if field is not None), None)
-        if not page_field or page_field.type != int:
+        if not page_field_size or page_field_size.type != int:
             return None
 
         # Return the first repeated field.
--- a/gapic/schema/wrappers.py
+++ b/gapic/schema/wrappers.py
@@ -866,12 +866,17 @@ class Method:
         """Return the response pagination field if the method is paginated."""
         # If the request field lacks any of the expected pagination fields,
         # then the method is not paginated.
-        for page_field in ((self.input, int, 'page_size'),
-                           (self.input, str, 'page_token'),
+        for page_field in ((self.input, str, 'page_token'),
                            (self.output, str, 'next_page_token')):
             field = page_field[0].fields.get(page_field[2], None)
             if not field or field.type != page_field[1]:
                 return None
+        page_fields = [self.input.fields.get('max_results', None),
+                       self.input.fields.get('page_size', None)]
+        page_field = next(
+            (field for field in page_fields if field is not None), None)
+        if not page_field or page_field.type != int:
+            return None
 
         # Return the first repeated field.
         for field in self.output.fields.values():
--- a/gapic/schema/wrappers.py
+++ b/gapic/schema/wrappers.py
@@ -866,16 +866,19 @@ class Method:
         """Return the response pagination field if the method is paginated."""
         # If the request field lacks any of the expected pagination fields,
         # then the method is not paginated.
-        for page_field_token in ((self.input, str, 'page_token'),
+
+        # The request must have page_token and next_page_token as they keep track of pages
+        for source, source_type, name in ((self.input, str, 'page_token'),
                            (self.output, str, 'next_page_token')):
-            field = page_field_token[0].fields.get(page_field_token[2], None)
-            if not field or field.type != page_field_token[1]:
+            field = source.fields.get(name, None)
+            if not field or field.type != source_type:
                 return None
 
+        # The request must have max_results or page_size
         page_fields = (self.input.fields.get('max_results', None),
                        self.input.fields.get('page_size', None))
         page_field_size = next(
-            (field for field in page_fields if field is not None), None)
+            (field for field in page_fields if field), None)
         if not page_field_size or page_field_size.type != int:
             return None
 
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/pagers.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/pagers.py.j2
@@ -6,7 +6,7 @@
 {# This lives within the loop in order to ensure that this template
    is empty if there are no paged methods.
  -#}
-from typing import Any, AsyncIterable, Awaitable, Callable, Iterable, Sequence, Tuple
+from typing import Any, AsyncIterable, Awaitable, Callable, Iterable, Sequence, Tuple, Optional
 
 {% filter sort_lines -%}
 {% for method in service.methods.values() | selectattr('paged_result_field') -%}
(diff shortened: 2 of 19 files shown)


gapic/schema/wrappers.py
```diff
@@ -866,13 +866,19 @@ def paged_result_field(self) -> Optional[Field]:
...
         # then the method is not paginated.
-        for page_field in ((self.input, int, 'page_size'),
-                           (self.input, str, 'page_token'),
+        for page_field_token in ((self.input, str, 'page_token'),
                            (self.output, str, 'next_page_token')):
-            field = page_field[0].fields.get(page_field[2], None)
-            if not field or field.type != page_field[1]:
+            field = page_field_token[0].fields.get(page_field_token[2], None)
+            if not field or field.type != page_field_token[1]:
                 return None
```
software-dov, line 873: Great extension. Since we're changing things here, can we tweak the loop so that we're destructuring into names instead of indexing the tuple?

```python
for source, name in ((self.input, "page_token"), (self.output, "next_page_token")):
```
Something like that

gapic/schema/wrappers.py
```diff
@@ -866,13 +866,19 @@ def paged_result_field(self) -> Optional[Field]:
...
 
+        page_fields = (self.input.fields.get('max_results', None),
+                       self.input.fields.get('page_size', None))
+        page_field_size = next(
+            (field for field in page_fields if field is not None), None)
```
software-dov, line 878: Nit: we don't have to explicitly check for `None`.
```python
next((field for field in page_fields if field), None)
```

software-dov (commented):
Please add some comments in `paged_result_field` detailing the two options and how they work internally.

tests/unit/schema/wrappers/test_method.py
```diff
@@ -123,6 +123,16 @@ def test_method_paged_result_field_no_page_field():
...
 
+    method = make_method('Foo',
+                         input_message=make_message(name='FooRequest', fields=(
+                             make_field(name='page_token', type=9),  # str
+                         )),
+                         output_message=make_message(name='FooResponse', fields=(
+                             make_field(name='next_page_token', type=9),  # str
+                         ))
+                         )
+    assert method.paged_result_field is None
```
vchudnov-g, line 134: Do we add tests in this file for `max_results` and for mapped responses?

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -1162,7 +1162,122 @@ def test_{{ method.name|snake_case }}_rest_flattened_error():
...
+        assert all(isinstance(i, {{ method.paged_result_field.type.ident }})
+                   for i in results)
+        {% endif %}
+
+        pages = list(client.{{ method.name|snake_case }}(request={}).pages)
+        for page_, token in zip(pages, ['abc','def','ghi', '']):
+            assert page_.raw_page.next_page_token == token
+
+        
+{% endif %} {# paged methods #}
```
vchudnov-g, line 1279: Are the results of this template visible in a generated file within this repo? (just asking, the repo may not be structured that way)

vchudnov-g (commented):
Looks good! Minor comments and some questions for my own learning.

gapic/schema/wrappers.py
yon-mg: I just wanted to be extra explicit in case there was some false-ish value `field` might have. I suppose that's not the case for the `Field` class though.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -1162,7 +1162,122 @@ def test_{{ method.name|snake_case }}_rest_flattened_error():
...
+                    {{ method.paged_result_field.type.ident }}(),
+                ],
+            ),
+        )
+        {% endif %}
+        # Two responses for two calls
+        response = response + response
+
+        # Wrap the values into proper Response objs
+        response = tuple(map(lambda x: {{ method.output.ident }}.to_json(x), response))
```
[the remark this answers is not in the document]
software-dov: Nit: you don't need to call both `map` and `tuple`: `tuple` can take generator expressions as its iterable.
```python
response = tuple({{ method.output.ident }}.to_json(x) for x in response)
```

gapic/schema/wrappers.py
yon-mg: Good idea. Done

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -1162,7 +1162,122 @@ def test_{{ method.name|snake_case }}_rest_flattened_error():
...
+        side_effect = tuple(map(lambda x: Response(), response))
```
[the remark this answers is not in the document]
software-dov: We can be a little bit clever here:
```python
side_effect = (Response(),) * len(response)
```

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -1162,7 +1162,122 @@ def test_{{ method.name|snake_case }}_rest_flattened_error():
...
+        pager = client.{{ method.name|snake_case }}(request={})
+
+        assert pager._metadata == metadata
+
+        {% if method.paged_result_field.map %}
+        assert isinstance(pager.get('a'), {{ method.paged_result_field.type.fields.get('value').ident }})
+        assert pager.get('h') is None
+        {% endif %}
+
+        results = [i for i in pager]
```
software-dov, line 1260: Judgement call: can also do `results = list(pager)`

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -1162,7 +1162,122 @@ def test_{{ method.name|snake_case }}_rest_flattened_error():
...
+        assert len(results) == 6
+        {% if method.paged_result_field.map %}
+        assert all(
+            isinstance(i, tuple) and
+            tuple(map(lambda x: type(x), results[0])) == (str, {{ method.paged_result_field.type.fields.get('value').ident }})
+                   for i in results)
```
software-dov, line 1266: Am I misreading this expression, or is this a loop invariant? If so, let's lift it up. In either case, let's clarifiy it to
```python
tuple(type(r) for r in results[0])
```

Actually, the whole line looks like a loop invariant.

tests/unit/schema/wrappers/test_method.py
```diff
@@ -123,6 +123,16 @@ def test_method_paged_result_field_no_page_field():
                          )
     assert method.paged_result_field is None
 
+    method = make_method('Foo',
```
software-dov, line 126: Formatting nit: please add a newline and then indent here for better readability

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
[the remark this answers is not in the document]
yon-mg: I had a misunderstanding with `map()` I think software-dov addressed in a later comment

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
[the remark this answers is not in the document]
yon-mg: I think this does not work because new Response objects are not made. Instead it is aliases to the same object and editing a value within one changes it for all.

tests/unit/schema/wrappers/test_method.py
yon-mg: Mapped responses are treated the same here. Checking for repeated fields should be sufficient since mapped fields are also repeated. Test for `max_results` is now also added.

tests/unit/schema/wrappers/test_method.py
yon-mg: Done.. I think.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
yon-mg: This is cleaner. Done.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
yon-mg: No. It may be useful to include a generated client but that might clog the repo. The showcase client is generated as part of CircleCI tests though.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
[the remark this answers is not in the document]
yon-mg: Thanks for that. Not sure why I was dead set on using map.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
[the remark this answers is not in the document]
software-dov: Aha, good point.

gapic/templates/%namespace/%name_%version/%sub/services/%service/pagers.py.j2
```diff
@@ -68,14 +68,25 @@ class {{ method.name }}Pager:
             self._response = self._method(self._request, metadata=self._metadata)
             yield self._response
 
+    {% if method.paged_result_field.map %}
+    def __iter__(self) -> Iterable[Tuple[str, {{ method.paged_result_field.ident | replace('Sequence[', '') | replace(']', '') }}]]:
```
vam-google, line 72: Please fix the typing issue as we discussed offline.

gapic/templates/%namespace/%name_%version/%sub/services/%service/pagers.py.j2
```diff
@@ -68,14 +68,25 @@ class {{ method.name }}Pager:
...
+        for page in self.pages:
+            yield from page.{{ method.paged_result_field.name}}.items()
+
+    def get(self, key: str) -> {{ method.paged_result_field.ident | replace('Sequence', 'Optional') }}:
```
vam-google, line 76: Same here

yon-mg:
@software-dov friendly ping for review.

software-dov (approved):

PR #753, opened 2021-01-27
fix: Make gapic-generator-python compatible with protobuf 3.14.0 (native namespace package)

vam-google:
More details about Python namespace packaging here: https://packaging.python.org/guides/packaging-namespace-packages/#native-namespace-packages

The protobuf changes, which made this fix necessary are here:
https://github.com/protocolbuffers/protobuf/pull/7902
https://github.com/protocolbuffers/protobuf/pull/7908

The tracking bug for this issue (probably not the only one) https://github.com/googleapis/gapic-generator/issues/3334

This is only part of the fix, for the proper fix other google-namespaced python packages must be migrated to [pkgutil-style-namespace-packages](https://packaging.python.org/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages) and make sure they **do not** have `namespace_packages` in their `setup.py` file (an artifact from the legacy `pkg_resources-style`packages)

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -57,8 +57,8 @@ py_binary(
     python_version = "PY3",
     visibility = ["//visibility:public"],
     deps = [
-        "@com_google_protobuf//:protobuf_python",
         "@com_github_grpc_grpc//src/python/grpcio/grpc:grpcio",
+        requirement("protobuf"),
         requirement("click"),
         requirement("google-api-core"),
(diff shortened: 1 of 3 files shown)


noahdietz:
LGTM but we should let the python folks approve. Thanks Vadym.

busunkim96 (approved):
Thanks Vadym!

software-dov (approved):

PR #757, opened 2021-01-31
fix: Fix namespace packages conflict issue

vam-google:
This fixes the https://github.com/googleapis/gapic-generator/issues/3334 by excluding system-wide site-packages dir from python packages resolution path completely. 

It is done by simply setting `PYTHONNOUSERSITE` environment variable and then "re-running" python interpreter replacing the current process, but this time with the site packages excluded.

This pretty much implements the long-standing featrue request for rules_python https://github.com/bazelbuild/bazel/issues/4939, but only in scope of gapic-generator-python.


busunkim96 (approved):

noahdietz (approved):
Awesome! LGTM

PR #761, opened 2021-02-02
fix: don't use integer for enums in json encoding

yon-mg:
Integers for enums in json encoded requests is not desired.


busunkim96 (commented):
> Integers for enums in json encoded requests is not desired.

Out of curiosity is there a doc somewhere stating this? Is it for human readability?

busunkim96 (approved):

software-dov (approved):

vchudnov-g (approved):
Thanks for doing this!

If there is already a unit test that checks the REST request sent out, it would be good to capture this there. But if there isn't, don't worry about it right now; we'll do it later.

vchudnov-g:
@busunkim96 Yes, the internal design docs specify this. Essentially, we can't guarantee that the numeric values will be preserved when converting from the Discovery file, so we use the string representations.

PR #766, opened 2021-02-04
chore: Create main.yml

software-dov:


busunkim96:
Whenever the dust has cleared, please remember to edit https://github.com/googleapis/gapic-generator-python/blob/master/.github/sync-repo-settings.yaml so repo settings bot doesn't trample the required checks. :)

.github/sync-repo-settings.yaml
```diff
@@ -5,24 +5,6 @@ branchProtectionRules:
 - pattern: master
   isAdminEnforced: true
   requiredStatusCheckContexts:
-    - 'ci/circleci: docs'
```
alexander-fenster, line 8: should we add the actions CI here?

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,331 @@
+# WIP github actions
```
alexander-fenster, line 1: still WIP? :)

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,331 @@
...
+name: Tests
+
+# Controls when the action will run.
+on:
+  pull_request:
+    branches: [ master ]
```
alexander-fenster, line 7: If you use `$default-branch` here it won't break when we eventually switch to `main`

.github/sync-repo-settings.yaml
software-dov: I was under the impression that sync-repo-settings.yaml was generated based on settings saved in the UI and was not intended to be hand edited. @busunkim96 can you confirm or correct?

alexander-fenster (approved):
Wow, this is one of the most complicated CI setups I've seen in our org.
Thanks for doing it @software-dov!

.github/workflows/tests.yaml
software-dov: I mean...kind of? Publish package I've deferred deliberately, and publish_image is commented out as well. I'll remove the comment, though.

.github/workflows/tests.yaml
software-dov: Good catch.

.github/sync-repo-settings.yaml
busunkim96: @alexander-fenster is correct, the names of the new actions CI should be added here so sync-repo-settings bot preserves it as a required checks. 

The file is added to repos that have non-default CI configurations and overrides the default settings.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,330 @@
...
+            unzip protoc-3.12.1.zip
+            sudo ln -s /usr/src/protoc/bin/protoc /usr/local/bin/protoc
+      - name: Run showcase tests.
+        run: nox -s ${{ matrix.target }}
+  showcase-mtls:
+    strategy:
+      matrix:
+        target: [showcase_mtls, showcase_mtls_alternative_templates]
+    # needs:
+    #   - [docs, mypy, unit, showcase-unit, showcase-mypy]
```
busunkim96, line 112: Should this be deleted or is it a palceholder?

busunkim96 (approved):

.github/workflows/tests.yaml
software-dov: It's just cruft. Removing.

.github/sync-repo-settings.yaml
software-dov: Done

PR #774, opened 2021-02-10
feat: support self-signed JWT flow for service accounts

busunkim96:
See [RFC (internal only)](https://docs.google.com/document/d/1SNCVTmW6Rtr__u-_V7nsT9PhSzjj1z0P9fAD3YUgRoc/edit#) and https://aip.dev/auth/4111.

Support the self-signed JWT flow for service accounts by passing `default_scopes` and `default_host` in calls to the auth library and `create_channel`. This depends on features exposed in the following PRs: https://github.com/googleapis/python-api-core/pull/134, https://github.com/googleapis/google-auth-library-python/pull/665.

It may be easier to look at https://github.com/googleapis/python-translate/pull/107/files for a diff on a real library.

This change is written so that the library is (temporarily) compatible with older `google-api-core` and `google-auth` versions. Because of this it not possible to reach 100% coverage on a single unit test run. `pytest` runs twice in two of the `nox` sessions.

Miscellaneous changes:
- sprinkled in `__init__.py` files in subdirs of the `test/` directory, as otherwise pytest-cov seems to fail to collect coverage properly in some instances.
- new dependency on `packaging` for Version comparison https://pypi.org/project/packaging/

### DIFF
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
@@ -3,9 +3,11 @@
 {% block content %}
 import abc
 import typing
+import packaging.version
 import pkg_resources
 
 from google import auth  # type: ignore
+import google.api_core
 from google.api_core import exceptions  # type: ignore
 from google.api_core import gapic_v1    # type: ignore
 from google.api_core import retry as retries  # type: ignore
@@ -34,6 +36,17 @@ try:
 except pkg_resources.DistributionNotFound:
     DEFAULT_CLIENT_INFO = gapic_v1.client_info.ClientInfo()
 
+try:
+    # google.auth.__version__ was added in 1.26.0
+    _GOOGLE_AUTH_VERSION = auth.__version__
+except AttributeError:
+    try:  # try pkg_resources if it is available
+        _GOOGLE_AUTH_VERSION = pkg_resources.get_distribution("google-auth").version
+    except pkg_resources.DistributionNotFound:  # pragma: NO COVER
+        _GOOGLE_AUTH_VERSION = None
+
+_API_CORE_VERSION = google.api_core.__version__
+
 class {{ service.name }}Transport(abc.ABC):
     """Abstract transport class for {{ service.name }}."""
 
@@ -43,12 +56,14 @@ class {{ service.name }}Transport(abc.ABC):
         {%- endfor %}
     )
 
+    DEFAULT_HOST = {% if service.host %}'{{ service.host }}'{% else %}{{ None }}{% endif %}
+
     def __init__(
             self, *,
-            host: str{% if service.host %} = '{{ service.host }}'{% endif %},
+            host: str = DEFAULT_HOST,
             credentials: credentials.Credentials = None,
             credentials_file: typing.Optional[str] = None,
-            scopes: typing.Optional[typing.Sequence[str]] = AUTH_SCOPES,
+            scopes: typing.Optional[typing.Sequence[str]] = None,
             quota_project_id: typing.Optional[str] = None,
             client_info: gapic_v1.client_info.ClientInfo = DEFAULT_CLIENT_INFO,
             **kwargs,       
@@ -66,7 +81,7 @@ class {{ service.name }}Transport(abc.ABC):
             credentials_file (Optional[str]): A file with credentials that can
                 be loaded with :func:`google.auth.load_credentials_from_file`.
                 This argument is mutually exclusive with credentials.
-            scope (Optional[Sequence[str]]): A list of scopes.
+            scopes (Optional[Sequence[str]]): A list of scopes.
             quota_project_id (Optional[str]): An optional project to use for billing
                 and quota.
             client_info (google.api_core.gapic_v1.client_info.ClientInfo):	
@@ -83,6 +98,15 @@ class {{ service.name }}Transport(abc.ABC):
         # Save the scopes.
         self._scopes = scopes or self.AUTH_SCOPES
 
+        # TODO: Remove this if/else once google-auth >= 1.25.0 is required
+        if _GOOGLE_AUTH_VERSION and (
+            packaging.version.parse(_GOOGLE_AUTH_VERSION)
+            >= packaging.version.parse("1.25.0")
+        ):
+            scopes_kwargs = {"scopes": self._scopes, "default_scopes": self.AUTH_SCOPES}
+        else:
+            scopes_kwargs = {"scopes": self._scopes}
+
         # If no credentials are provided, then determine the appropriate
         # defaults.
         if credentials and credentials_file:
@@ -91,12 +115,12 @@ class {{ service.name }}Transport(abc.ABC):
         if credentials_file is not None:
             credentials, _ = auth.load_credentials_from_file(
                                 credentials_file,
-                                scopes=self._scopes,
+                                **scopes_kwargs,
                                 quota_project_id=quota_project_id
                             )
 
         elif credentials is None:
-            credentials, _ = auth.default(scopes=self._scopes, quota_project_id=quota_project_id)
+            credentials, _ = auth.default(**scopes_kwargs, quota_project_id=quota_project_id)
 
         # Save the credentials.
         self._credentials = credentials
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
@@ -67,7 +67,7 @@ class {{ service.name }}Transport(abc.ABC):
             scopes: Optional[Sequence[str]] = None,
             quota_project_id: Optional[str] = None,
             client_info: gapic_v1.client_info.ClientInfo = DEFAULT_CLIENT_INFO,
-            **kwargs,       
+            **kwargs,
             ) -> None:
         """Instantiate the transport.
 
@@ -85,10 +85,10 @@ class {{ service.name }}Transport(abc.ABC):
             scopes (Optional[Sequence[str]]): A list of scopes.
             quota_project_id (Optional[str]): An optional project to use for billing
                 and quota.
-            client_info (google.api_core.gapic_v1.client_info.ClientInfo):	
-                The client info used to send a user-agent string along with	
-                API requests. If ``None``, then default info will be used.	
-                Generally, you only need to set this if you're developing	
+            client_info (google.api_core.gapic_v1.client_info.ClientInfo):
+                The client info used to send a user-agent string along with
+                API requests. If ``None``, then default info will be used.
+                Generally, you only need to set this if you're developing
                 your own client library.
         """
         # Save the hostname. Default to port 443 (HTTPS) if none is specified.
@@ -99,13 +99,13 @@ class {{ service.name }}Transport(abc.ABC):
         scopes_kwargs = self._get_scopes_kwargs(self._host, scopes)
 
         # Save the scopes.
-        self._scopes = scopes_kwargs["scopes"]
+        self._scopes = scopes or self.AUTH_SCOPES
 
         # If no credentials are provided, then determine the appropriate
         # defaults.
         if credentials and credentials_file:
             raise exceptions.DuplicateCredentialArgs("'credentials_file' and 'credentials' are mutually exclusive")
-        
+
         if credentials_file is not None:
             credentials, _ = auth.load_credentials_from_file(
                                 credentials_file,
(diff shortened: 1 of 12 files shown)


gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
```diff
@@ -80,6 +95,21 @@ class {{ service.name }}Transport(abc.ABC):
             host += ':443'
         self._host = host
 
+        # If a custom API endpoint is set, set scopes to ensure the auth
+        # library does not used the self-signed JWT flow for service
+        # accounts
+        if host != self.DEFAULT_HOST and not scopes:
+            scopes = self.AUTH_SCOPES
```
busunkim96, line 102: If port is provided (say `pubsub.googleapis.com:443` instead of `pubsub.googleapis.com`) the library will think the OAuth2 flow is necessary.

I initially tried splitting on `:` but that broke the showcase tests where default host _is_ localhost.  Is there a better way to parse the provided host reliably?

software-dov: Silly questions: which is the OAuth2 flow, and why do we not want it all the time? Is it the "old style"? What are we trying to parse out? We can always fall back to regexes, something like

host_re = re.compile("(?P<host>[^:]+)(:(?P<port>\d+))?")

Or is that not the part of parsing out the host that's problematic?

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -81,12 +81,14 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
...
         # TODO(yon-mg): resolve other ctor params i.e. scopes, quota, etc.
+        # TODO: When custom host (api_endpoint) is set, `scopes` must *also* be set on the
+        # credentials object 
         super().__init__(
             host=host,
             credentials=credentials,
             client_info=client_info,
         )
-        self._session = AuthorizedSession(self._credentials)
+        self._session = AuthorizedSession(self._credentials, default_host=cls.DEFAULT_HOST)
```
software-dov, line 91: 'cls' isn't a visible name here. This should be self.DEFAULT_HOST

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -109,7 +111,9 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
                 grpc_helpers.create_channel(
                     self._host,
                     credentials=self._credentials,
-                    scopes=self.AUTH_SCOPES,
+                    default_scopes=self.AUTH_SCOPES,
+                    scopes=scopes,
+                    default_host=cls.DEFAULT_HOST,
```
software-dov, line 116: 'cls' isn't a visible name here. This should be self.DEFAULT_HOST

software-dov (commented):
Wow, this is quite the endeavor. Thanks for doing it!

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
busunkim96: Ah sorry for being vague.

The OAuth2 flow (currently always used for service accounts) requires a call out to `oauth2.googleapis.com`. With a service account, it's sometimes possible to "self-sign" the token, which avoids a network round trip.

The self-signed JWT option can be used when:

1. A service account is being used
2. A scope has not been set by the user
3. A target_audience has not been set by the user (irrelevant as we don't have an client option for this)
4. An api_endpoint has not been set by the user

Number 4 is what's relevant here. If the user is using the non-default endpoint we want to make sure the auth library uses OAuth2. The suggested way to do this in the [doc](https://docs.google.com/document/d/1SNCVTmW6Rtr__u-_V7nsT9PhSzjj1z0P9fAD3YUgRoc/edit#heading=h.r0s8lehl5drr) is to pass user scopes to the auth library.

I think that regex will work! I'll experiment and push a new commit.

arithmetic1728 (approved):

arithmetic1728:
FYI, cl/357773081 is submitted, so I guess custom endpoint will no longer be an issue for self signed JWT going forward.

busunkim96:
@bshaffer Would you be able to take a look?

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
```diff
@@ -2,10 +2,13 @@
 
 {% block content %}
 import abc
+import re
```
software-dov, line 5: Unused import

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc_asyncio.py.j2
```diff
@@ -75,13 +77,36 @@ class {{ service.grpc_asyncio_transport_name }}({{ service.name }}Transport):
...
+        # accounts
+        if (
+            host is not None
+            and cls.DEFAULT_HOST is not None
+            and host.split(":")[0] != cls.DEFAULT_HOST.split(":")[0]
+            and not scopes
+        ):
+            scopes = cls.AUTH_SCOPES
+
+        self_signed_jwt_kwargs = {}  # type: Dict[str, Union[Optional[Sequence[str]], str]]
```
software-dov, line 91: Nit: prefer the inline type syntax over comment based.

software-dov (approved):

bshaffer:
> FYI, cl/357773081 is submitted, so I guess custom endpoint will no longer be an issue for self signed JWT going forward. @bshaffer

This is true but it still will be some time before it rolls out. It's up to the owners of this repo if that's acceptable to wait for the rollout or go ahead with the fix we have here. I'd prefer the latter.

arithmetic1728:
> > FYI, cl/357773081 is submitted, so I guess custom endpoint will no longer be an issue for self signed JWT going forward. @bshaffer
> 
> This is true but it still will be some time before it rolls out. It's up to the owners of this repo if that's acceptable to wait for the rollout or go ahead with the fix we have here. I'd prefer the latter.

Yep, agreed, it is just a FYI. It still takes one or two weeks for the roll out.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
```diff
@@ -138,7 +162,7 @@ class {{ service.name }}GrpcTransport({{ service.name }}Transport):
                 credentials=credentials,
                 credentials_file=credentials_file,
                 ssl_credentials=ssl_credentials,
-                scopes=scopes or self.AUTH_SCOPES,
+                scopes=scopes,
```
bshaffer, line 165: I am assuming this is the same as "pass scopes when self-signed JWT isn't used".

However, I am confused why grpc needs to know scopes when the access token is being handled by the auth library.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
```diff
@@ -80,6 +97,26 @@ class {{ service.name }}Transport(abc.ABC):
             host += ':443'
         self._host = host
 
+        # If a custom API endpoint is set, set scopes to ensure the auth
+        # library does not used the self-signed JWT flow for service
+        # accounts
+        if (
+            host is not None
+            and self.DEFAULT_HOST is not None
```
bshaffer, line 105: if `self.DEFAULT_HOST` is `None`, then shouldn't default scopes be used? Without a default host I do not think the library can self-sign a JWT.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
```diff
@@ -101,6 +104,27 @@ class {{ service.name }}GrpcTransport({{ service.name }}Transport):
...
         """
+    
+        # If a custom API endpoint is set, set scopes to ensure the auth
+        # library does not used the self-signed JWT flow for service
+        # accounts
+        if (
+            host is not None
+            and self.DEFAULT_HOST is not None
+            and host.split(":")[0] != self.DEFAULT_HOST.split(":")[0]
+            and not scopes
```
bshaffer, line 115: Why does this logic need to be here and in `base.py`? I also see it two more times below for `grpc_async`. Is that entirely necessary? Can we make this DRYer? Four times for the same logic makes me nervous.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
busunkim96: `grpc_helpers.create_channel` is supposed to somewhat centralize the creation of the `google.auth `credential object (if it is `None`). It looks like this transport *also* initializes a credential if it's `None`, but I think it is redundant and am working on getting it removed in a separate PR.

https://github.com/googleapis/python-api-core/blob/b4860fec797f7d9f02a7241b2b6cef2d27f96f28/google/api_core/grpc_helpers.py#L195

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
busunkim96: That is a fair point. I opened up https://github.com/googleapis/gapic-generator-python/pull/808/ to try to remove some of the existing duplication, which should make it possible to shrink this PR as well.

software-dov (approved):

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
busunkim96: I thought about this a bit more, and I think this case (no default host) is taken care of in the auth library: https://github.com/googleapis/google-auth-library-python/blob/d80c85f285ae1a44ddc5a5d94a66e065a79f6d19/google/auth/transport/grpc.py#L82-L85.

The `if` statement got a bit convoluted because I was trying to avoid creating multiple branches (which would drop coverage). I changed `self.DEFAULT_HOST` to be a `str`, so I don't have to check `is not None` before parsing the string.

busunkim96:
@bshaffer PTAL. :)

software-dov:
Does @bshaffer need to take a look, or is this good to merge?

busunkim96:
@bshaffer or @arithmetic1728 would you mind taking another look?

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
```diff
@@ -43,15 +57,17 @@ class {{ service.name }}Transport(abc.ABC):
         {%- endfor %}
     )
 
+    DEFAULT_HOST: str = {% if service.host %}'{{ service.host }}'{% else %}{{ '' }}{% endif %}
```
busunkim96, line 60: @bshaffer Are there any concerns with making DEFAULT_HOST a public attribute?

busunkim96 (commented):
@bshaffer I removed the scope override with custom endpoints to match guidance in https://docs.google.com/document/d/1FoJPYQu8ZHdGBFY9TBRZCDrWJJnxw3fAz5CG5mdZR9E/edit. Could you take a look?

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
bshaffer: I don't believe there is any issue there!

busunkim96:
@software-dov Could you take one more look? Thank you for your help getting this over the line. :)

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
```diff
@@ -99,13 +99,13 @@ class {{ service.name }}Transport(abc.ABC):
         scopes_kwargs = self._get_scopes_kwargs(self._host, scopes)
 
         # Save the scopes.
-        self._scopes = scopes_kwargs["scopes"]
+        self._scopes = scopes or self.AUTH_SCOPES
```
bshaffer, line 102: If I'm understanding this correctly, `scopes_kwargs["scopes"]` was essentially `scopes` but with a check for if the `endpoint` was supplied.  And I am interpreting `scopes or self.AUTH_SCOPES` to mean "user supplied scopes _or_ the default scopes"

Is that accurate?

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
```diff
@@ -120,19 +120,7 @@ class {{ service.name }}Transport(abc.ABC):
         self._credentials = credentials
 
 
-    @classmethod
-    def _get_user_scopes(cls, host: str, scopes: Optional[Sequence[str]]) -> Optional[Sequence[str]]:
```
bshaffer: 👍

bshaffer (commented):
One question I still have is, unfortunately, for the case of the regional endpoint `us-documentai.googleapis.com`, we need a check to make sure the `DEFAULT_HOST` doesn't contain a regional endpoint. 

If this is something we should do here, or somewhere else, I am not sure, I just want to verify it's being considered!

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
busunkim96: Yep! Since the `endpoint` restriction has been lifted I went back to "user supplied scopes or the default scopes"

busunkim96:
> One question I still have is, unfortunately, for the case of the regional endpoint us-documentai.googleapis.com, we need a check to make sure the DEFAULT_HOST doesn't contain a regional endpoint.

@bshaffer From reading the doc it seems like documentai is the exception, so I am planning on handling that with a synthtool/owlbot replace. If we anticipate more APIs using the regional endpoints as the default host I can build logic into the generator.

bshaffer:
> build logic into the generator.

That sounds great!

bshaffer (approved):

software-dov (approved):

PR #781, opened 2021-02-17
feat: add gapic metadata file

software-dov:
The GAPIC metadata file is used to track code, samples, and test
coverage for every RPC and library method.


software-dov:
Note: hand testing of the generated file indicates that it is valid JSON and behaves in a reasonable manner.

busunkim96 (approved):

gapic/ads-templates/%namespace/%name/%version/gapic_metadata.json.j2
```diff
@@ -0,0 +1,33 @@
+{
+    "schema": "1.0",
```
vchudnov-g, line 2: If we continue with this template approach, at least specify in both templates that we're adhering to the .proto schema in http://github.com/....

vchudnov-g (commented):
Looks good, but I just have to ask: is it not practical to populate a metadata proto object with the data and then serialize it to JSON? That way we're more tightly tied to the source of truth (the .proto) and we don't have to worry about presentation logic ourselves. I want to avoid shutting the .proto out of the loop, as it's conceivable we may want to add items later. (That said, if there's a practical reason for emitting straight JSON, that's fine, but let's mention it in the PR.)

software-dov:
There is no deep reason for emitting straight JSON.
This is a reading comprehension failure on my part; I simply missed the bottom part of the design doc that described where the corresponding proto file lived and that we should use it. I'll amend the PR immediately.

software-dov:
**NOTE:** in light of a design doc reread, this PR is on hold until https://pypi.org/project/googleapis-common-protos/ is updated to include the python binding for `metadata.proto`.

vchudnov-g:
Any updates or blockers on the the googleapis-common-protos issue?

busunkim96:
@software-dov Opened https://github.com/googleapis/python-api-common-protos/pull/38. If you'd like to try a pip VCS install `git+https://github.com/googleapis/python-api-common-protos.git@add-gapic-metadata` should work

software-dov:
Approved the change to api-common-protos and just about to experiment with the it in local code.

software-dov:
The updated api-common-protos seems to be fine.

tests/unit/schema/test_api.py
```diff
@@ -1287,3 +1289,224 @@ def test_map_field_name_disambiguation():
...
+                    )
+                }
+            ),
+            "Squid": gapic_metadata_pb2.GapicMetadata.ServiceForTransport(
+                clients={
+                    "grpc": gapic_metadata_pb2.GapicMetadata.ServiceAsClient(
+                        library_client="SquidClient",
+                        rpcs={
+                            "Giant": gapic_metadata_pb2.GapicMetadata.MethodList(methods=["giant"]),
+                            "Humboldt": gapic_metadata_pb2.GapicMetadata.MethodList(methods=["humboldt"]),
```
vchudnov-g, line 1482: Can you ever have more than one Python method for a given RPC? If so, it might be good to test that, as well.

vchudnov-g (commented):
Looks good. One minor suggestion.
(Apologies for not getting to this earlier.)

gapic/schema/api.py
```diff
@@ -392,6 +394,45 @@ def subpackages(self) -> Mapping[str, 'API']:
...
+        for service in sorted(self.services.values(), key=lambda s: s.name):
+            service_desc = gm.services.get_or_create(service.name)
+
+            # At least one of "grpc" or "rest" is guaranteed to be present because
+            # of the way that Options instances are created.
+            # This assumes the options are generated by the class method factory.
+            transports = []
+            if "grpc" in options.transport:
+                transports.append(("grpc", service.client_name))
+                transports.append(("grpcAsync", service.async_client_name))
```
vchudnov-g, line 418: I would suggest using `grpc-async` for the name.

tests/unit/schema/test_api.py
software-dov: Per library client, no. As addressed elsewhere, asynchronous variants belong to a separate client class.

PR #795, opened 2021-03-02
feat: add flag for gapic metadata

software-dov:
As per the design doc, gapic metadata should not be generated by
default, and should be gated by the 'metadata' flag.

When invoking the generator via protoc, toggle the metadata flag like so:
--python_gapic_opt="metadata"

Subsequent Bazel integration is WIP.

### DIFF
--- a/gapic/utils/options.py
+++ b/gapic/utils/options.py
@@ -50,7 +50,7 @@ class Options:
     OPT_FLAGS: FrozenSet[str] = frozenset((
         'add-iam-methods',      # microgenerator implementation for `reroute_to_grpc_interface`
         'lazy-import',          # requires >= 3.7
-        'metadata',             # generate gapic metadata JSON file
+        'metadata',             # generate GAPIC metadata JSON file
         'old-naming',           # TODO(dovs): Come up with a better comment
         'retry-config',         # takes a path
         'samples',              # output dir
--- a/gapic/utils/options.py
+++ b/gapic/utils/options.py
@@ -41,20 +41,22 @@ class Options:
     lazy_import: bool = False
     old_naming: bool = False
     add_iam_methods: bool = False
+    metadata: bool = False
     # TODO(yon-mg): should there be an enum for transport type?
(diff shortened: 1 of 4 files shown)


gapic/utils/options.py
```diff
@@ -41,20 +41,22 @@ class Options:
...
+    metadata: bool = False
     # TODO(yon-mg): should there be an enum for transport type?
     transport: List[str] = dataclasses.field(default_factory=lambda: [])
 
     # Class constants
     PYTHON_GAPIC_PREFIX: str = 'python-gapic-'
     OPT_FLAGS: FrozenSet[str] = frozenset((
+        'add-iam-methods',      # microgenerator implementation for `reroute_to_grpc_interface`
+        'lazy-import',          # requires >= 3.7
+        'metadata',             # generate gapic metadata JSON file
```
vchudnov-g, line 53: totally insignificant nit: in mixed-case comments, we can capitalize GAPIC, since it's an acronym

tests/unit/generator/test_generator.py
```diff
@@ -116,6 +116,22 @@ def test_get_response_fails_invalid_file_paths():
...
+        lt.return_value = ["gapic/gapic_metadata.json.j2"]
+        with mock.patch.object(jinja2.Environment, "get_template") as gt:
+            gt.return_value = jinja2.Template(
+                "This is not something we want to see")
+            res = g.get_response(
+                api_schema=make_api(),
+                opts=Options.build(""),
+            )
+
+            # We don't expect any files.
```
vchudnov-g, line 131: nit: maybe be explicit with "because opts.metadata is not set"

vchudnov-g (approved):
Looks good. Thanks for doing this!

Before turning on the metadata file, we should resolve the issue in the other pr about the dual transports.

gapic/utils/options.py
software-dov: Fixed

PR #806, opened 2021-03-05
chore: __init__.py file output is now more deterministic

software-dov:
Sort everything that is iteratively produced, e.g. message and enum
classes provided.


busunkim96 (approved):
Ah I was wondering why these still seemed to move around sometimes. Thank you!

software-dov:
I had hoped it would get better once the generation process started using 3.7, where dicts are ordered and the air smells like warm root beer. In any case, being explicit is better.

PR #808, opened 2021-03-09
refactor: remove duplication in grpc transport init

busunkim96:
Remove duplication in the `grpc` and `grpc_asyncio` transport `__init__`s.

Changes:
- rely on the base transport to set the host (adding `:443` if it doesn't exist), credential, and scopes (user scopes vs. `self.AUTH_SCOPES`).
- only have one call to `create_channel`

No behavior changes are intended.

### DIFF
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
@@ -98,8 +98,8 @@ class {{ service.name }}Transport(abc.ABC):
         # Save the credentials.
         self._credentials = credentials
 
-        # Lifted into its own function so it can be stubbed out during tests.
-        self._prep_wrapped_messages(client_info)
+        # Save the scopes.
+        self._scopes = scopes or self.AUTH_SCOPES
 
 
     def _prep_wrapped_messages(self, client_info):
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
@@ -80,6 +80,9 @@ class {{ service.name }}Transport(abc.ABC):
             host += ':443'
         self._host = host
 
+        # Save the scopes.
+        self._scopes = scopes or self.AUTH_SCOPES
+
         # If no credentials are provided, then determine the appropriate
         # defaults.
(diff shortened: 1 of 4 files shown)


gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
```diff
@@ -88,19 +91,16 @@ class {{ service.name }}Transport(abc.ABC):
...
         elif credentials is None:
-            credentials, _ = auth.default(scopes=scopes, quota_project_id=quota_project_id)
+            credentials, _ = auth.default(scopes=self._scopes, quota_project_id=quota_project_id)
 
         # Save the credentials.
         self._credentials = credentials
 
-        # Lifted into its own function so it can be stubbed out during tests.
-        self._prep_wrapped_messages(client_info)
-
```
arithmetic1728, line 103: why moved `self._prep_wrapped_messages(client_info)` to both grpc and grpc_asyncio?

busunkim96: [`self._prep_wrapped_messages`](https://github.com/googleapis/python-automl/blob/bab596b3b0fc04581602dd5817b08f876db846fe/google/cloud/automl_v1beta1/services/prediction_service/transports/base.py#L106-L115) creates a dict of the the method callables. The [`grpc` method callables](https://github.com/googleapis/python-automl/blob/bab596b3b0fc04581602dd5817b08f876db846fe/google/cloud/automl_v1beta1/services/prediction_service/transports/grpc.py#L300-L306) need `self.grpc_channel` to exist.

With the change in this PR the base transport needs to set `self._host`, `self._scopes`, and `self._credential` _before_ the gRPC channel can be created. As a result I moved the `self._prep_wrapepd_messages` call to after channel creation in `grpc` and `grpc_asyncio`. 

Suggestions on cleaner ways to do this much appreciated. :)

arithmetic1728 (approved):

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
arithmetic1728: I see. I guess maybe we can add a comment here in grpc and grpc_asyncio to remind us to keep the order. Thank you for simplifying the logic :)

software-dov (approved):

PR #809, opened 2021-03-10
feat: update templates to permit enum aliases

software-dov:
Certain APIs like RecommendationEngine use multiple enum variant
monikers to reference the same value. Achieving this in protos
requires explicit support from proto-plus and the generated surface.

Hand testing indicates compliance.


busunkim96 (approved):
Does the minimum `proto-plus` version in `setup.py` need to be bumped?

software-dov:
Yes it does, good catch.

PR #811, opened 2021-03-11
feat: add bazel support for gapic metadata

software-dov:


busunkim96 (approved):
Meta-comment: Should we create a docs page with all the bazel options?

vchudnov-g (approved):

software-dov:
> Meta-comment: Should we create a docs page with all the bazel options?

Probably. It's definitely getting to the point of tribal knowledge/code archaeology.

PR #828, opened 2021-03-29
chore: deprecate return_immediately flag in Pub/Sub pull()

### LINKED ISSUE #827: Deprecate return_immediately flag in subscriber client's pull() method
Original issue: https://github.com/googleapis/python-pubsub/issues/350

It will be cleaner to make this change in the generator itself as opposed to fiddling with `synth.py`.

### LINKED ISSUE #16454: Deprecate return_immediately flag in subscriber client's pull() method
Original issue: https://github.com/googleapis/python-pubsub/issues/350

It will be cleaner to make this change in the generator itself as opposed to fiddling with `synth.py`.

plamut:
Closes googleapis/google-cloud-python#16454.

As per ticket description and internal discussion, the said flag should emit a `DeprecationWarning` if set to `True`.

~~Might run out of time today to update the tests, though.~~ Done.

### DIFF
--- a/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
@@ -6,6 +6,7 @@ from distutils import util
 import os
 import re
 from typing import Callable, Dict, Optional, {% if service.any_server_streaming %}Iterable, {% endif %}{% if service.any_client_streaming %}Iterator, {% endif %}Sequence, Tuple, Type, Union
+import warnings
 import pkg_resources
 
 from google.api_core import client_options as client_options_lib  # type: ignore
@@ -380,6 +381,14 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
             {%- endif %} {# field.map #}
             {%- endfor %} {# key, field in method.flattened_fields.items() #}
             {%- endif %} {# method.client_streaming #}
+
+        {%- if method.name|snake_case == "pull" %}
+        if request.return_immediately:
+            warnings.warn(
+                "The return_immediately flag is deprecated and should be set to False.",
+                category=DeprecationWarning,
+            )
+        {%- endif %}
 
         # Wrap the RPC method; this adds retry and timeout information,
         # and friendly error handling.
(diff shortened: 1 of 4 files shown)


gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
```diff
@@ -381,6 +382,14 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
             {%- endfor %} {# key, field in method.flattened_fields.items() #}
             {%- endif %} {# method.client_streaming #}
 
+        {%- if method.name|snake_case == "pull" %}
+        if request.return_immediately:
+            warnings.warn(
+                "The return_immediately flag is deprecated and should be set to False.",
+                category=DeprecationWarning,
+            )
+        {%- endif %}
```
software-dov, line 391: I am in general not comfortable adding this change. It's extremely specific to one method of one API, the check is too generic for safety, and it complicates the template for a warning. People mostly ignore warnings.

busunkim96: +1 to this seeming too generic to keep here. If some other API happens to have a field with the same name it will result in a spurious warning.

It looks like this was field was marked deprecated in the protos last year.

https://github.com/googleapis/googleapis/blob/a1ab4d44db02d59ff58810c6d4182d84e4b9abaa/google/pubsub/v1/pubsub.proto#L994-L1003
https://github.com/googleapis/googleapis/commit/01c681586d8d6dbd60155289b587aee678530bd9

I think ideally the generators would notice fields marked 'deprecated' and automatically update the docstring and raise a warning.

plamut: What would be a reasonable alternative in your opinion? We can still hack it in through `synth.py`, but the regex will likely be quite messy.

There's also an ongoing initiative (AFAIK?) to move most of the code from `synth.py` into the generator itself (please correct me if I'm wrong on that).

software-dov: I think this is primarily a documentation problem. That documentation can take multiple concurrent forms: docstrings, samples with comments, and hand written documention. I am tentatively in favor of the generator checking for the `deprecated` flag and autogenerating some docstring boilerplate and mostly **not** in favor of printing warnings.

software-dov: I think that there are likely to always be situations where the most reasonable and practical solution is `synth.py` hacks. Removing most of the code from `synth.py` is definitely an explicit goal, but moving oneoff fixes into the generator are half solutions.

plamut: Re: deprecation warning - here it's an explicit [request](https://github.com/googleapis/python-pubsub/issues/350), was decided in one of the meetings. I was not present, though, thus I'll re-check if it is a hard requirement.

plamut:
As discussed offline, it's fine to have such specific changes in `synth.py` instead of in the generator.

PR #855, opened 2021-04-30
perf: reduce unnecessary copies, optimize Address comparison

### LINKED ISSUE #805: Ordering is not consistent for generated init files
We are receiving synthtool PR changes, but it's just a reordering of the classes. They seem to keep flipping back and forth, as I've merged in the opposite ordering previously: https://github.com/googleapis/python-bigtable/pull/237/files

Opposite example: https://github.com/googleapis/python-bigtable/pull/235/files

### LINKED ISSUE #852: Version 0.44.0 adds extra whitespace in Ads templates
Sort of an odd one, but starting in 0.44.0 we started seeing a single extra space in the generated version of [this file](https://github.com/googleapis/gapic-generator-python/blob/master/gapic/ads-templates/%25namespace/%25name/%25version/__init__.py.j2#L5)

### LINKED ISSUE #853: 0.44.1 causes OOM errors for ads v7
See cl/370989216

> This version of gapic-generator-python appears to have a memory leak and doesn't work with the googleads v7 protos. This causes the python microgenerator job to run for ~1 hours and then die with an [OOM error](https://fusion2.corp.google.com/invocations/492dcc4b-a7fb-4288-b2d1-8dae8ac581be/targets/actools%2Fgoogleapis%2Finternal%2Fpresubmit/log). Can also reproduce the unbounded runtime / memory growth locally on glinux. We need to roll this back to 44.0 so the ads libraries can build.

software-dov:
Includes various other performance optimizations and minor whitespace fixes.

Local testing cuts down Ads generation time to ~1m (and prevents OOM) and DialogflowCX from ~10 minutes to ~5 minutes.

### DIFF
--- a/gapic/ads-templates/%namespace/%name/%version/%sub/__init__.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/%sub/__init__.py.j2
@@ -8,21 +8,23 @@
     them again.
 -#}
 __all__ = (
-    {% for subpackage in api.subpackages|dictsort %}
+    {% filter sort_lines -%}
+    {% for subpackage in api.subpackages -%}
     '{{ subpackage }}',
-    {% endfor %}
-    {% for service in api.services.values()|sort(attribute='client_name')
-            if service.meta.address.subpackage == api.subpackage_view %}
+    {% endfor -%}
+    {% for service in api.services.values()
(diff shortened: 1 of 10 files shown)


busunkim96 (approved):
Thank you!

PR #861, opened 2021-05-04
feat: add async samples

busunkim96:


software-dov (approved):

gapic/samplegen/samplegen.py
```diff
@@ -944,17 +953,19 @@ def generate_sample_specs(api_schema: api.API, *, opts) -> Generator[Dict[str, A
...
+            else:
+                transport_type = "sync"
             for rpc_name, method_list in client.rpcs.items():
                 # Region Tag Format:
                 # [{START|END} ${apishortname}_generated_${api}_${apiVersion}_${serviceName}_${rpcName}_{sync|async}_${overloadDisambiguation}]
                 region_tag = f"{api_short_name}_generated_{api_schema.naming.versioned_module_name}_{service_name}_{rpc_name}_{transport_type}"
                 spec = {
                     "sample_type": "standalone",
                     "rpc": rpc_name,
+                    "transport": transport_name,
```
tseaver, line 968: Should this be `transport_type`?  I think `transport_name` would raise `NameError` here.

busunkim96: `transport_name` is created on 956 and is one of `grpc`, `grpc-async` or `rest`

tseaver (approved):

software-dov (commented):
Looks good, just a few optional nits and what looks like a copy/paste error.
One other thing: I'm not sure I like transport types being magic strings. If I were making code additions in this area, I would _definitely_ mess up at some point and use "grpc_async" or "grpcAsync" and not catch it immediately. How annoying would it be to use named constants instead?

busunkim96:
@software-dov Thank you for the review! Will clean this up and re-request a review.

> One other thing: I'm not sure I like transport types being magic strings. If I were making code additions in this area, I would definitely mess up at some point and use "grpc_async" or "grpcAsync" and not catch it immediately. How annoying would it be to use named constants instead?

Named constants sound like they will be less error prone :)

busunkim96:
@software-dov PTAL.

I'm not sure why codecov isn't running. Is there a way to trigger it manually? Should I try pushing another commit?

software-dov:
I don't know why codecov isn't running. The tubes must be clogged.
I think the best way to try to solve this is to just rerun all the CI. That's easy to do from the dashboard.

software-dov (approved):

PR #869, opened 2021-05-07
fix: also add the async client to __all__

### LINKED ISSUE #815: foo_v1/__init__.py missing Async Client
See https://github.com/googleapis/python-kms/blob/a6358c1dcdd57064c77e8cafb3f302033688dba5/google/cloud/kms_v1/__init__.py#L18 for an example

busunkim96:
Follow up to #859

Closes #815 

Also delete a stray } in feature_fragments which was resulting in almost empty files getting generated in clients. https://github.com/googleapis/googleapis-gen/blob/master/google/appengine/v1/google-cloud-appengine-v1-py/examples/feature_fragments

### DIFF
--- a/gapic/templates/examples/feature_fragments.j2
+++ b/gapic/templates/examples/feature_fragments.j2
@@ -145,7 +145,7 @@ with open({{ attr.input_parameter }}, "rb") as f:
 {% macro render_client_setup(module_name, client_name) %}
 # Create a client
 client = {{ module_name }}.{{ client_name }}()
-{% endmacro %}}
+{% endmacro %}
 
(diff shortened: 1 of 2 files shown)


gapic/templates/examples/feature_fragments.j2
```diff
@@ -145,7 +145,7 @@ with open({{ attr.input_parameter }}, "rb") as f:
 {% macro render_client_setup(module_name, client_name) %}
 # Create a client
 client = {{ module_name }}.{{ client_name }}()
-{% endmacro %}}
+{% endmacro %}
```
busunkim96, line 148: Unrelated fix to the template file.

software-dov (approved):

PR #874, opened 2021-05-11
style: update Python client library templates to align with the Python Style Guide

klmilam:
The template conflicts with the [Python Style Guide](https://google.github.io/styleguide/pyguide.html) in a couple of areas:
- ["Be consistent with your choice of string quote character within a file. Pick ' or " and stick with it. It is okay to use the other quote character on a string to avoid the need to \\ escape within the string."](https://google.github.io/styleguide/pyguide.html#310-strings)
- Lambda functions: ["Okay to use them for one-liners. If the code inside the lambda function is longer than 60-80 chars, it’s probably better to define it as a regular nested function."](https://google.github.io/styleguide/pyguide.html#2104-decision)
- ["Maximum line length is 80 characters."](https://google.github.io/styleguide/pyguide.html#32-line-length)
- [Function docstrings](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) including descriptive-style phrasing and newlines.

This will improve the style of the auto-generated Python client library; however, the changes are not exhaustive because the code is still auto-generated. It does not change any functionality.

### DIFF
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
@@ -1,4 +1,4 @@
-{% extends '_base.py.j2' %}
+{% extends "_base.py.j2" %}
 
 {% block content %}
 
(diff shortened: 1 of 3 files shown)


software-dov:
Most of this is nice to have but not strictly necessary. Due to fundamental limitations of jinja, the whitespace and line length of generated clients is considered to be best effort. Post-processing via black (preferred) or autopep8 is encouraged.

software-dov (approved):

PR #875, opened 2021-05-11
feat: support protobuf method deprecation option [gapic-generator-python]

miraleung:


gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
```diff
@@ -334,6 +334,9 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
                 {{ method.client_output.meta.doc|rst(width=72, indent=16, source_format='rst') }}
         {% endif %}
         """
+        {% if method.is_deprecated %}
+        warnings.warn("{{ method.name|snake_case }} is deprecated", warnings.DeprecationWarning)
```
tseaver, line 338: Two points:

- We likely want to have the warning point back to the caller using `stacklevel`, e.g.:
```suggestion
        warnings.warn("{{ method.name|snake_case }} is deprecated", warnings.DeprecationWarning, stacklevel=1)
```
- Ideally, a deprecation warning would point to the user to the recommended replacement usage.  I have no idea how to do this in generated code -- I'm guessing that the `is_deprecated` flag doesn't have any accompanying descriptive language.

tseaver (changes requested):

gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
miraleung: Done for the first point. For the second, we have [AIP 192](aip.dev/192#deprecations) and an [API linter rule](https://github.com/googleapis/api-linter/pull/836) that will enforce the presence of accompanying descriptive comments, which should get propagated to this method's comments automatically.

tseaver: Hmm, maybe the deprecation message should include a URL to the method's documentation, to make finding that comment easier?

software-dov (commented):
The targeted additions look good. Can you discard the extraneous formatting changes?

miraleung:
> The targeted additions look good. Can you discard the extraneous formatting changes?

Done.

software-dov (approved):

gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
miraleung: As a result of the API linter rule, the first line the RPC comment will be `"Deprecated: <reason>"`, which will show up in the Python method comment. We otherwise don't have the infrastructure set up to locate and propagate a proto's public URL from googleapis, so that first line should be visible and sufficient.

tseaver: FWIW, I meant to suggest that the warning message contain a link to the client library's method docs, rather than the proto docs.  Perhaps there is no standard way to generate such a link, but if there were, it would be useful.

tseaver (approved):

PR #883, opened 2021-05-12
fix: consistently use _pb2 identifier

busunkim96:
Small change to fix bug noticed in https://github.com/googleapis/python-dialogflow/pull/288.

The import was `from google.rpc import status_pb2` but code expected `gr_status`


software-dov (commented):
Can you add a quick test to verify the new behavior?

software-dov (approved):

PR #888, opened 2021-05-17
feat: Add `x-goog-api-client` header to rest clients

vam-google:
Also add `Content-Type: application/json` header.

This PR depends on https://github.com/googleapis/python-api-core/pull/189, and assumes that google-api-core version 1.27.0 has been already released (which was not the case on the moment of creation of this PR).

### DIFF
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
@@ -33,6 +33,12 @@ try:
         gapic_version=pkg_resources.get_distribution(
             '{{ api.naming.warehouse_package_name }}',
         ).version,
+        {% if 'grpc' not in opts.transport %}
+        grpc_version=None,
+        {% endif %}
+        {% if 'rest' in opts.transport %}
+        rest_version=pkg_resources.get_distribution("requests").version,
+        {% endif %}
     )
 except pkg_resources.DistributionNotFound:
     DEFAULT_CLIENT_INFO = gapic_v1.client_info.ClientInfo()
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
@@ -6,6 +6,9 @@ import abc
 from typing import Awaitable, Callable, Dict, Optional, Sequence, Union
 import packaging.version
 import pkg_resources
+{% if 'rest' in opts.transport %}
+from requests import __version__ as requests_version
+{% endif %}
 
 import google.auth  # type: ignore
 import google.api_core  # type: ignore
@@ -37,7 +40,7 @@ try:
         grpc_version=None,
         {% endif %}
         {% if 'rest' in opts.transport %}
-        rest_version=pkg_resources.get_distribution("requests").version,
+        rest_version=requests_version,
         {% endif %}
     )
 except pkg_resources.DistributionNotFound:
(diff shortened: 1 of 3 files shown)


gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -207,8 +207,11 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
         url += '?{}'.format('&'.join(query_params)).replace(' ', '+')
 
         # Send the request
+        headers = dict(metadata)
+        headers['Content-Type'] = 'application/json'
```
software-dov, line 211: Naive question: this is not `x-goog-api-client`. Where does the name `content-type` come from? Is there documentation somewhere describing it?

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
```diff
@@ -33,6 +33,12 @@ try:
         gapic_version=pkg_resources.get_distribution(
             '{{ api.naming.warehouse_package_name }}',
         ).version,
+        {% if 'grpc' not in opts.transport %}
+        grpc_version=None,
+        {% endif %}
+        {% if 'rest' in opts.transport %}
+        rest_version=pkg_resources.get_distribution("requests").version,
```
busunkim96, line 40: Can we use `requests.__version__` instead? We're trying to move away from using `pkg_resources` as it may not work depending on how the library is installed.

```py
Python 3.8.3 (default, Jun 15 2020, 16:29:21) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import requests
>>> requests.__version__
'2.24.0'
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
vam-google: The `Content-Type` header is part of HTTP standard: https://datatracker.ietf.org/doc/html/rfc2616#section-14.17

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
vam-google: Sure, if it is the better way of doing it!

vam-google: Actually that implies that I would have to import `requests` module explicitly in `base.py` files. If that is ok with you I'll do it, but please confirm.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
vchudnov-g: @software-dov See also the internal REGAPIC gRPC transcoding design doc, where we call this out explicitly (relatively recent addition, since we found some RPCs fail without it)

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
busunkim96: `requests` is a requirement of `google-api-core` so I am OK with importing it explicitly in the GAPIC surface.

https://github.com/googleapis/python-api-core/blob/c4ff747db17cdeedb0eb2741b308c7b2d2a596fb/setup.py#L35

vam-google:
@busunkim96 PTAL

busunkim96 (approved):
LGTM. 

I suspect the coverage drop is from the clauses I added to keep compatibility with older google-auth and google-api-core versions in https://github.com/googleapis/gapic-generator-python/commit/89d6f35c54b0a9b81c9b5f580d2e9eb87352ed93#diff-f7a16a65f061822bcc73b8296f4dc837353d379d8d9cc5307982cb6941442835.

busunkim96:
Opened PR to remove code / tests for google-api-core < 1.26.0. See https://github.com/googleapis/gapic-generator-python/pull/893

Once that is merged updating this branch should resolve the coverage issues.

busunkim96:
@vam-google CI is now passing.

PR #893, opened 2021-05-19
fix: remove support for google-api-core<1.26.0

busunkim96:
Drop support for google-api-core < 1.26.0. This code was originally added in #774.

This should unblock the coverage failures in the generated showcase unit tests for #888.


tseaver (approved):
Would be nice to be able to drop support for the older `google-auth` versions, too.

busunkim96:
@tseaver Definitely! I'm planning on a cutting a release for google-api-core that requires a newer google-auth version.

PR #894, opened 2021-05-19
Update Ads templates to support Python 3.6 as a minimum

BenRKarl:
Includes a backport to `pep562` module to support module-level __getattr__ methods where used.

### DIFF
--- a/gapic/ads-templates/%namespace/%name/%version/__init__.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/__init__.py.j2
@@ -6,8 +6,8 @@ import importlib
 import sys
 
 
-if sys.version_info < (3, 7):
-    raise ImportError('This module requires Python 3.7 or later.')
+if sys.version_info < (3, 6):
+    raise ImportError('This module requires Python 3.6 or later.')
 
 
 _lazy_type_to_package_map = {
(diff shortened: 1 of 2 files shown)


gapic/ads-templates/%namespace/%name/%version/__init__.py.j2
```diff
@@ -108,3 +108,9 @@ __all__ = (
 )
 {% endif %} {# lazy import #}
 {% endblock %}
+
+
+if not sys.version_info >= (3, 7):
+    from pep562 import Pep562
```
tseaver, line 114: This would seem to require adding the following to `gapic/ads-templates/setup.py.j2`:

```python
...
 setuptools.setup(
     ....
     install_requires=(
         'google-api-core >= 1.22.2, < 2.0.0dev',
         'googleapis-common-protos >= 1.5.8',
         'grpcio >= 1.10.0',
         'proto-plus >= 1.15.0',
+        'pep562; python_version < 3.7',
         ...
```

busunkim96: I believe you can also enable the 3.6 showcase unit tests:

https://github.com/googleapis/gapic-generator-python/blob/81932a2b71e6ca5f424ddc5c52933ad1d452583a/.github/workflows/tests.yaml#L174-L181

BenRKarl:
Closing this because we're going to deprecate 3.6 in the next release, so this change is no longer needed.

PR #900, opened 2021-05-20
feat: Determine URL protocol based on hostname

vchudnov-g:
This will make connections to `localhost` use HTTP. This is insecure but useful for testing. In a future change, we will allow setting transport options at client instantiation time.


gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -230,6 +231,33 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
...
+
+    Returns:
+        str: The protocol (`http` or `https`) to use for connecting to `host`.
+    """
+    
+    LOCALHOST='localhost'
+    LOCALHOST_LEN=len(LOCALHOST)
+    
+    if host.startswith(LOCALHOST) and (len(host)==LOCALHOST_LEN or host[LOCALHOST_LEN]==":"):
+        return 'http'
```
busunkim96, line 259: I have a slight preference for regex match here instead (especially since the docstring already seems to describe this in terms of regex :)) https://docs.python.org/3/library/re.html

busunkim96 (commented):
Could you add a unit test that exercises `protocol_for`?

vchudnov-g (commented):
Made the code changes. Still need to implement the unit test. (Need to figure out how to run the tests locally first!)

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -230,6 +232,29 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
...
+         parameters, such as secure/insecure connections, and have tests use
+         that. Then remove this function.
+         Tracking issue:
+         https://github.com/googleapis/gapic-generator-python/issues/901
+#}
+def _protocol_for(host: str) -> str:
+    """Determine the URL protocol to use for connecting to the given host.
+
+    This function returns 'http' for hosts that match
+    `^localhost(:.*)?$` and `https` for all others. This is intended
```
vam-google, line 245: Do we have anything like this in the other languages? Overall this looks suspicious I would prefer to not have something like this. Determining protocol from host name does not seem reliable and also we are modifying production code for testing purposes. Also what if the host is specified as an IP (land specifically 127.0.0.1)?

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -230,6 +232,29 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
...
+    locally running server.
+
+    Args:
+        host (str): The hostname for REST connection.
+
+    Returns:
+        str: The protocol (`http` or `https`) to use for connecting to `host`.
+    """
+    if not hasattr(_protocol_for, 'regex'):
+        _protocol_for.regex = re.compile("^localhost(:.*)?$")
```
vam-google, line 256: Will this regex be executed for every call made? If yes, it will have some performance implications (not big probably, but here we sacrifice production code to support rare testing case, which is suboptimal).

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
vchudnov-g: Yes. This was discussed in an internal doc; I'll ping you the link. We're specifically looking for `localhost` because yes, this is a temporary special case to allow testing (see https://github.com/googleapis/gapic-generator-python/issues/901 for the long-term solution)

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
vchudnov-g: The compilation of the regex will only happen once because I have it memoized via this attribute.

The matching will happen for every call for now, though we could refactor things so the protocol is only computed once per client and stored in an attribute. However, given that the performance hit is small, I'm inclined to think it's preferable to keep things simple.

software-dov (approved):

miraleung (approved):

parthea:
I'm going to close this PR due to it being stale but please feel free to re-open and update it.

PR #905, opened 2021-05-21
feat(tests): Add integration test framework, goldens for 4 APIs [gapic-generator-python]

miraleung:
The relevant files to review are those ending in `.bzl`. Any `.py` files in this PR are generated goldens only.


.github/workflows/tests.yaml
```diff
@@ -330,4 +330,4 @@ jobs:
             python -m pip install autopep8
       - name: Check diff
         run: |
-            find gapic tests -name "*.py" | xargs autopep8 --diff --exit-code
+            find gapic tests -name "*.py" -not -path 'tests/integration/goldens/*' | xargs autopep8 --diff --exit-code
```
busunkim96, line 333: Can a job be added that checks the goldens?

busunkim96 (commented):
Thanks @miraleung! My Bazel knowledge is lacking so I'll let other folks review.

.github/workflows/tests.yaml
miraleung: Done.

vam-google (commented):
Overall looks good to me, but a few general questions/concerns:

1) It looks like a massive update, do we really need that many golden files? Can we pick like one API instead (with most feature coverage in it)? This is simply to reduce the number of goldens to maintain.

2) Most importantly this will change the development workflow - with every new feature or substantial modification, the golden tests most probably will have to be updated. Also, unfortunately, bazel is and alien technology for most of the Python developers (including gapic-generator-python developers), and making them to deal with unexpected nasty failures and necessity to run/update the tests in the bazel way can make them really unhappy. Basically I was under impression that if we do goldens for gapic-generator-python it must be done in the python-idiomatic way. Comparing to gapic-generato-java - the generator itself is build via bazel, so using bazel for goldens infra makes perfect sense. For python the things are a bit different. I'm afraid if we introduce the goldens in this current form, relatively soon the maintainers of the repo will remove them as very painful/non-idiomatic to maintain.

3) Even if we go with the goldens-via-bazel infra, is there some documentation speaking about how the development workflow is going to be changed by this, what developers need to run before posting the PR to make sure that the stuff passes and what to do if the goldens need to be updated (sorry if it is already there - i did try to all potential readme/.rst files in this PR but did not find documentation like that in those files).

miraleung:
Comments inline:

1. In the Java generator, there were often distinct edge cases between different APIs - there isn't one that provides the most coverage. Maintaining golden files should be easy, since one only needs to run the Bazel target to update them all (and presumably devs should check the diffs manually to validate correctness).  This set is meant to be a starter, will leave it to @busunkim96 and @software-dov  to decide how many or which APIs we should actually be used.
2. Since the microgenerator has Bazel rules, using Bazel in this repo seems to align with that. It also seems like we are missing a test path that validates they continue to work as expected, which is an added benefit of these integration tests. 
3. Updated `DEVELOPMENT.md`. The Git hooks in #908 will ensure all the pre-commit checks will run automatically on each commit.

vam-google (approved):
LGTM, but please, wait for Bu Sun an Dov to approve (and also, still hope that we can reduce the number of golden files here).

busunkim96 (approved):
LGTM. Thanks Mira!

PR #908, opened 2021-05-26
feat(dev): Add Git pre-commit hooks [gapic-generator-python]

miraleung:
After copying this into `.git/hooks` as described in `DEVELOMPENT.md`, these checks (a subset of CI tasks) will run locally on each commit.


.githooks/pre-commit
```diff
@@ -0,0 +1,132 @@
+#!/bin/sh
```
busunkim96, line 1: I believe this needs a license header

busunkim96 (approved):

.githooks/pre-commit
miraleung: Done.

PR #914, opened 2021-06-08
fix: samplegen always produces sample dicts with "response"

busunkim96:
- fix samplegen logic to always set a `sample[response]`
- enable `autogen-snippets` for the golden integration tests that are run through Bazel
- install some extra stub packages for mypy that are needed with the new release

### DIFF
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -301,8 +301,6 @@ class Validator:
         # Add reasonable defaults depending on the type of the sample
         if not rpc.void:
             sample.setdefault("response", [{"print": ["%s", "$resp"]}])
-        else:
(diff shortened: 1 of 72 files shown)


gapic/samplegen/samplegen.py
```diff
@@ -301,6 +301,8 @@ def preprocess_sample(sample, api_schema: api.API, rpc: wrappers.Method):
         # Add reasonable defaults depending on the type of the sample
         if not rpc.void:
             sample.setdefault("response", [{"print": ["%s", "$resp"]}])
+        else:
+            sample["response"] = []
```
busunkim96, line 305: Code / templates expect `response` to exist.

parthea (approved):

PR #915, opened 2021-06-08
chore: pin mypy version to 0.812

busunkim96:
https://github.com/python/mypy/issues/10596

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -273,7 +273,8 @@ def showcase_mypy(
     """Perform typecheck analysis on the generated Showcase library."""
 
     # Install pytest and gapic-generator-python
-    session.install("mypy")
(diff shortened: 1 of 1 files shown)


noxfile.py
```diff
@@ -273,7 +273,8 @@ def showcase_mypy(
     """Perform typecheck analysis on the generated Showcase library."""
 
     # Install pytest and gapic-generator-python
-    session.install("mypy")
+    # Pin to 0.89.0 until fix is released for https://github.com/python/mypy/issues/10596
+    session.install("mypy==0.812")
```
tseaver, line 277: https://pypi.org/project/mypy/0.901/, just now released, is supposed to have a fix for that bug.

busunkim96: Ah great! I'll retry my other PRs then. :D

PR #918, opened 2021-06-10
chore: Update Showcase to the latest released version, 0.16.0

vchudnov-g:


busunkim96 (approved):

software-dov (commented):
I think we can just delete the circleci configs instead of keeping them updated.

I'm very confused about the errors as well.

arithmetic1728:
> I think we can just delete the circleci configs instead of keeping them updated.
> 
> I'm very confused about the errors as well.

Yes, that works for me.

vchudnov-g:
@software-dov @busunkim96 Do these sound like spurious failures? All checks passed before merging master....

vchudnov-g:
So it looks like we're failing because the generated Showcase GAPIC has 0% code coverage, I guess because it doesn't have unit tests (?). I'm confused why this wasn't a problem before this PR.

Suggestions?

software-dov:
Took a look at the failing CI, and I believe that the code cov failure is possibly a red herring. The tests definitely run locally, definitely run in CI, but the coverage isn't being tracked correctly.

software-dov:
The update made it's way in via another PR. @vchudnov-g mind if we close this?

vchudnov-g:
Yes. let's close this, since this was done in #925

PR #937, opened 2021-06-28
test: re-add coverage requirement

busunkim96:
* Re-add `cov-fail-under=100` to unit test sessions to enforce coverage requirement.

* Remove codecov checks from expected status checks.
* Add unit test for `any_deprecated` added in #930


busunkim96:
Interestingly`showcase-unit-3.6` seems to pick up coverage while `showcase_unit_alternative_templates-3.7` fails to pick up any coverage. 🤔

busunkim96:
@parthea @software-dov I'm not sure why, but doing an editable install `-e .` seems to make coverage happy again. PTAL.

parthea (approved):

PR #939, opened 2021-06-28
fix: disable always_use_jwt_access

busunkim96:
Some APIs with Cloud Storage integrations seem to have failing
samples when non-default scopes are provided. This PR disables the feature
globally for now while we investigate.

We can no longer rollback da119c72c8 as some libraries have been
released with the change and removing the kwarg would be a breaking
change.

See internal issue 192297181 .

### DIFF
--- a/.github/sync-repo-settings.yaml
+++ b/.github/sync-repo-settings.yaml
@@ -6,8 +6,6 @@ branchProtectionRules:
   isAdminEnforced: true
   requiredStatusCheckContexts:
     - 'cla/google'
-    - 'codecov/patch'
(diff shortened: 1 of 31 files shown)


arithmetic1728 (approved):

parthea (approved):

busunkim96:
@arithmetic1728 @parthea Could I get another review? I looked at this again and the logic seemed off.

I added `always_use_jwt_access` to the `grpc` and `grpc_asyncio` transport constructors to match the base constructor.

arithmetic1728 (approved):

Issue #701, opened 2020-11-18
Resource patterns with wildcard '*' results in failing test for parse_path


busunkim96:
Asset has a resource path annotated with pattern `*`

https://github.com/googleapis/googleapis/blob/f1e0ce4dbc5f59448f441093014d712af442984f/google/cloud/asset/v1/assets.proto#L96-L100

```proto
message Asset {
  option (google.api.resource) = {
    type: "cloudasset.googleapis.com/Asset"
    pattern: "*"
  };
```

This results in a unhelpful set of resource name helpers: 
```py
    @staticmethod
    def asset_path() -> str:
        """Return a fully-qualified asset string."""
        return "*".format()

    @staticmethod
    def parse_asset_path(path: str) -> Dict[str, str]:
        """Parse a asset path into its component segments."""
        m = re.match(r"^*$", path)
        return m.groupdict() if m else {}

```

The generated tests break because of the `*`.
```py
def test_asset_path():
    expected = "*".format()
    actual = AssetServiceClient.asset_path()
    assert expected == actual


def test_parse_asset_path():
    expected = {}
    path = AssetServiceClient.asset_path(**expected)

    # Check that the path construction is reversible.
    actual = AssetServiceClient.parse_asset_path(path)
    assert expected == actual

```

```
                if not item or item[0][0] is AT:
>                   raise source.error("nothing to repeat",
                                       source.tell() - here + len(this))
E                   re.error: nothing to repeat at position 1

../../.pyenv/versions/3.8.1/lib/python3.8/sre_parse.py:668: error
```

The `*` pattern is uncommon and this method isn't actually helpful for parsing paths. I'm planning on deleting it from the generated output for now.

busunkim96:
Similar issue with monitoring which has a path with `**`. 

https://github.com/googleapis/googleapis/blob/69697504d9eba1d064820c3085b4750767be6d08/google/monitoring/v3/metric_service.proto#L39-L46

```proto
option (google.api.resource_definition) = {
  type: "monitoring.googleapis.com/MetricDescriptor"
  pattern: "projects/{project}/metricDescriptors/{metric_descriptor=**}"
  pattern: "organizations/{organization}/metricDescriptors/{metric_descriptor=**}"
  pattern: "folders/{folder}/metricDescriptors/{metric_descriptor=**}"
  pattern: "*"
  history: ORIGINALLY_SINGLE_PATTERN
};
```

```py
    def metric_descriptor_path(project: str,) -> str:
[...]

PR #941, opened 2021-06-29
feat(snippetgen): generate mock input for required fields

busunkim96:
Generate mock field inputs for all required fields.

* For oneofs, the first option is selected.
* For enums, the last option is selected (as 0 is often unspecified).
* Mock input for string fields that reference resources (`resource_pb2.resource_reference`) is the pattern provided in the protos.

**TODOs:**
- When a resource's pattern is `*` this results in unhelpful strings
    ```proto
    message Asset {
      option (google.api.resource) = {
        type: "cloudasset.googleapis.com/Asset"
        pattern: "*"
      };
    ```

    ```py
    request = asset_v1.BatchGetAssetsHistoryRequest(
       parent="*",
    )
    ```
- Map fields are not yet handled.

**Other changes:**

* Fields are set like `foo.bar` rather than through dict access `foo["bar"]`. This is because the former allows you to set fields nested more than one field deep `foo.bar.baz`.

    **Before**:
    ```py
    output_config = {}
    output_config["gcs_destination"]["uri"] = "uri_value" # fails as there is no nested dict
    ```

    **After**:
    ```py
    output_config = asset_v1.IamPolicyAnalysisOutputConfig()
    output_config.gcs_destination.uri = "uri_value"
    ```


tests/integration/goldens/asset/samples/generated_samples/cloudasset_generated_asset_v1_asset_service_batch_get_assets_history_sync.py
```diff
@@ -35,6 +35,7 @@ def sample_batch_get_assets_history():
 
     # Initialize request argument(s)
     request = asset_v1.BatchGetAssetsHistoryRequest(
+        parent="*",
```
leahecole, line 38: I'm not familiar with this syntax - is this asterisk something that's typical for this service?

tests/integration/goldens/asset/samples/generated_samples/cloudasset_generated_asset_v1_asset_service_create_feed_sync.py
```diff
@@ -34,7 +34,13 @@ def sample_create_feed():
...
 
     # Initialize request argument(s)
+    feed = asset_v1.Feed()
+    feed.name = "name_value"
+
     request = asset_v1.CreateFeedRequest(
+        parent="parent_value",
+        feed_id="feed_id_value",
+        feed=feed,
     )
```
leahecole, line 45: Okay I know it's not part of this generation, but I noticed .format - are we supporting less than Python 3.6 and that's why no f-strings?

tests/integration/goldens/asset/samples/generated_samples/cloudasset_generated_asset_v1_asset_service_delete_feed_async.py
```diff
@@ -35,6 +35,7 @@ async def sample_delete_feed():
 
     # Initialize request argument(s)
     request = asset_v1.DeleteFeedRequest(
+        name="projects/{project}/feeds/{feed}",
```
leahecole, line 38: This syntax seems kind of weird to me - I don't understand what `{project}` and `{feed}` mean

tests/integration/goldens/asset/samples/generated_samples/cloudasset_generated_asset_v1_asset_service_delete_feed_sync.py
```diff
@@ -35,6 +35,7 @@ def sample_delete_feed():
 
     # Initialize request argument(s)
     request = asset_v1.DeleteFeedRequest(
+        name="projects/{project}/feeds/{feed}",
```
leahecole, line 38: Same comment about `{project}` and `{feed}`

tests/integration/goldens/credentials/samples/generated_samples/iamcredentials_generated_credentials_v1_iam_credentials_generate_access_token_async.py
```diff
@@ -35,6 +35,8 @@ async def sample_generate_access_token():
 
     # Initialize request argument(s)
     request = credentials_v1.GenerateAccessTokenRequest(
+        name="projects/{project}/serviceAccounts/{service_account}",
+        scope=['scope_value'],
```
leahecole, line 39: if the scope is of type list is it worth adding more than one list item?

tests/integration/goldens/redis/samples/generated_samples/redis_generated_redis_v1_cloud_redis_update_instance_async.py
```diff
@@ -34,7 +34,13 @@ async def sample_update_instance():
     client = redis_v1.CloudRedisAsyncClient()
 
     # Initialize request argument(s)
+    instance = redis_v1.Instance()
+    instance.name = "name_value"
+    instance.tier = "STANDARD_HA"
+    instance.memory_size_gb = 1499
```
leahecole, line 40: i'm very curious at why 1499 is this generated value? Is it particularly large and could it get someone into trouble by initializing something very expensive?

tests/snippetgen/goldens/mollusca_generated_mollusca_v1_snippets_method_bidi_streaming_async.py
```diff
@@ -34,11 +34,12 @@ async def sample_method_bidi_streaming():
...
 
     # Initialize request argument(s)
-    request = mollusca_v1.SignatureRequest(
+    request = mollusca_v1.SignatureRequestOneRequiredField(
+        a_string="a_string_value",
     )
 
     # Make the request
-    stream = await client.method_bidi_streaming([])
+    stream = await client.method_bidi_streaming([a_string="a_string_value"])
```
leahecole, line 42: nit - is `a_string` "a string" or "a-string" (with the implication of a "b-string" or a "c-string"?) If "a string" and it's something we have the power to change, then I feel like `my_string` or `your_string` is more useful but if not, plz disregard

leahecole (commented):
Okay @busunkim96 - I looked at some - didn't make it through all, but a lot of my comments will apply to multiple samples, I think. Hope it was at least a little helpful!

parthea (approved):

tests/integration/goldens/asset/samples/generated_samples/cloudasset_generated_asset_v1_asset_service_create_feed_sync.py
busunkim96: Nope, no support for less than 3.6. I think the `.format` is from a previous iteration that _was_ supposed to support older Python versions. Opened https://github.com/googleapis/gapic-generator-python/issues/960 to track switching to f-strings.

tests/integration/goldens/asset/samples/generated_samples/cloudasset_generated_asset_v1_asset_service_delete_feed_async.py
busunkim96: I'm copying the [pattern](https://github.com/googleapis/googleapis/blob/5d301f95d17d50f78e0b2a6889301c180b947357/google/cloud/asset/v1/asset_service.proto#L609) verbatim from the protos with the hope it will nudge folks towards the correct resource name.

I guess it is weird that it _almost_ looks like an f-string, and there are no `project` or `feed` variables defined. Would sometime like this be more helpful?

```py
async def sample_delete_feed():
    """Snippet for delete_feed"""
    # Create a client
    client = asset_v1.AssetServiceAsyncClient()

    # Initialize request argument(s)
    project = "my-project"
    feed = "my-feed"

    request = asset_v1.DeleteFeedRequest(
        name=f"projects/{project}/feeds/{feed}",
```

tests/integration/goldens/redis/samples/generated_samples/redis_generated_redis_v1_cloud_redis_update_instance_async.py
busunkim96: The number is deterministically generated from the name of the field. It's re-using the same logic that's used for unit tests. The numbers are fine for unit tests since no real requests get sent, but it might not be that great for samples...

https://github.com/googleapis/gapic-generator-python/blob/5c8eeaf8118932d63729a18f387154aec0b1d6c7/gapic/schema/wrappers.py#L115-L118

 🤔  @amanda-tarafa How does C# generate mock numeric values for snippetgen?

tests/snippetgen/goldens/mollusca_generated_mollusca_v1_snippets_method_bidi_streaming_async.py
busunkim96: Ah this is from the field naming in the mock proto file.  It is completely changeable, I will do that in a follow up commit.  :D https://github.com/googleapis/gapic-generator-python/blob/cc3de569ac610b3ca60259a53c572e60a6cea180/tests/snippetgen/snippets.proto

tests/integration/goldens/credentials/samples/generated_samples/iamcredentials_generated_credentials_v1_iam_credentials_generate_access_token_async.py
busunkim96: That sounds reasonable and should be an easy fix - opened https://github.com/googleapis/gapic-generator-python/issues/961

tests/integration/goldens/redis/samples/generated_samples/redis_generated_redis_v1_cloud_redis_update_instance_async.py
amanda-tarafa: We use the default values for all number type, which is 0. This has the same drawback as generating something as you are, which is, it may or may not work. I think this is fine for now. For generating more meaningfull values, we need configs to be specified, so we can do this for next iterations.

amanda-tarafa: Also, we are adding this note for each snippet:

```
/// This snippet has been automatically generated for illustrative purposes only.
/// It may require modifications to work in your environment.
```

tests/snippetgen/goldens/mollusca_generated_mollusca_v1_snippets_method_bidi_streaming_async.py
busunkim96: Fixed in https://github.com/googleapis/gapic-generator-python/pull/941/commits/95282b502a08d0c72fb45537cd2ed3fb1853462f

tests/integration/goldens/asset/samples/generated_samples/cloudasset_generated_asset_v1_asset_service_batch_get_assets_history_sync.py
busunkim96: The * wouldn't actually work if passed to the service. The snippetgen logic pulls the first resource pattern out of the protos, and in some uncommon cases it is just a `*` rather than some more specific pattern like `projects/{project}`. The `*` pattern creates problems for generating resource helper methods and tests too  - see #701

tests/integration/goldens/asset/samples/generated_samples/cloudasset_generated_asset_v1_asset_service_create_feed_sync.py
software-dov: @busunkim96 is correct. Earlier in the dev process for the python microgenerator, there was an expected need to support 3.5.

tests/integration/goldens/asset/samples/generated_samples/cloudasset_generated_asset_v1_asset_service_delete_feed_async.py
software-dov: The posted example is definitely more clear. +1 for that if it's not too much effort.

busunkim96: Opened https://github.com/googleapis/gapic-generator-python/issues/965 to track making these strings nicer. Thanks for the feedback @leahecole @software-dov 😄

tests/integration/goldens/asset/samples/generated_samples/cloudasset_generated_asset_v1_asset_service_delete_feed_sync.py
busunkim96: Tracking a fix in https://github.com/googleapis/gapic-generator-python/issues/965

tests/integration/goldens/redis/samples/generated_samples/redis_generated_redis_v1_cloud_redis_update_instance_async.py
busunkim96: OK, I will stick with the somewhat random looking numbers, but if anyone thinks a set value (0, 1) would be better it should be an easy change.

gapic/samplegen/samplegen.py
```diff
@@ -940,6 +948,56 @@ def parse_handwritten_specs(sample_configs: Sequence[str]) -> Generator[Dict[str
...
+                {"field": field_name, "value": field.enum.values[-1].name})
+        else:
+            # This is a message type, recurse
+            # TODO(busunkim): when optional fields are supported
+            # add a check to make sure nested messages terminate
+            request += generate_request_object(
+                api_schema, service, field.type,
+                field_name_prefix=field_name,
+            )
+
```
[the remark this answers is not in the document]
busunkim96: 👍  This is definitely something that needs to be addressed as config-based snippets are introduced. I've tweaked the wording in the comment in 70e1cea.

gapic/schema/wrappers.py
```diff
@@ -1199,6 +1266,29 @@ def gen_indirect_resources_used(message):
...
+        service_resource_messages = dict(
+            (r.resource_type_full_path, r)
+            for r in list(self.resource_messages)
+        )
+        common_resource_messages = dict(
+            (resource_path, resource.message_type)
+            for resource_path, resource in self.common_resources.items()
+        )
+
+        service_resource_messages.update(common_resource_messages)
```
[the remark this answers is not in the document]
busunkim96: Changed in 70e1cea

gapic/samplegen/samplegen.py
```diff
@@ -293,17 +293,25 @@ def preprocess_sample(sample, api_schema: api.API, rpc: wrappers.Method):
...
         # Assume the gRPC transport if the transport is not specified
         sample.setdefault("transport", api.TRANSPORT_GRPC)
 
         if sample["transport"] == api.TRANSPORT_GRPC_ASYNC:
             sample["client_name"] = api_schema.services[sample["service"]
                                                         ].async_client_name
         else:
-            sample["client_name"] = api_schema.services[sample["service"]].client_name
+            sample["client_name"] = service.client_name
```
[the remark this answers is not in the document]
busunkim96: Changed in 70e1cea

busunkim96:
@software-dov Thank you for the review! PTAL

gapic/samplegen/samplegen.py
```diff
@@ -940,6 +945,58 @@ def parse_handwritten_specs(sample_configs: Sequence[str]) -> Generator[Dict[str
...
+        field_name = ".".join([field_name_prefix, field.name]).lstrip('.')
+
+        # TODO(busunkim): Properly handle map fields
+        if field.is_primitive:
+            placeholder_value = field.mock_value_original_type
+            # If this field identifies a resource use the resource path
+            if service.resource_messages_dict.get(field.resource_reference):
+                placeholder_value = service.resource_messages_dict[
+                    field.resource_reference].resource_path
+            request.append({"field": field_name, "value": placeholder_value})
```
software-dov, line 981: Wild, wild stylistic preference nitpicking, feel free to ignore.
```python
maybe_res = service.resource_message_dict.get(field.resource_reference)
placeholder_val = maybe_res.resource_path if maybe_res else field.mock_value_original_type
```

gapic/schema/wrappers.py
```diff
@@ -158,6 +164,44 @@ def inner_mock(self, stack, visited_fields):
         # Done; return the mock value.
         return answer
 
+    def inner_mock_as_original_type(self) -> Union[bool, str, bytes, int, float, List[Any], None]:
```
software-dov, line 167: I think it might be simpler and cleaner to rework `mock` and `inner_mock` to return the type and then just convert to string when need be, or make the `_as_original_type` variants the 'real' one and have the others just be thin wrappers. I'm concerned that duplicating much of the logic is going to cause maintenance problems. If this turns out to be non trivial then just ignore this comment.

gapic/schema/wrappers.py
```diff
@@ -1199,6 +1266,29 @@ def gen_indirect_resources_used(message):
...
+
+        Returns:
+            Dict[str, MessageType]: A mapping from resource path
+                string to the corresponding MessageType.
+                `{"locations.googleapis.com/Location": MessageType(...)}`
+        """
+        service_resource_messages = dict(
+            (r.resource_type_full_path, r)
+            for r in list(self.resource_messages)
+        )
```
software-dov, line 1282: Outrageous nit: slightly more elegant to do
```python
service_resource_messages = {r.resource_type_full_path : r for r in self.resource_messages}
```
Don't need to turn `resource_messages` into a list because it's already iterable.

software-dov (commented):
Mostly nitpicking and one minor concern, looks great! Thanks for doing this, and I hope the existing samplegen code wasn't too unpleasant to work with.

gapic/schema/wrappers.py
busunkim96: I tried to de-dupe some of th logic in https://github.com/googleapis/gapic-generator-python/pull/941/commits/b5c8e2495ca70112e4aa57c93ee1058f50271e4b. Let me know what you think. 

The complicating factor seems to be that different primitive types need different transformations (strings need to be wrapped with another pair of quotes, ints and floats can just be turned into strings, etc.)  to render correctly in the client library unit test templates.

software-dov (approved):

PR #945, opened 2021-07-01
fix: remove "type" from set of reserved names

software-dov:
Fix for https://github.com/googleapis/proto-plus-python/issues/235

### DIFF
--- a/gapic/utils/reserved_names.py
+++ b/gapic/utils/reserved_names.py
@@ -24,6 +24,6 @@ RESERVED_NAMES = frozenset(
         # We CANNOT make exceptions for keywords.
         keyword.kwlist,
         # We make SOME exceptions for certain names that collide with builtins.
-        set(dir(builtins)) - {"filter", "map", "id", "input", "property"},
+        set(dir(builtins)) - {"filter", "map", "id", "input", "property", "type"},
     )
 )
--- a/gapic/ads-templates/noxfile.py.j2
+++ b/gapic/ads-templates/noxfile.py.j2
@@ -7,7 +7,7 @@ import os
 import nox  # type: ignore
 
 
-@nox.session(python=['3.7', '3.8'])
+@nox.session(python=['3.7', '3.8', '3.9'])
 def unit(session):
     """Run the unit test suite."""
 
@@ -25,7 +25,7 @@ def unit(session):
     )
 
 
-@nox.session(python=['3.7', '3.8'])
+@nox.session(python=['3.7', '3.8', '3.9'])
 def mypy(session):
     """Run the type checker."""
     session.install('mypy')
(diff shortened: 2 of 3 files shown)


gapic/utils/reserved_names.py
```diff
@@ -24,6 +24,6 @@
         # We CANNOT make exceptions for keywords.
         keyword.kwlist,
         # We make SOME exceptions for certain names that collide with builtins.
-        set(dir(builtins)) - {"filter", "map", "id", "input", "property"},
+        set(dir(builtins)) - {"filter", "map", "id", "input", "property", "type"},
```
busunkim96, line 27: I believe this change on its own would cause breaking changes in many GA libraries that already have `type_` for fields named `type`. (See usage in [googleapis-gen](https://github.com/googleapis/googleapis-gen/search?l=Python&q=type_)).

software-dov: Correct. This is a breaking change in multiple generated surfaces in order to fix an issue with field masks in Ads.
_Any_ solution to that problem generates a breaking change at least for Ads. The alternative I have involves a somewhat convoluted and unergonomic change in proto-plus; I can discuss the alternative if you think it's better to restrict the breaking change and move the unpleasantness into proto-plus.

busunkim96: Given the new breaking change and deprecation policy for Cloud I would prefer a solution that doesn't involve breaking changes on the client library surface. If this turns out to be the only option I believe we'll have to go through a formal approval process.

BenRKarl: Also my issue with one-off'ing `type` is we run the risk of hitting this same problem with another reserved word. It's not out of the question for a field to be named something like: `range`, `list`, `input`, `bytes`, `max`, etc. I'd rather retain the existing interface and find a solution that solves the problem for all reserved words.

BenRKarl: Also I'm assuming this isn't just an issue for Ads, is it? Do the other libraries not have mutable fields named `type` or do they construct field masks differently?

busunkim96: @BenRKarl Yep I don't think this is Ads-specific. Cloud definitely has fields named `type`.  I think we've either gotten lucky and don't have this into this particular combo (reserved name + field mask), or that no one has field an issue about this yet.

software-dov: I don't think there's a way of resolving the issue without having a breaking change for ads, at least.
I'll post my alternative workaround in proto-plus as a draft and link it here; people can take a look and consider the alternatives.

software-dov: Proto-plus change is here: https://github.com/googleapis/proto-plus-python/pull/237

software-dov:
Closed in favor of https://github.com/googleapis/proto-plus-python/pull/237

PR #951, opened 2021-07-10
fix: fix rest transport unit test and required query prams handling

vam-google:


gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -196,23 +198,17 @@ class {{ service.name }}RestTransport({{ service.name }}Transport):
         if {{ method.input.ident }}.{{ field }} in request:
             query_params['{{ field|camel_case }}'] = request.{{ field }}
         {% else %}
-        if request.{{ field }}:
-            query_params['{{ field|camel_case }}'] = request.{{ field }}
+        query_params['{{ field|camel_case }}'] = request.{{ field }}
```
busunkim96, line 201: Is passing empty values no longer a concern? https://github.com/googleapis/gapic-generator-python/pull/868

busunkim96 (approved):

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
vam-google: Kind of. I have hard feelings about it. First - this is possible only for required fields, which are not common. For the requried fields, they are required to be set, and if somebody set them to empty value, then, supposedly, they know what they are doing.

PR #957, opened 2021-07-15
fix: use mock from unittest if available

major:
Try to import `mock` from the `unittest` builtin package if possible, otherwise fall back to importing the `mock` package.

### DIFF
--- a/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
+++ b/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
@@ -3,7 +3,10 @@
 {% block content %}
 
 import os
-import mock
+try:
+    from unittest import mock
+except ImportError:
+    import mock
 import packaging.version
 
 import grpc


tseaver (commented):
In an ideal world, we would just have `from unittest import mock` (the generated code no longer has to support Python 2.7).  However, the stdlib versions of `unittest.mock` is missing some features from the third-party shim:  notably, the 3.6 and 3.7 versions lack `mock.AsyncMock`, which we need even for some synchronous tests.

major:
@tseaver Okay, thanks for the explanation. I'm packaging some things for Fedora Linux and I'll carry a patch to make these tests work.

tseaver:
@major Thank you for the patch! I'm sorry we are stuck with supporting the older Python3 versions -- in fact, I knew about the `mock.AsyncMock` bit because I have been working on dropping Python 2.7 support from the core libraries (see [this PR](https://github.com/googleapis/python-api-core/pull/212/#issuecomment-862631125) where I first noticed it).

PR #958, opened 2021-07-21
fix: enable self signed jwt for grpc

arithmetic1728:
b/192297181 is fixed, this PR re-enables self signed jwt for grpc transport.


arithmetic1728:
Just verified b/192297181 is fixed. Please take a look at this PR. Thanks!

busunkim96 (approved):

PR #966, opened 2021-08-03
fix: add 'dict' type annotation to 'request'

busunkim96:
The generated libraries have always accepted both a generated type or an equivalent dict. This PR updates the type hints/docstrings to include dict. 

It looks like [`TypedDict`](https://docs.python.org/3/library/typing.html#typing.TypedDict) could be useful too, but we won't be able to use it until 3.8 is the lowest supported python version.

### DIFF
--- a/tests/integration/goldens/asset/google/cloud/asset_v1/services/asset_service/client.py
+++ b/tests/integration/goldens/asset/google/cloud/asset_v1/services/asset_service/client.py
@@ -351,7 +351,7 @@ class AssetServiceClient(metaclass=AssetServiceClientMeta):
             )
 
     def export_assets(self,
-            request: asset_service.ExportAssetsRequest = None,
+            request: Union[asset_service.ExportAssetsRequest, dict] = None,
             *,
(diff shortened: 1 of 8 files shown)


parthea (approved):

tests/integration/goldens/asset/google/cloud/asset_v1/services/asset_service/client.py
```diff
@@ -442,7 +442,7 @@ def list_assets(self,
         paged results in response.
 
         Args:
-            request (google.cloud.asset_v1.types.ListAssetsRequest):
+            request (Union[google.cloud.asset_v1.types.ListAssetsRequest, dict]):
```
tseaver, line 445: Not an issue with this PR, but why are we using parens inconsistenly among these generated methods?

busunkim96: Are there parens missing for some args? I'm having trouble finding it.

PR #969, opened 2021-08-11
feat: add api key support

arithmetic1728:
This PR adds API key support. Internal doc: go/gapic-api-key-support

If api_key is provided via client_options, use the api key. Note that GOOGLE_API_KEY is handled by google-auth so we don't need to do anything in microgenerator.

Related PRs:
https://github.com/googleapis/google-auth-library-python/pull/826
https://github.com/googleapis/python-api-core/pull/248

Tested with language api:
https://github.com/arithmetic1728/python-language/commit/da9b1215d1a46e5dc2ccc10bf9a6d1edd9cb8940

### DIFF
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
@@ -309,12 +309,16 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
 
         api_endpoint, client_cert_source_func = self.get_mtls_endpoint_and_cert_source(client_options)
 
+        api_key_value = getattr(client_options, "api_key", None)
+        if api_key_value and credentials:
+            raise ValueError("client_options.api_key and credentials are mutually exclusive")
+
         # Save or instantiate the transport.
         # Ordinarily, we provide the transport, but allowing a custom transport
         # instance provides an extensibility point for unusual situations.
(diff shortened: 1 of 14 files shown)


gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -295,10 +298,19 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
                 )
             self._transport = transport
         else:
+            credentials_file = client_options.credentials_file
+            if hasattr(_default, "_get_api_key_credentials"):
```
busunkim96, line 302: Can this be implemented without depending on private modules in google-auth?

arithmetic1728: Updated the code to use a public `api_key.get_api_key_credentials` method. `google.auth.api_key` module is not available yet, so the unit tests for api key are skipped unless google-auth version >= 2.0.3 (next google-auth release). If the api key PR doesn't go into release 2.0.3 in the future, I will update the required version in the tests.

arithmetic1728: currently the code coverage will fail. It should work once the `api_key` module is added to google-auth

Pojielp (approved):

busunkim96 (approved):

PR #970, opened 2021-08-16
fix: improper types in pagers generation

dpcollins-google:
Generators are Iterators not Iterables https://docs.python.org/3/library/typing.html#typing.Generator

### DIFF
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/pagers.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/pagers.py.j2
@@ -7,7 +7,7 @@
 {# This lives within the loop in order to ensure that this template
    is empty if there are no paged methods.
  #}
-from typing import Any, AsyncIterable, Awaitable, Callable, Iterable, Sequence, Tuple, Optional
+from typing import Any, AsyncIterator, Awaitable, Callable, Sequence, Tuple, Optional, Iterator
 
 {% filter sort_lines %}
 {% for method in service.methods.values() | selectattr('paged_result_field') %}
@@ -65,7 +65,7 @@ class {{ method.name }}Pager:
         return getattr(self._response, name)
 
     @property
-    def pages(self) -> Iterable[{{ method.output.ident }}]:
+    def pages(self) -> Iterator[{{ method.output.ident }}]:
         yield self._response
         while self._response.next_page_token:
             self._request.page_token = self._response.next_page_token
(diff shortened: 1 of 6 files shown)


tseaver (approved):
The change is correct, if a bit nit-picky (`Iterable` is a less restrictive contract than `Iterator`, and a strict superset).  Do we have cases where type-checking actually complains about the difference?  I guess that would be clients calling `next(thing)` directly?

dpcollins-google:
> The change is correct, if a bit nit-picky (`Iterable` is a less restrictive contract than `Iterator`, and a strict superset). Do we have cases where type-checking actually complains about the difference? I guess that would be clients calling `next(thing)` directly?

This is not quite correct. Iterables are *not* required to have __next__ methods, they only usually do :) https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable

In fact, you can see this with <*>Pager itself: it is an Iterable (with __iter__) but not an Iterator (no __next__)

busunkim96:
@dpcollins-google could you update the golden files? That should resolve the one failing `Tests / integration` check. See [instructions here](https://github.com/googleapis/gapic-generator-python/blob/master/DEVELOPMENT.md#integration-tests).

tseaver:
@dpcollins-google I think we are in violent agreement here:  `Iterable` requires only `__iter__`, while `Iterator` must also have `__next__` which is a) more restrictive than `Iterable`, and b) makes the contract for `Iterator` a "sub-type" of `Iterable` (no types which implement `Iterator` do not also implement `Iterable`).

At any rate, my question remains:  do we have reported errors from `pytype` in this repo, or even in the wild, where declaring the less-restrictive type (`Iterable`) is causing errors?  My surmise is that either in this repo or elsewhere, somebody is using `next(response)`, rather than `for foo in response:`.

busunkim96:
It looks like this is causing a problem for Pub/Sub (see [log](https://source.cloud.google.com/results/invocations/0455f961-697c-44ac-bc98-e8220a89dd2f/targets;collapsed=/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-pubsublite%2Fpresubmit%2Fpresubmit/log), so I'm inclined to merge it now and re-visit if it causes an issue. 

@software-dov @atulep Could you also take a look as the generator owners?

atulep:
From https://docs.python.org/3/library/typing.html#typing.Generator, generator can have a return type of either `Iterable` or `Iterator`. Why is this change needed?

Does Pub/Sub fail because a caller of the `pager` expects an iterator instead of iterable?

dpcollins-google:
Please see my comment above. The Pager itself does not model `Iterable` because its `__iter__` method states incorrectly that it returns an `Iterable` when it needs to return an `Iterator`.

atulep (approved):

PR #976, opened 2021-08-20
fix: timeouts are handled by rest clients, retries silently ignored

software-dov:
Passes hand tests

### DIFF
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
@@ -9,6 +9,7 @@ from typing import Callable, Dict, Optional, Sequence, Tuple
 from google.api_core import operations_v1
 {% endif %}
 from google.api_core import gapic_v1       # type: ignore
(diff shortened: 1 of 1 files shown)


busunkim96 (approved):
Is the plan to handle retries at a later date?

It looks like the two manually authored REST clients [BigQuery](√blob/519d99c20e7d1101f76981f3de036fdf3c7a4ecc/google/cloud/bigquery/retry.py) and [Storage](https://github.com/googleapis/python-storage/blob/4abb40310eca7ec45afc4bc5e4dfafbe083e74d2/google/cloud/storage/retry.py) have their own retry class that uses google-api-core's retry object. It might be nicer to create a REST equivalent of [`_GapicCallable`](https://github.com/googleapis/python-api-core/blob/master/google/api_core/gapic_v1/method.py) for REGAPIC instead of interacting with AuthorizedSession directly.

software-dov:
I believe that's the plan, yes. In the meantime, it's better to allow the parameter we can support easily and gracefully drop the one that will require more effort than blowing up because we weren't expecting the parameter.

PR #983, opened 2021-08-30
fix: disable self signed jwt if cred is provided

arithmetic1728:
Self signed jwt doesn't work for non-prod endpoint, so this PR allows users to pass credentials to disable self signed jwt.

see https://github.com/googleapis/python-aiplatform/issues/553#issuecomment-887124490, b/197117709

### DIFF
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
@@ -305,7 +305,7 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
                 quota_project_id=client_options.quota_project_id,
                 client_info=client_info,
                 {% if "grpc" in opts.transport %}
-                always_use_jwt_access=(
+                always_use_jwt_access= not credentials and (
                     Transport == type(self).get_transport_class("grpc")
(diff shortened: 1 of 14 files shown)


parthea (approved):

busunkim96 (commented):
Isn't the self-signed JWT usable most of the time even when the user passes their own credential?

arithmetic1728:
@busunkim96 The problem is that currently we always turn on self signed jwt and there is no way to disable it.

In client we always pass `always_use_jwt_access=True` to transport:

https://github.com/googleapis/gapic-generator-python/blob/f6dc5f546e737c3b55af8604b3d818c69e84a1cb/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/client.py.j2#L308-L311

then in transport, self signed jwt is enabled

https://github.com/googleapis/gapic-generator-python/blob/f6dc5f546e737c3b55af8604b3d818c69e84a1cb/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/transports/base.py.j2#L132-L133

therefore the walk around provided in https://github.com/googleapis/python-aiplatform/issues/553#issuecomment-887124490 won't work. I don't see a good way to let the users disable self signed jwt except the walk around though, any thoughts?

PR #987, opened 2021-09-07
feat: add support for context manager in client

### LINKED ISSUE #575: Feature Request: allow clients to act as context manager and add close() method to cleanup resources
Since a `transport` object may be passed into a client constructor, the client does not close that transport in when the client object is deleted. This results in leaking socket resources in programs that create client objects dynamically.

**Proposed design:**

Add the following two ways to explicitly clean-up sockets owned by the client: (1, preferred) allow clients to act as a context manager and (2) a close() method.

1.  Context manager

    Allow all client objects to be used as a context manager. Call close() on the client in the __exit__ method to clean up sockets owned by the client.

    In code samples with short-lived clients, this is the cleanup method we should recommend.

Example:

```python
from google.cloud import bigquery_datatransfer_v1

with bigquery_datatransfer_v1.DataTransferServiceClient() as client:
    print(client.get_transfer_run('some run id').state)
```

2.  `close()` method

    Add a `close()` method to all client objects. When called, clean up sockets owned by the client, including the transport channel.

    Example:

```python
from google.cloud import bigquery_datatransfer_v1

client = bigquery_datatransfer_v1.DataTransferServiceClient()
print(client.get_transfer_run('some run id').state)
client.close()  # clean up any open sockets
```

**References:**

* go/closable-clients-python (Google internal)
* https://github.com/googleapis/google-cloud-python/issues/9457
* https://github.com/googleapis/google-cloud-python/pull/9894
* https://github.com/googleapis/python-kms/issues/2

atulep:
Addresses feature request filed in https://github.com/googleapis/gapic-generator-python/issues/575.

### DIFF
--- a/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
@@ -154,9 +154,10 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
     def __exit__(self, type, value, traceback):
         """Releases underlying transport's resources.
 
-        WARNING: ONLY use as a context manager if the transport is NOT shared
-        with other clients! Exiting the with block will CLOSE the transport
-        and may cause errors in other clients!
+        .. warning::
+            ONLY use as a context manager if the transport is NOT shared
+            with other clients! Exiting the with block will CLOSE the transport
+            and may cause errors in other clients!
         """
         self.transport.close()
 
--- a/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
@@ -148,6 +148,18 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
         """
         return self._transport
 
+    def __enter__(self):
+        self.transport.__enter__()
+        return self
+
+    def __exit__(self, type, value, traceback):
+        """Releases underlying transport's resources.
+
+        WARNING: ONLY use as a context manager if the transport is NOT shared
+        with other clients! Exiting the with block will CLOSE the transport
+        and may cause errors in other clients!
+        """
+        return self.transport.__exit__(type, value, traceback)
 
     {% for message in service.resource_messages|sort(attribute="resource_type") %}
     @staticmethod
--- a/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
@@ -149,7 +149,6 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
         return self._transport
 
     def __enter__(self):
-        self.transport.__enter__()
         return self
 
     def __exit__(self, type, value, traceback):
@@ -159,7 +158,7 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
         with other clients! Exiting the with block will CLOSE the transport
         and may cause errors in other clients!
         """
-        return self.transport.__exit__(type, value, traceback)
+        self.transport.close()
 
     {% for message in service.resource_messages|sort(attribute="resource_type") %}
     @staticmethod
--- a/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/base.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/base.py.j2
@@ -103,10 +103,12 @@ class {{ service.name }}Transport(metaclass=abc.ABCMeta):
             {% endfor %} {# precomputed wrappers loop #}
         }
 
-    def __enter__(self):
-        raise NotImplementedError()
+    def close(self):
+        """Closes resources associated with the transport.
 
-    def __exit__(self, type, value, traceback):
+        WARNING: Only call this method if the transport is NOT shared
+        with other clients - this may cause errors in other clients!
+        """
         raise NotImplementedError()
 
     {% if service.has_lro %}
--- a/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/base.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/base.py.j2
@@ -106,7 +106,7 @@ class {{ service.name }}Transport(metaclass=abc.ABCMeta):
     def close(self):
         """Closes resources associated with the transport.
 
-        :: warning::
+        .. warning::
             Only call this method if the transport is NOT shared
             with other clients - this may cause errors in other clients!
         """
--- a/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/base.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/base.py.j2
@@ -106,8 +106,9 @@ class {{ service.name }}Transport(metaclass=abc.ABCMeta):
     def close(self):
         """Closes resources associated with the transport.
 
-        WARNING: Only call this method if the transport is NOT shared
-        with other clients - this may cause errors in other clients!
+        :: warning::
+            Only call this method if the transport is NOT shared
+            with other clients - this may cause errors in other clients!
         """
         raise NotImplementedError()
 
--- a/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/base.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/base.py.j2
@@ -103,6 +103,11 @@ class {{ service.name }}Transport(metaclass=abc.ABCMeta):
             {% endfor %} {# precomputed wrappers loop #}
         }
 
+    def __enter__(self):
+        raise NotImplementedError()
+
+    def __exit__(self, type, value, traceback):
+        raise NotImplementedError()
 
     {% if service.has_lro %}
 
--- a/gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
+++ b/gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
@@ -654,6 +654,11 @@ def test_{{ service.name|snake_case }}_base_transport():
         with pytest.raises(NotImplementedError):
             getattr(transport, method)(request=object())
 
+    with pytest.raises(NotImplementedError):
+        transport.__enter__()
+    with pytest.raises(NotImplementedError):
+        transport.__exit__(None, None, None)
+
     {% if service.has_lro %}
     # Additionally, the LRO client (a property) should
     # also raise NotImplementedError
@@ -903,5 +908,28 @@ def test_client_withDEFAULT_CLIENT_INFO():
         )
         prep.assert_called_once_with(client_info)
 
+def test_grpc_transport_enter_exit():
+    client = {{ service.client_name }}(
+        credentials=ga_credentials.AnonymousCredentials(),
+        transport='grpc',
+    )
+    with mock.patch.object(type(client.transport._grpc_channel), 'close') as chan_close:
+        with client as _:
+            chan_close.assert_not_called()
+        chan_close.assert_called_once()
+
+def test_grpc_client_enter_exit():
+    client = {{ service.client_name }}(
+        credentials=ga_credentials.AnonymousCredentials(),
+        transport='grpc',
+    )
+    # Test client calls underlying transport.
+    with mock.patch.object(type(client.transport), "__enter__") as enter:
+        with mock.patch.object(type(client.transport), "__exit__") as exit:
+            enter.assert_not_called()
+            exit.assert_not_called()
+            with client as _:
+                enter.assert_called_once()
+            exit.assert_called()
 
 {% endblock %}
--- a/gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
+++ b/gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
@@ -655,9 +655,7 @@ def test_{{ service.name|snake_case }}_base_transport():
             getattr(transport, method)(request=object())
 
     with pytest.raises(NotImplementedError):
-        transport.__enter__()
-    with pytest.raises(NotImplementedError):
-        transport.__exit__(None, None, None)
+        transport.close()
 
     {% if service.has_lro %}
     # Additionally, the LRO client (a property) should
@@ -908,7 +906,7 @@ def test_client_withDEFAULT_CLIENT_INFO():
         )
         prep.assert_called_once_with(client_info)
 
-def test_grpc_transport_enter_exit():
+def test_grpc_transport_close():
     client = {{ service.client_name }}(
         credentials=ga_credentials.AnonymousCredentials(),
         transport='grpc',
@@ -918,18 +916,16 @@ def test_grpc_transport_enter_exit():
             chan_close.assert_not_called()
         chan_close.assert_called_once()
 
-def test_grpc_client_enter_exit():
+def test_grpc_client_ctx():
     client = {{ service.client_name }}(
         credentials=ga_credentials.AnonymousCredentials(),
         transport='grpc',
     )
     # Test client calls underlying transport.
-    with mock.patch.object(type(client.transport), "__enter__") as enter:
-        with mock.patch.object(type(client.transport), "__exit__") as exit:
-            enter.assert_not_called()
-            exit.assert_not_called()
-            with client as _:
-                enter.assert_called_once()
-            exit.assert_called()
+    with mock.patch.object(type(client.transport), "close") as close:
+        close.assert_not_called()
+        with client as _:
+            pass
+        close.assert_called()
 
 {% endblock %}
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
@@ -599,6 +599,12 @@ class {{ service.async_client_name }}:
         return response
     {% endif %}
 
+    async def __aenter__(self):
+        return self
+
+    async def __aexit__(self, exc_type, exc, tb):
+        await self.transport.close()
+
 try:
     DEFAULT_CLIENT_INFO = gapic_v1.client_info.ClientInfo(
         gapic_version=pkg_resources.get_distribution(
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
@@ -481,7 +481,6 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
     {% endfor %}
 
     def __enter__(self):
-        self.transport.__enter__()
         return self
 
     def __exit__(self, type, value, traceback):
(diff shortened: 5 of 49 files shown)


gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -480,6 +480,17 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
         {{ "\n" }}
     {% endfor %}
 
+    def __enter__(self):
+        self.transport.__enter__()
+        return self
+
+    def __exit__(self, type, value, traceback):
```
software-dov, line 487: I think we need a stronger warning, something on the lines of "ONLY use as a context manager if the transport is NOT shared with other clients! Exiting the with block will CLOSE the transport and may cause errors in other clients!"

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
```diff
@@ -224,6 +224,15 @@ class {{ service.name }}GrpcTransport({{ service.name }}Transport):
...
         """
         return self._grpc_channel
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, type, value, traceback):
+        """Closes underlying gRPC channel.
+        """
+        self.grpc_channel.close()
```
software-dov, line 234: These are the things we should make generated unit tests for as well.

software-dov (commented):
Looks good, just some nits around testing.

I think it would be good to have some autogenerated unit tests as well so we can keep generated code coverage at 100%.
Maybe something that injects a dummy transport that can tell us that it was told to enter, something like (strawman expressions in the template, use the correct ones :P )
```python
class DummyTransport:
    def __init__(self):
        self.trigger_state = 0
        
    def __enter__(self):
        self.trigger_state = 1
        
    def __exit__(self, *args, **kwargs):
        self.trigger_state = 2
        
client = {{ api.module_name }}.{{ service.client_name }}(transport=DummyTransport())
assert c.transport.trigger_state == 0
with client:
    assert c.transport.trigger_state == 1
    
assert c.transport.trigger_state == 2
```

And then associated mox tests for the grpc and rest transport enter/exit methods (I always have to steal mox setup from existing code).

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
atulep: Done.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
atulep: Done.

atulep:
Looks like golden files need to be updated. I can't get `bazel run :update` on my Mac due to some weird error related to wrong version of Python. I'm investigating it now.

gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -903,5 +908,28 @@ def test_client_withDEFAULT_CLIENT_INFO():
         )
         prep.assert_called_once_with(client_info)
 
+def test_grpc_transport_enter_exit():
+    client = {{ service.client_name }}(
+        credentials=ga_credentials.AnonymousCredentials(),
+        transport='grpc',
+    )
+    with mock.patch.object(type(client.transport._grpc_channel), 'close') as chan_close:
+        with client as _:
```
software-dov, line 917: Nit: if we don't need the result, we can just ignore it from the with statement, like so:
```python
with client:
    ...
```

gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -903,5 +908,28 @@ def test_client_withDEFAULT_CLIENT_INFO():
...
+    client = {{ service.client_name }}(
+        credentials=ga_credentials.AnonymousCredentials(),
+        transport='grpc',
+    )
+    # Test client calls underlying transport.
+    with mock.patch.object(type(client.transport), "__enter__") as enter:
+        with mock.patch.object(type(client.transport), "__exit__") as exit:
+            enter.assert_not_called()
+            exit.assert_not_called()
+            with client as _:
```
software-dov, line 931: Same here: don't need the `as _`.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -2459,4 +2464,56 @@ async def test_test_iam_permissions_from_dict_async():
...
+            with client as _:
+                enter.assert_called_once()
+            exit.assert_called()
+{% endif %}
+
+{% if 'rest' in opts.transport %}
+def test_rest_transport_enter_exit():
+    client = {{ service.client_name }}(
+        credentials=ga_credentials.AnonymousCredentials(),
+        transport='rest',
```
software-dov, line 2497: I bet we can generalize this like so:

```python
def test_transport_ctx():
    transports = [
        {% if 'rest' in opts.transport %}
        'rest',
        {% endif %}
        {% if 'grpc' in opt.transport %}
        'grpc',
        {% endif %}
    ]
    for transport in transports:
        client = {{ service.client_name }})
            transport=transport
        )
```

gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -654,6 +654,11 @@ def test_{{ service.name|snake_case }}_base_transport():
         with pytest.raises(NotImplementedError):
             getattr(transport, method)(request=object())
 
+    with pytest.raises(NotImplementedError):
+        transport.__enter__()
```
software-dov, line 658: Nit: please add a newline after.

software-dov (commented):
Looks good, just a few more obnoxious nitpicking comments :)

gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
atulep: Thank you, done.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -2459,4 +2465,51 @@ async def test_test_iam_permissions_from_dict_async():
...
+        client = {{ service.client_name }}(
+            credentials=ga_credentials.AnonymousCredentials(),
+            transport=transport
+        )
+        {% if 'rest' in opts.transport %}
+        with mock.patch.object(type(client.transport._session), 'close') as close:
+        {% endif %}
+        {% if 'grpc' in opts.transport %}
+        with mock.patch.object(type(client.transport._grpc_channel), 'close') as close:
+        {% endif %}
```
software-dov, line 2487: My bad, I may have screwed this up in my endeavor to generalize the test. I don't think this is quite right now; we want to patch the object based on what type it is, not just on what the valid transports are. Both can be valid during the same generator run. We can probably fix it like so:
```python
transports = {
    {% conditions, don't want to retype %}
    "rest": "_session",
    {% conditions, don't want to retype %}
    "grpc": "_grpc_channel",
}

for transport, close_name in transports.items():
    client  = ...
    with mock.patch.object(type(getattr(client.transport, close_name)), "close") as close:
        ...
```

But that feels like it may be trying to be too clever.
One alternative is to go back to the previous change, and just duplicate the majority of the template, but I'm not hugely happy with that either. What do you think?

atulep: I like using `dict` here since it doesn't tangle test function code (unlike the `array` approach), and also there is less redundant code (from two different tests).

busunkim96 (approved):
Thank you for implementing this!

CC @tswast @nicain I think both of you were interested in this feature, PTAL if you have a moment.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -480,6 +480,19 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
         {{ "\n" }}
     {% endfor %}
 
+    def __enter__(self):
+        self.transport.__enter__()
+        return self
```
tseaver, line 485: I'm not sure why we expect `transport` to behave with context manager semantics, but yet cannot use them so.  Don't we really just want a `close` method on the transport, to be called in our `__exit__`?

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
```diff
@@ -224,6 +224,15 @@ class {{ service.name }}GrpcTransport({{ service.name }}Transport):
         """Return the channel designed to connect to this service.
         """
         return self._grpc_channel
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, type, value, traceback):
+        """Closes underlying gRPC channel.
```
tswast, line 232: Missing the same warning as the one in `ads` templates?

gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
```diff
@@ -148,6 +148,18 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
         return self._transport
 
+    def __enter__(self):
+        self.transport.__enter__()
+        return self
+
+    def __exit__(self, type, value, traceback):
+        """Releases underlying transport's resources.
+
+        WARNING: ONLY use as a context manager if the transport is NOT shared
```
tswast, line 158: Will this appear anywhere a developer can see it in the documentation? An explicit `close()` method could be useful, even if it's sole purpose is to document that you shouldn't close something where you don't want the underlying transport to be closed.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
atulep: What do you mean by "but yet cannot use them so"? Nothing stops developers from using transport as context manager. 

I am unsure if we want `transport` to be used a context manager. Intuitively, it does not need to be. As long, each `transport` knows how to `close` we would get the desired functionality (i.e. client = context manager) and also perhaps have a cleaner interface. 

@software-dov any thoughts?

gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
atulep: I don't believe we currently generate docs for this, so this would only show up in `doctsring`. I don't like `close` since it assumes that the object needs to be `opened`, which doesn't make sense for a client. Having both `__exit__` and `close` would be redundant too.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
tseaver: Having one context manager call another object's `__enter__` and `__exit__` does not feel like idiomatic Python usage:  as with other "dunder" methods, such "protocol helpers" are not intended to be called directly.

atulep: I was also wondering about this, but after reading some posts on [SO](https://stackoverflow.com/questions/26635684/calling-enter-and-exit-manually), I came to conclusion it's not a bad thing.

gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
busunkim96: @tswast Is the concern that `__exit__` won't show up in the generated refdocs?

tswast: That is my primary concern, yes.

tswast: I disagree that one needs `open()` for there to be a `close()`, but I would expect most people to use `__exit__` if available. There is utility in having an explicit `close()` method, but my main concern is the discoverability in the docs.

tswast: `close()` would allow some flexibility that just a context manager doesn't allow. For example, see the PR some customers have linked to this issue: https://github.com/coiled/dask-bigquery/pull/4#discussion_r710418038

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
atulep: Thank you for your suggestions. I'm going to replace `__enter__`, `__exit__` in `transport` with `close`. I'll leave `close` from `client` at this moment, but we can consider adding it in the future. 

@tseaver @tswast @software-dov @busunkim96

gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
atulep: Let's keep `close()` from client for now. If this problem arises, we can follow-up in separate CL.

atulep:
@lidizheng PTAL async client

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
```diff
@@ -599,6 +599,12 @@ class {{ service.async_client_name }}:
         return response
     {% endif %}
 
+    async def __aenter__(self):
+        return self
+
+    async def __aexit__(self, exc_type, exc, tb):
+        await self.transport.close()
```
lidizheng, line 606: Are we certain the `transport` will exist at the time of exiting? Same for other `aexits`.

lidizheng (commented):
The added context manager method looks good, just one minor comment.

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
atulep: Yes, because it gets created in `__init__`. @software-dov correct me if Im wrong.

software-dov: Yes, it should. The client will construct a transport in `__init__` if one isn't passed in. The user is of course free to set the field manually or create a client that breaks assumptions, but this should be safe.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
atulep: Obsolete, since `__exit__` was removed.

atulep:
@software-dov @tseaver @tswast @busunkim96 PTAL. I've addressed all your comments.

gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
tswast: @atulep We still need to make sure this warning appears somewhere visible in the docs. When you generate the docs for one of these clients, do you see `__exit__` in the documentation?

atulep: It shows up in generated reference. I also added an appropriate infobox around warning to make it stand out.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
```diff
@@ -183,6 +183,14 @@ class {{ service.name }}Transport(abc.ABC):
             {% endfor %} {# precomputed wrappers loop #}
         }
 
+    def close(self):
+        """Closes resources associated with the transport.
+
+       :: warning::
```
tswast, line 189: I think this should have the same syntax as the other warning infoboxes, right?

```suggestion
       .. warning::
```

gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/base.py.j2
```diff
@@ -103,6 +103,14 @@ class {{ service.name }}Transport(metaclass=abc.ABCMeta):
             {% endfor %} {# precomputed wrappers loop #}
         }
 
+    def close(self):
+        """Closes resources associated with the transport.
+
+        :: warning::
```
tswast, line 109:
```suggestion
        .. warning::
```

tests/integration/goldens/asset/google/cloud/asset_v1/services/asset_service/transports/base.py
```diff
@@ -259,6 +259,15 @@ def _prep_wrapped_messages(self, client_info):
             ),
          }
 
+    def close(self):
+        """Closes resources associated with the transport.
+
+       :: warning::
```
tswast, line 265:
```suggestion
       .. warning::
```

tests/integration/goldens/credentials/google/iam/credentials_v1/services/iam_credentials/transports/base.py
```diff
@@ -196,6 +196,15 @@ def _prep_wrapped_messages(self, client_info):
             ),
          }
 
+    def close(self):
+        """Closes resources associated with the transport.
+
+       :: warning::
```
tswast, line 202:
```suggestion
       .. warning::
```

tests/integration/goldens/logging/google/cloud/logging_v2/services/logging_service_v2/transports/base.py
```diff
@@ -231,6 +231,15 @@ def _prep_wrapped_messages(self, client_info):
             ),
          }
 
+    def close(self):
+        """Closes resources associated with the transport.
+
+       :: warning::
```
tswast, line 237:
```suggestion
       .. warning::
```

tests/integration/goldens/logging/google/cloud/logging_v2/services/metrics_service_v2/transports/base.py
```diff
@@ -210,6 +210,15 @@ def _prep_wrapped_messages(self, client_info):
             ),
          }
 
+    def close(self):
+        """Closes resources associated with the transport.
+
+       :: warning::
```
tswast, line 216:
```suggestion
       .. warning::
```

tests/integration/goldens/redis/google/cloud/redis_v1/services/cloud_redis/transports/base.py
```diff
@@ -195,6 +195,15 @@ def _prep_wrapped_messages(self, client_info):
             ),
          }
 
+    def close(self):
+        """Closes resources associated with the transport.
+
+       :: warning::
```
tswast, line 201:
```suggestion
       .. warning::
```

tests/integration/goldens/logging/google/cloud/logging_v2/services/config_service_v2/transports/base.py
```diff
@@ -323,6 +323,15 @@ def _prep_wrapped_messages(self, client_info):
             ),
          }
 
+    def close(self):
+        """Closes resources associated with the transport.
+
+       :: warning::
```
tswast, line 329:
```suggestion
       .. warning::
```

software-dov (approved):

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
atulep: Which one are you talking about? 

I am following syntax from the [docs](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html).

tseaver: Per the docs you linked, Restructured Text  directives like `note`, `warning`, `code-block` all start with two periods, then a space, then the directive name, then two colons.

atulep: Ah, thank you. I see now.

PR #988, opened 2021-09-09
fix: add a separate DEFAULT_CLIENT_INFO for rest clients

jameslynnwu:
Previously, DEFAULT_CLIENT_INFO would be used for rest transports. However, if you generated both rest and grpc transports, rest version information would appear in both transport's client_infos. Now, we create our own DEFAULT_CLIENT_INFO for rest clients and copy the common information from the base DEFAULT_CLIENT_INFO.


tseaver (approved):

vam-google (approved):
LGTM

PR #776, opened 2021-02-13
feat: add grpc transcoding, clean up unimplemented rest method/test generation

yon-mg:
This pr introduces usage of grpc transcoding and is reliant on changes made to to,
https://github.com/googleapis/python-api-core/pull/146.

With this implementation of grpc transcoding, there are a few more changes that need to be made to ensure Showcase tests can run correctly. LRO rpcs, streaming rpcs and rpcs without `google.api.http` annotations need to also be handled. In this pr they are not implemented fully and instead rather raise errors if used in a generated client. The intention is to implement them at a later time and remove them as temporary blockers to push the grpc transcoding changes.

PR #993, opened 2021-09-14
feat: Support alternative http bindings in the gapic schema.

kbandes:
This completes part of yon-mg's PR, https://github.com/googleapis/gapic-generator-python/pull/776.  The template changes will be in a separate PR.

### DIFF
--- a/gapic/schema/wrappers.py
+++ b/gapic/schema/wrappers.py
@@ -823,27 +823,27 @@ class Method:
         return next((tuple(pattern.findall(verb)) for verb in potential_verbs if verb), ())
 
     def http_rule_to_tuple(self, http_rule: http_pb2.HttpRule) -> Tuple[Optional[str], Optional[str], Optional[str]]:
-      """Represent salient info in an http rule as a tuple.
+        """Represent salient info in an http rule as a tuple.
 
-      Args:
-        http_rule: the http option message to examine.
+        Args:
+          http_rule: the http option message to examine.
 
-      Returns:
-        A tuple of (method, uri pattern, body or None),
-          or None if no method is specified.
-      """
+        Returns:
+          A tuple of (method, uri pattern, body or None),
+            or None if no method is specified.
+        """
 
-      http_dict: Mapping[str, str]
+        http_dict: Mapping[str, str]
 
-      method = http_rule.WhichOneof('pattern')
-      if method is None or method == 'custom':
-        return (None, None, None)
+        method = http_rule.WhichOneof('pattern')
+        if method is None or method == 'custom':
+            return (None, None, None)
 
-      uri = getattr(http_rule, method)
-      if not uri:
-        return (None, None, None)
-      body = http_rule.body if http_rule.body else None
-      return (method, uri, body)
+        uri = getattr(http_rule, method)
+        if not uri:
+            return (None, None, None)
+        body = http_rule.body if http_rule.body else None
+        return (method, uri, body)
 
     @property
     def http_options(self) -> List[Dict[str, str]]:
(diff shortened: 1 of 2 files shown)


gapic/schema/wrappers.py
```diff
@@ -821,13 +822,64 @@ def field_headers(self) -> Sequence[str]:
 
         return next((tuple(pattern.findall(verb)) for verb in potential_verbs if verb), ())
 
+    def http_rule_to_tuple(self, http_rule: http_pb2.HttpRule) -> Tuple[str, str, str]:
```
software-dov, line 825: Minor digression: I think it would be more helpful to return either a dataclass or a named tuple. I'd prefer to make method, uri, body named fields instead of positional convention.

gapic/schema/wrappers.py
```diff
@@ -821,13 +822,64 @@ def field_headers(self) -> Sequence[str]:
...
+        http_dict: Mapping[str, str]
+
+        method = http_rule.WhichOneof('pattern')
+        if method is None or method == 'custom':
+            return ('', '', '')
+
+        uri = getattr(http_rule, method)
+        if not uri:
+            return ('', '', '')
+        body = http_rule.body if http_rule.body else None
```
software-dov, line 845: Outrageous nit, feel free to ignore: can simplify this into
```python
body = http_rule.body or None
```

gapic/schema/wrappers.py
```diff
@@ -821,13 +822,64 @@ def field_headers(self) -> Sequence[str]:
...
+        e.g. [{'method': 'post'
+              'uri': '/some/path'
+              'body': '*'},]
+
+        """
+        http = self.options.Extensions[annotations_pb2.http]
+        # shallow copy is fine here (elements are not modified)
+        http_options = list(http.additional_bindings)
+        # Main pattern comes first
+        http_options.insert(0, http)
```
software-dov, line 861: Minor optimization, feel free to ignore:
```python
http_options = [http]
http_options.extend(http.additional_bindings)
```

gapic/schema/wrappers.py
```diff
@@ -821,13 +822,64 @@ def field_headers(self) -> Sequence[str]:
...
+        for http_rule in http_options:
+            method, uri, body = self.http_rule_to_tuple(http_rule)
+            if not method:
+                continue
+            answer: Dict[str, str] = {}
+            answer['method'] = method
+            answer['uri'] = uri
+            if body:
+                answer['body'] = body
+            answers.append(answer)
```
software-dov, line 873: Since we're going to be testing the answers for 'body' anyway, let's commit to it with a named structure.

```python
@dataclasses.dataclass(frozen=True)
class HttpRule:
    method: str
    uri: str
    body: Optional[str]
    
    @classmethod
    def try_parse_http_rule(http_rule) -> Optional[HttpRule]:
        method = http_rule.WhichOneof("pattern")
        if method is None or method == "custom":
            return None
            
        uri = getattr(http_rule, method)
        if not uri:
            return None
            
        body = http_rule.body or None
        return cls(method, uri, body)
        

# class Method
    @property
    def http_options(self) -> List[HttpRule]
        http = self.options.Extensions[annotations_pb2.http]
        http_options = [http] + http.additional_bindings
        opt_gen = (HttpRule.try_parse_http_rule(http_rule) for http_rule in http_options)
        return [rule for rule in opt_gen if rule]
```

software-dov (commented):
Question: are these additions intended to replace or complement the existing `http_opt` property? I think we should probably have only one. The names are too similar to be easily distinguished without reading the code.

gapic/schema/wrappers.py
kbandes: Done - rewrote as dataclass.

gapic/schema/wrappers.py
kbandes: Done.

gapic/schema/wrappers.py
kbandes: Superseded by comment below.

gapic/schema/wrappers.py
kbandes: Done with minor tweaks.  e.g., I had to do 
`def http_options(self) -> List['HttpRule']`
because the HttpRule class is not defined at the point of reference.
Also,
`http_options = [http] + http.additional_bindings`
became
`http_options = [http] + list(http.additional_bindings)`
because additional_bindings is iterable but not explicitly a list and can't be added to a list.
(I could have used `.extend()` but I like this better.)

gapic/schema/wrappers.py
```diff
@@ -821,13 +843,22 @@ def field_headers(self) -> Sequence[str]:
...
+    def http_options(self) -> List[HttpRule]:
+        """Return a list of the http bindings for this method."""
+        http = self.options.Extensions[annotations_pb2.http]
+        http_options = [http] + list(http.additional_bindings)
+        opt_gen = (HttpRule.try_parse_http_rule(http_rule)
+                   for http_rule in http_options)
+        return [rule for rule in opt_gen if rule]
+
     @property
     def http_opt(self) -> Optional[Dict[str, str]]:
```
kbandes, line 856: I agree that this property should be removed - I would prefer to do this in a follow-up PR, after fixing the rest template that currently relies on it.

software-dov: Fine by me

gapic/schema/wrappers.py
software-dov: Sounds good. Glad my proposal worked out.

software-dov (approved):

PR #997, opened 2021-09-17
fix: re-create gRPC channel on 'socket operation on non-socket' error

busunkim96:
Under some specific conditions in Cloud Functions, the original gRPC channel will become unusable and the method call will return `google.api_core.exceptions.ServiceUnavailable: 503 Socket operation on non-socket`.

This PR:

- Adds a function `reinitialize_grpc_channel` that re-creates the gRPC channel with the original arguments
- Automatically calls `reinitialize_grpc_channel` before retrying if the exception is `google.api_core.exceptions.ServiceUnavailable: 503 Socket operation on non-socket` https://googleapis.dev/python/google-api-core/latest/retry.html

**Note:** the automatic grpc channel creation before retry is only being added to methods _with an existing retry policy_. This is not all methods, since most gRPC configs have methods that [only specify a deadline](https://github.com/googleapis/googleapis/blob/master/google/privacy/dlp/v2/dlp_grpc_service_config.json). I believe omitting methods with no retry policy is necessary to comply with https://google.aip.dev/client-libraries/4221 and https://google.aip.dev/194. Please let me know if my interpretation is incorrect! I've left `reintialize_grpc_channel` as a public method so users can call it in their own retry objects.

Googlers see b/199551791.

CC @lidizheng

### DIFF
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2
@@ -148,6 +148,12 @@ class {{ service.name }}Transport(abc.ABC):
 
         return scopes_kwargs
 
+    def _refresh_transport(self, exc: Exception) -> None:
+        """If the exception indicates the connection has been lost,
+        this function attempts to re-establish it.
+        Intended to be passed to ``on_error`` on ``google.api_core.retry.Retry``.
+        """
+        pass
 
     def _prep_wrapped_messages(self, client_info):
         # Precompute the wrapped methods.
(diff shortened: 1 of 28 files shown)


lidizheng:
In an offline discussion with @srini100, we found there might be a risk for retry like this.

We are fixing the Cloud Function scenario where the socket broke due to process inactive, so no traffic is sent to the wire. However, there could be other cases causing the same status code and status message. E.g., the socket broke during client receiving message, then the outbound message is sent, if we retry, there could be duplicated requests.

Core is working hard on a [transparent retry](https://github.com/grpc/proposal/blob/master/A6-client-retries.md#transparent-retries) feature (aiming Q4 2021) that automatically retry/reconnect if it's safe to do so.

busunkim96:
@lidizheng Thanks for the detailed explanation! I'm excited for transparent retry, is there a doc or bug I can subscribe to?

It also looks like this specific scenario might be avoidable with changes to the functions framework, so I've decided to close this. https://github.com/GoogleCloudPlatform/functions-framework-python/pull/147

PR #1017, opened 2021-10-09
Update setup.py

anukaal:
updated classifiers


tseaver:
@anukaal Thanks for the patch!  It prompted me to open #1019, which also adds Python 3.10 to the testing and templated noxfile.

PR #1020, opened 2021-10-12
chore: add tests for python 3.10

software-dov:


software-dov:
I took a deep dive through the grpc async code, and it looks like we will need to wait for a new python release of grpc (that we will then need to pin) to fix the root cause. The fix is trivial, and it looks like it's been made, but the change is not in the latest official release.
Leaving this PR open and will check back periodically to see if a new release that fixes the issue has been made.

anukaal (approved):

parthea (approved):

PR #1034, opened 2021-10-15
fix: fix test sample generation logic for null uri templates

vam-google:
Previously code supported `complex/{uris=*}/like/{this=**}`
With fix it also supports `complex/{uris}/like/{this}` uris.


vam-google:
@kbandes FYI

tseaver (commented):
I know this PR didn't add it this way, but can we please move the regex out to a module scope variable (using `re.compile` and use the "verbose" form (see [docs](https://docs.python.org/3/library/re.html#re.VERBOSE)) to annotate it?  Reasoning about complicated regexes is much easier if they are well documented.

software-dov (commented):
Looks good, just some minor nits

PR #1044, opened 2021-10-27
feat: generate code snippets by default

busunkim96:


gapic/utils/options.py
```diff
@@ -132,6 +132,19 @@ def tweak_path(p):
         # Build the options instance.
         sample_paths = opts.pop('samples', [])
 
+        # autogen-snippets is True by default, so make sure users can disable
+        # by passing `autogen-snippets=false`
+        if opts.get("autogen-snippets") and opts.pop("autogen-snippets")[0] in ("False", "false"):
+            autogen_snippets = False
+        else:
+            autogen_snippets = True
```
software-dov, line 140: `dict.pop` takes a default parameter, like so:
```python
autogen_snippets = opts.pop("autogen-snippets", ["True"])[0] in ("True", "true", "T", "t", "TRUE")
```

software-dov (approved):
LGTM. It would be good to have a test for options,Options for default behavior and disabling based on old-naming, but I'm not going to block merge on that.

gapic/utils/options.py
busunkim96: Done.

PR #1047, opened 2021-10-28
fix: methods returning Operation w/o operation_info are now allowed.

kbandes:
Formerly, a method that returned google.longrunning.Operation but did not have the google.longrunning.operation_info extension was considered an error; but there are valid cases for this, particularly in the Operation service itself.
This change makes it acceptable to have a method like this.  It is not treated as a long-running operation.


software-dov (approved):
LGTM, just outrageously picky grammar nits in new comments.

PR #1049, opened 2021-10-29
fix: fix tests generation logic

vam-google:
This includes:
1) Fix test logic for grpc+rest case, when clients with both transports need to be initialized in parametrized tests
2) Fix 100% coverage problem for rest clients, when the http error (>= 400 error code) case logic was not covered.

This should conclude the rest transport related fixes (for the time being at least)


kbandes (approved):

busunkim96 (approved):
One non-blocking question, otherwise LGTM.

software-dov (approved):
LGTM. Just one question: how the hell did you get around the transport `__call__` patch problem? I had to do this massive rewrite of the rest layer, but you seem to have avoided it.

PR #1055, opened 2021-10-29
test: add tests for autogen-snippets option

busunkim96:
Follow up to #1044

### DIFF
--- a/tests/unit/generator/test_options.py
+++ b/tests/unit/generator/test_options.py
@@ -141,9 +141,7 @@ def test_options_service_config(fs):
 
 
 def test_options_bool_flags():
-    # All these options are default False.
-    # If new options violate this assumption,
-    # this test may need to be tweaked.
+    # Most options are default False.
     # New options should follow the dash-case/snake_case convention.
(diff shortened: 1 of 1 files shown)


tests/unit/generator/test_options.py
```diff
@@ -161,3 +159,18 @@ def test_options_bool_flags():
...
+    options = Options.build("autogen-snippets=False")
+    assert not options.autogen_snippets
+
+
+def test_options_autogen_snippets_false_for_old_naming():
+    # NOTE: Snippets are not currently correct for the alternative (Ads) templates
+    # so always disable snippetgen in that case
+    # https://github.com/googleapis/gapic-generator-python/issues/1052
+    options = Options.build("old-naming")
+    assert not options.autogen_snippets
```
software-dov, line 176: I think, just for the sake of completeness, it would be good to have a test where both `autogen-snippets=True` and `old-naming` are set to make sure that the builder logic handles that correctly.

software-dov (approved):
LGTM, one minor test case I think should be added, but not worth blocking the review over.

PR #1056, opened 2021-10-29
feat: add fragment tests

### LINKED ISSUE #413: Method signature elements that are repeated fields are not handled properly

### LINKED ISSUE #414: google.protobuf.Value field in request message breaks corresponding generated flattened method test
The following:
```proto
import "google/protobuf/struct.proto";
import "google/api/client.proto";

service MyService {
  option (google.api.default_host) = "my.example.com";

  rpc MyMethod(MethodRequest) returns (MethodResponse) {
  	/* option (google.api.method_signature) = "inputs,parameters"; */
    option (google.api.method_signature) = "parameter";
  }
}

message MethodRequest {
  google.protobuf.Value parameter = 1;
}
```

Yields the following error in the `test_my_flattened_method` generated unit test:
```
 assert args[0].parameter == struct.Value(null_value=struct.NullValue.NULL_VALUE)
E           AssertionError: assert None == null_value: NULL_VALUE\n
E            +  where None = parameters {\n  null_value: NULL_VALUE\n}\n.parameters
E            +  and   null_value: NULL_VALUE\n = <class 'google.protobuf.struct_pb2.Value'>(null_value=0)
E            +    where <class 'google.protobuf.struct_pb2.Value'> = struct.Value
E            +    and   0 = <google.protobuf.internal.enum_type_wrapper.EnumTypeWrapper object at 0x7f85c7f3059
0>.NULL_VALUE                                                                                                 
E            +      where <google.protobuf.internal.enum_type_wrapper.EnumTypeWrapper object at 0x7f85c7f30590>
 = struct.NullValue           
```

software-dov:
Fragment tests are defined by a small proto file describing an API
surface with characteristics such that it is desirable to test the
generated surface for correctness or to prevent regressions.

As part of a fragment test, the generator is run on a fragment to
create a GAPIC library for the fragment. The generated unit tests
for the fragment are then executed to test the surface.

Includes fragment tests and fixes for #413 and #414 as well as fragment tests for various other previously fixed issues.

Note: the parallelization of the fragment tests is currently a standalone commit and can be removed. It dramatically speeds up execution, however, since most of the time spent per fragment is waiting on I/O or subprocesses.

### DIFF
--- a/tests/fragments/google/api/client.proto
+++ b/tests/fragments/google/api/client.proto
@@ -1,0 +1,99 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package google.api;
+
+import "google/protobuf/descriptor.proto";
+
+option go_package = "google.golang.org/genproto/googleapis/api/annotations;annotations";
+option java_multiple_files = true;
+option java_outer_classname = "ClientProto";
+option java_package = "com.google.api";
+option objc_class_prefix = "GAPI";
+
+extend google.protobuf.MethodOptions {
+  // A definition of a client library method signature.
+  //
+  // In client libraries, each proto RPC corresponds to one or more methods
+  // which the end user is able to call, and calls the underlying RPC.
+  // Normally, this method receives a single argument (a struct or instance
+  // corresponding to the RPC request object). Defining this field will
+  // add one or more overloads providing flattened or simpler method signatures
+  // in some languages.
+  //
+  // The fields on the method signature are provided as a comma-separated
+  // string.
+  //
+  // For example, the proto RPC and annotation:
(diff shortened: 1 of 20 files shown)


tests/fragments/google/api/client.proto
```diff
@@ -0,0 +1,99 @@
...
+
+import "google/protobuf/descriptor.proto";
+
+option go_package = "google.golang.org/genproto/googleapis/api/annotations;annotations";
+option java_multiple_files = true;
+option java_outer_classname = "ClientProto";
+option java_package = "com.google.api";
+option objc_class_prefix = "GAPI";
+
+extend google.protobuf.MethodOptions {
```
busunkim96, line 27: Where are these proto files from?

busunkim96 (approved):
The golden files seem to need updating.

tests/fragments/google/api/client.proto
software-dov: I copied them from googleapis and protobuf. My original idea was something complicated involving a hand-rolled semaphore/caching system to pull those repos as dependencies before running the fragment tests. I realized that was way too complicated, however, and just copied the three files I needed for these fragments.

PR #1058, opened 2021-11-01
chore: add bazel caching to integration tests

software-dov:
The integration test CI takes ~8 minutes to run, most of which is
installing bazel and setting up tool dependencies.

It is a massive time optimization to cache this work and prevent its
recreation on subsequent builds.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -321,32 +321,35 @@ jobs:
         run: nox -s unit-${{ matrix.python }}
   integration:
     runs-on: ubuntu-latest
+    container: gcr.io/gapic-images/googleapis-bazel:20210105
     steps:
       - name: Cancel Previous Runs
         uses: styfle/cancel-workflow-action@0.7.0
         with:
           access_token: ${{ github.token }}
       - uses: actions/checkout@v2
-      - name: Set up Python 3.8
-        uses: actions/setup-python@v2
+      - name: Cache Bazel files
+        id: cache-bazel
+        uses: actions/cache@v2
         with:
-          python-version: 3.8
-      - name: Install system dependencies.
+          path: ~/.cache/bazel
+          key: ${{ runner.os }}-bazel-20210105-${{ secrets.CACHE_VERSION }}
+      - name: Cache not found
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -312,32 +312,35 @@ jobs:
...
-      - name: Install system dependencies.
+          path: ~/.cache/bazel
+          key: ${{ runner.os }}-bazel-20210105-${{ secrets.CACHE_VERSION }}
+      - name: Cache not found
+        if: steps.cache-bazel.outputs.cache-hit != 'true'
         run: |
-          sudo apt-get update
-          sudo apt-get install -y curl pandoc unzip gcc
-      - name: Install Bazel
+          echo "No cache found."
```
tseaver, line 331: Shouldn't a cache miss imply that we actually do something to populate the missed files?

software-dov: This is done implicitly by bazel during startup. It checks for the existence of `~/.cache/bazel` and if not found runs all the tedious startup of compiling `gRPC`, `protoc`, etc. which are then put into the cache.
The github action saves this directory persistently. At the start of a CI run, it checks for that directory in persistent storage and, if found, populates it in the container.

software-dov: Anecdotally, this cuts the integration test down to about 1 minute 30 seconds.

.github/workflows/tests.yaml
```diff
@@ -312,32 +312,35 @@ jobs:
...
-      - name: Set up Python 3.8
-        uses: actions/setup-python@v2
+      - name: Cache Bazel files
+        id: cache-bazel
+        uses: actions/cache@v2
         with:
-          python-version: 3.8
-      - name: Install system dependencies.
+          path: ~/.cache/bazel
+          key: ${{ runner.os }}-bazel-20210105-${{ secrets.CACHE_VERSION }}
```
busunkim96, line 327: Does the '20210105' bit ever need to be updated? If so, please explain in a comment.

busunkim96 (approved):
LGTM (with the caveat that I don't know much about Bazel 👀 )

Issue #1043, opened 2021-10-27
Turn on snippetgen by default


busunkim96:
CC @dizcology

There's a small chance snippetgen falls apart on a real API, so I plan to take the following steps to avoid blocking other updates:

- Merge the open release PR: #1037 
- Turn on snippetgen
- Release the new release PR

busunkim96:
There appears to be a bug for client streaming so I wasn't able to enable this globally:

If it takes a while to fix I will back out #1044 so it doesn't block other changes.

```py
    sample["request"] = v.validate_and_transform_request(
  File "/usr/local/google/home/busunkim/.cache/bazel/_bazel_busunkim/5f7ad274936c83fa5125630e373edd54/sandbox/linux-sandbox/1362/execroot/com_google_googleapis/bazel-out/host/bin/external/gapic_generator_python/gapic_plugin.runfiles/gapic_generator_python/gapic/samplegen/samplegen.py", line 560, in validate_and_transform_request
    raise types.InvalidRequestSetup(
gapic.samplegen_utils.types.InvalidRequestSetup: Too many base parameters for client side streaming form
--python_gapic_out: protoc-gen-python_gapic: Plugin failed with status code 1.
Target //google/firestore/v1:firestore-v1-py failed to build
```

busunkim96:
Aand I've found at least one other bug - it was not caught by the goldens but was discovered when I tried to use #1077 as the generator version in the WORKSPACE file. I think it is related to https://github.com/googleapis/google-auth-library-python/pull/701. 

I'm planning on manually generating locally with 0.56.1 with every API to shake out remaining bugs blocking turning on snippetgen by default.

busunkim96:
#1012 started to use the resource path logic which wasn't written to handle nested sub-fields. The generation was failing as matching resource paths were not found for resource paths that are not on the base request.

busunkim96:
I've opened #1144 to track re-adding f-string support for resource paths (it requires addressing the bug mentioned in the last comment).

The resource path logic appears to have been the last blocking issue. I ran the internal googleapis presubmit with #1134 and all the APIs appear to be generate successfully. 😃

parthea:
@busunkim96 Please can you check if there are any outstanding items? If not could we close this issue?

PR #1061, opened 2021-11-02
fix(snippetgen): fix client streaming samples

### LINKED ISSUE #1014: snippetgen: streaming methods have syntax errors
Generated snippets for streaming methods have incorrect syntax:

https://github.com/googleapis/gapic-generator-python/blob/dbfb45ac6a0dc0b14efc895d7c584062b40ff2c0/tests/integration/goldens/logging/samples/generated_samples/logging_generated_logging_v2_logging_service_v2_tail_log_entries_async.py#L36-L44

busunkim96:
Fixes #1014 and unblocks #1043.

NOTE: Some real world APIs expect the first request to pass a config ([example](https://github.com/googleapis/googleapis/blob/1d163315aecf9d8a3b77eb9e1549a0b353103b32/google/cloud/speech/v1/cloud_speech.proto#L101-L119)) so the generated samples will not work out of the box. This will be addressed when the new sample config language is sorted out.

### DIFF
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -551,16 +551,6 @@ class Validator:
                     AttributeRequestSetup(**r_dup)  # type: ignore
                 )
 
-        client_streaming_forms = {
-            types.CallingForm.RequestStreamingClient,
-            types.CallingForm.RequestStreamingBidi,
-        }
-
-        if len(base_param_to_attrs) > 1 and calling_form in client_streaming_forms:
-            raise types.InvalidRequestSetup(
-                "Too many base parameters for client side streaming form"
-            )
-
         # We can only flatten a collection of request parameters if they're a
         # subset of the flattened fields of the method.
         flattenable = self.flattenable_fields >= set(base_param_to_attrs)
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -469,9 +469,7 @@ class Validator:
 
         Raises:
             InvalidRequestSetup: If a dict in the request lacks a "field" key,
-                                 a "value" key, if there is an unexpected keyword,
-                                 or if more than one base parameter is given for
-                                 a client-side streaming calling form.
+                                 a "value" key or if there is an unexpected keyword.
             BadAttributeLookup: If a request field refers to a non-existent field
                                 in the request message type.
             ResourceRequestMismatch: If a request attempts to describe both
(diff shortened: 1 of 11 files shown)


gapic/samplegen/samplegen.py
```diff
@@ -551,16 +551,6 @@ def validate_and_transform_request(
...
-        client_streaming_forms = {
-            types.CallingForm.RequestStreamingClient,
-            types.CallingForm.RequestStreamingBidi,
-        }
-
-        if len(base_param_to_attrs) > 1 and calling_form in client_streaming_forms:
-            raise types.InvalidRequestSetup(
-                "Too many base parameters for client side streaming form"
-            )
-
```
busunkim96, line 563: @software-dov 👋  I'm having trouble understanding this error, would you remember what this was written to prevent?

Some streaming method request messages have 2+ required fields ([pubsub](https://github.com/googleapis/googleapis/blob/bcfd386c224fded5f5c5be7a17bc5884a2a52f59/google/pubsub/v1/pubsub.proto#L1083-L1162)), so the current snippetgen logic adds two `field`s to the dictionary. That causes this error to be raised. Are fields expected to be set differently for streaming?


```py
# Request passed to `validate_and_transform_request`
[
   {
      "field":"subscription",
      "value":"projects/{project}/subscriptions/{subscription}"
   },
   {
      "field":"stream_ack_deadline_seconds",
      "value":2813
   }
]
```

I see two unit tests that raise this error, but I don't see any pre-existing tests showing a valid sample config for a client streaming method.



https://github.com/googleapis/gapic-generator-python/blob/a8d2b3a2c11984d8677e4e6547a0cbb76d9ed915/tests/unit/samplegen/test_samplegen.py#L1353-L1402

software-dov: I really don't remember what was going on here. My _guess_ is that I was mildly incorrect about the actual python method signature convention and I was trying to imply that we only wanted a single request to be put into a list...?

TL;DR: It's probably a bug or at least a misreading of the spec on my part.

busunkim96:
This is ready for review, PTAL @software-dov @parthea.

tests/integration/goldens/logging/samples/generated_samples/logging_generated_logging_v2_logging_service_v2_tail_log_entries_async.py
```diff
@@ -38,8 +38,17 @@ async def sample_tail_log_entries():
         resource_names=['resource_names_value_1', 'resource_names_value_2'],
     )
 
+    # This method expects an iterator which contains
+    # 'logging_v2.TailLogEntriesRequest' objects
+    # Here we create a generator that yields a single `request` for
+    # demonstrative purposes.
+    requests = [request]
```
busunkim96, line 45: One alternative is:

```py
requests = iter([requests])
```
I chose the generator over `iter()` to nudge folks towards generating requests on demand.

Opinions/suggestions on the best way to present client streaming welcome!

parthea (approved):

PR #1068, opened 2021-11-03
fix: handle required fields properly in query_params

kbandes:
A required field with a default value was getting dropped from query_params in the transcoding logic.  This fix ensures that required fields are included, regardless of value.


vam-google (approved):
LGTM

PR #1074, opened 2021-11-05
fix: fix rest unit test

vam-google:
In some cases `hostname:port` format was expected.

### DIFF
--- a/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
+++ b/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
@@ -1293,7 +1293,7 @@ def test_{{ method.name|snake_case }}_rest_flattened(transport: str = 'rest'):
         assert len(req.mock_calls) == 1
         _, args, _ = req.mock_calls[0]
         {% with uri = method.http_options[0].uri %}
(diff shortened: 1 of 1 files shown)


alexander-fenster (approved):

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -1293,7 +1293,7 @@ def test_{{ method.name|snake_case }}_rest_flattened(transport: str = 'rest'):
         assert len(req.mock_calls) == 1
         _, args, _ = req.mock_calls[0]
         {% with uri = method.http_options[0].uri %}
-        assert path_template.validate("https://{{ service.host }}{{ uri }}", args[1])
+        assert path_template.validate("https://%s{{ uri }}" % client.transport._host, args[1])
```
tseaver, line 1296: Do we not have a "golden" example for a REST API?

vam-google: No we do not, but I think we should.

vam-google: No we do not, but I think it is a good idea to have one.

PR #1076, opened 2021-11-08
tests: add system tests to check error details

atulep:
System tests to check error details added in https://github.com/googleapis/python-api-core/pull/286.

### DIFF
--- a/tests/system/test_error_details.py
+++ b/tests/system/test_error_details.py
@@ -20,23 +20,6 @@ from grpc_status import rpc_status
 from google.api_core import exceptions
 
 
-def create_bad_request_details():
-    bad_request_details = error_details_pb2.BadRequest()
-    field_violation = bad_request_details.field_violations.add()
-    field_violation.field = "test field"
-    field_violation.description = "test description"
-    return bad_request_details
-
-
-def create_precondition_failure_details():
-    pf_details = error_details_pb2.PreconditionFailure()
-    violation = pf_details.violations.add()
-    violation.type = "test type"
-    violation.subject = "test subject"
-    violation.description = "test description"
-    return pf_details
-
-
 def create_status(error_details=None):
     status = rpc_status.status_pb2.Status()
     status.code = 3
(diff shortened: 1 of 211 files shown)


atulep:
The commit history in this PR is misleading. It contains some of my older commits unrelated to this PR. Any ideas how to fix it?

tests/system/test_error_details.py
```diff
@@ -0,0 +1,81 @@
...
+
+import pytest
+from google import showcase
+from google.rpc import error_details_pb2
+from google.protobuf import any_pb2
+from grpc_status import rpc_status
+from google.api_core import exceptions
+
+
+def create_bad_request_details():
```
tseaver, line 23: Please just inline this function into its only caller below.

tseaver (commented):
In addition to my comments inline, two things:
- These are unit tests, not system tests (they don't exercise actual over-the-wire API calls).
- They belong in `python-api-core`, not in this repository.

tseaver:
@atulep For future reference, you can rebase your branch against `main` (e.g., `git rebase main`) to clean up the previous history.

atulep:
> In addition to my comments inline, two things:
> 
> * These are unit tests, not system tests (they don't exercise actual over-the-wire API calls).
> * They belong in `python-api-core`, not in this repository.

@tseaver 
Tests send API requests to showcase server and assert errors returned from real server response.

We want gapic clients to have access to `error.details`, so we should test it here. Additionally,`python-api-core` doesn't have any existing infrastructure for system tests, which suggested to me it's not a suitable place.

tseaver:
@atulep I apologize -- I somehow missed seeing that these tests use the `echo` fixture.

Please do rebase your branch against the current `main` and force push, e.g.:

```bash
$ git fetch --all --prune
$ git checkout main
$ git checkout error_details
$ git rebase main
$ git push -f origin error_details
```

I would also avoid re-using a branch after merging its PR.

atulep:
@tseaver, I followed your script, but it gave me: 
```
$ git rebase master
Current branch error_details is up to date.
```
Maybe it's easier to start a fresh PR with correct history.

summer-ji-eng (approved):
It's not blocker, but please add a test case for `error_info` once the [feature](https://github.com/googleapis/python-api-core/pull/307#pullrequestreview-802057101) merged.

Thank you very much. :)

PR #1085, opened 2021-11-13
feat: forward compatible diregapic LRO support

software-dov:
Detect whether a method fulfills the criteria for DIREGAPIC LRO.
If so, fudge the name of the generated method by adding the suffix
'_primitive'. This change is made for both the synchronous and async
client variants. Any generated unit tests are changed to use and
reference the fudged name.

The names of the corresponding transport method is NOT changed.


google/cloud/extended_operations_pb2.py
```diff
@@ -0,0 +1,132 @@
+# -*- coding: utf-8 -*-
```
vam-google, line 1: I believe we need to add these files into 
https://github.com/googleapis/python-api-common-protos/tree/main/google

We have done simiilar procedure in other langauges already, for example:
Java: https://github.com/googleapis/java-common-protos/tree/main/proto-google-common-protos/src/main/java/com/google/cloud

Ruby: https://github.com/googleapis/common-protos-ruby/tree/main/google-cloud-common/lib/google/cloud

Once we do that, I believe we can remove most of these files from this PR.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -472,7 +472,7 @@ def test_{{ method.name|snake_case }}(transport: str = 'grpc', request_type={{ m
         {% if method.client_streaming %}
         response = client.{{ method.name|snake_case }}(iter(requests))
         {% else %}
-        response = client.{{ method.name|snake_case }}(request)
+        response = client.{{ method_name }}(request)
```
vam-google, line 475: Can you please clarify, why `snake_case` is not needed anymore (here and in many other places)?

vam-google (commented):
Looks good! Thanks for doing it. I have only one major comment: most probably we would need to populate https://github.com/googleapis/python-api-common-protos with `externded_operations` stuff (see the corresponding comment for details), like we have done with the other langauges.

Another question: have we agreed on the `_primitive` suffix? I do not have a strong opinion on the suffix itself, as long as it is as consistent as possible with other languages.

google/cloud/extended_operations_pb2.py
busunkim96: https://github.com/googleapis/python-cloud-common/tree/main/google/cloud/common/types was created for `google/cloud/common_resources.proto`, so I think it will be a more appropriate repo/package for `extended_operations.proto`.

Can the BUILD.bazel https://github.com/googleapis/googleapis/blob/master/google/cloud/BUILD.bazel be updated to generate this proto as well?

Another note, we've switched to generating proto-plus types by default instead of _pb2.py, will that be a problem for the generator or generated library code? @software-dov

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
software-dov: Of course. We define `method_name` near the beginning of the file like so:
```
{% with method_name = method.name|snake_case + "_primitive" if method.operation_service else method.name|snake_case %}
```

Note the underscore: this is `method_name`, not `method.name`.

google/cloud/extended_operations_pb2.py
vam-google: https://github.com/googleapis/googleapis/blob/master/google/cloud/BUILD.bazel#L51 already has a python target for extended_operations. But that one will create default stubs. If you want generate proto-plus ones, then it is still straightforward, but woluld require changing that target to py_gapic_library and py_gapic_assembly targets instead. I'm happy to assist with that.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
vam-google: Oh, thanks! (I did not notice  `method_name` vs `method.name` difference and that what made me thing that snake_case just got removed

google/cloud/extended_operations_pb2.py
software-dov: The `extended_operations_pb2.py` file is just used by the generator; I don't believe generating proto-plus types will work because proto-plus can't handle message extensions.

software-dov: @vam-google The other files added to the PR are necessary for the fragment tests. The only file that can and should be moved is `extended_operations_pb2.py`.

software-dov:
@vam-google with regards to the `_primitive` suffix: the forward compatible LRO doc does not indicate that there is a consensus formed. Ruby does not use that naming convention, and it looks like go doesn't either.

vam-google:
@software-dov Which ones do ruby and go use?

tests/fragments/test_diregapic_forwardcompat_lro.proto
```diff
@@ -0,0 +1,101 @@
...
+      post: "/expensive/v1/computations/{computation}/fragments/{fragment}"
+    };
+  }
+
+  rpc StartCheapComputation(StartCheapComputationRequest) returns (StartCheapComputationResponse) {
+    option (google.api.http) = {
+      post: "/expensive/v1/computations/{computation}/fragments/{fragment}";
+    };
+  }
+}
```
vam-google, line 101: Please add new line end character here and in other places

vam-google (approved):
LGTM, but please first finalize the suffix we want to use for this (`_primitive` or something else) with @vchudnov-g, and upate this PR accordingly to feature the proper suffix.

PR #1089, opened 2021-11-17
fix: fix resource path args for paths with =**

busunkim96:
Some resources have `=**` in the segment. . A segment with `=**` is not matched by the current regex:

```py
Python 3.9.5 (default, Jul 27 2021, 22:06:04) 
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import re
>>> pattern = re.compile(r"\{([a-zA-Z0-9_-]+)\}")
>>> pattern.search("projects/{project}/metricDescriptors/{metric_descriptor=**}").groups()
('project',)

```

This pattern shows up in the some real APIs ([monitoring protos](https://github.com/googleapis/googleapis/blob/3b9c98eda3bded7bb01e2f5f5c7d20f4a5d3e121/google/monitoring/v3/metric_service.proto#L39-L46), `google/cloud/iap/v1`, `google/cloud/iap/v1beta1`, `google/cloud/recommendationengine/v1beta1`, `google/devtools/remoteworkers/v1test2`, `google/home/graph/v1`, `google/iam/v1`). `**` is mentioned in passing in [Resource Names](https://cloud.google.com/apis/design/resource_names#q_how_should_i_generate_and_parse_resource_names). I was not able to find an explanation of what wildcards are considered valid in https://google.aip.dev/122 or https://google.aip.dev/client-libraries/4231.

Monitoring Proto Example:
```proto
option (google.api.resource_definition) = {
  type: "monitoring.googleapis.com/MetricDescriptor"
  pattern: "projects/{project}/metricDescriptors/{metric_descriptor=**}"
  pattern: "organizations/{organization}/metricDescriptors/{metric_descriptor=**}"
  pattern: "folders/{folder}/metricDescriptors/{metric_descriptor=**}"
  pattern: "*"
  history: ORIGINALLY_SINGLE_PATTERN
};
```


gapic/samplegen/samplegen.py
```diff
@@ -110,7 +110,7 @@ class TransformedRequest:
 
     # Resource patterns look something like
     # kingdom/{kingdom}/phylum/{phylum}/class/{class}
-    RESOURCE_RE = re.compile(r"\{([^}/]+)\}")
```
busunkim96, line 113: This regex seemed to be serving the same purpose as the one in `wrappers.py`

tseaver (approved):

software-dov (approved):

PR #1098, opened 2021-12-01
fix: ensure rest unit tests have complete coverage

kbandes:
This PR fixes gaps in the coverage of the generated unit tests, specifically when the rest transport is included.

This required refactoring the logic for required field handling into a separate static method corresponding to each API method with required fields in its request message.  This allows that method to tested thoroughly without elaborate and repetitive mocking.

The intention for these changes is that the nox `showcase_unit` session should include both grpc and rest transports.  However, there are several requirements before the noxfile should be changed:
1.  The version of api_core with the rest operations client should be released (added in https://github.com/googleapis/python-api-core/pull/311, not yet released).
2. The missing signature field fix in showcase should be released (fixed in https://github.com/googleapis/gapic-showcase/pull/930, not yet released).
3. The extended operations module should be moved to api_core and released (no PR yet that I know of).

Once these changes are release, the noxfile for the present repo should be updated to reference those releases.  At that point, a full unit test of showcase with `transport=grpc+rest` should succeed and provide 100% coverage.

### DIFF
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
@@ -168,26 +168,17 @@ class {{service.name}}RestTransport({{service.name}}Transport):
     {%- if method.http_options and not (method.server_streaming or method.client_streaming) %}
 
     {% if method.input.required_fields %}
-    @staticmethod
-    def _{{ method.name | snake_case }}_populate_required_fields(message_dict):
-        required_fields = [
-            {% for req_field in method.input.required_fields %}
-            {% if req_field.is_primitive %}
-            (
-                "{{ req_field.name | camel_case }}",
-                {% if req_field.field_pb.default_value is string %}
-                "{{ req_field.field_pb.default_value }}"
-                {% else %}
-                {{ req_field.field_pb.default_value }}
-                {% endif %}{# default is str #}
-            ),
-            {% endif %}{# is primitive #}
-            {% endfor %}{# required fields #}
-        ]
+    __{{ method.name | snake_case }}_required_fields_default_values =  {
+    {% for req_field in method.input.required_fields if req_field.is_primitive %}
+        "{{ req_field.name | camel_case }}" : {% if req_field.field_pb.default_value is string %}"{{req_field.field_pb.default_value }}"{% else %}{{ req_field.field_pb.default_value }}{% endif %}{# default is str #}
+    {% endfor %}
+    }
 
-        for field_name, default_value in required_fields:
-            if field_name not in message_dict:
-              message_dict[field_name] = default_value
+
+    @staticmethod
+    def _{{ method.name | snake_case }}_get_unset_required_fields(message_dict):
+        return {k: v for k, v in {{service.name}}RestTransport.__{{ method.name | snake_case }}_required_fields_default_values.items() if k not in message_dict}
+
 
     {% endif %}{# required fields #}
 
@@ -231,21 +222,6 @@ class {{service.name}}RestTransport({{service.name}}Transport):
             {% endfor %}
         ]
 
-        {% if method.input.required_fields %}
-        required_fields = [
-            # (snake_case_name, camel_case_name)
-            {% for req_field in method.input.required_fields %}
-            {% if req_field.is_primitive %}
-            (
-                "{{ req_field.name | snake_case }}",
-                "{{ req_field.name | camel_case }}"
-            ),
-            {% endif %}{# is primitive #}
-            {% endfor %}{# required fields #}
-        ]
-
-        {% endif %}
-
         request_kwargs = {{method.input.ident}}.to_dict(request)
         transcoded_request = path_template.transcode(
             http_options, **request_kwargs)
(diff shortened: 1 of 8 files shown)


gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -166,6 +166,31 @@ class {{service.name}}RestTransport({{service.name}}Transport):
     {% endif %}{# service.has_lro #}
     {% for method in service.methods.values() %}
     {%- if method.http_options and not (method.server_streaming or method.client_streaming) %}
+
+    {% if method.input.required_fields %}
+    @staticmethod
+    def _{{ method.name | snake_case }}_populate_required_fields(message_dict):
```
software-dov, line 172: I would prefer it if we moved mutation to the caller, something like

```python
__{{ method.name | snake_case }}_required_fields_default_values =  {
    {% for req_field in method.input.required_fields if req_field.is_primitive %}
        "{{ req_field.name | camel_case }}" : {% if req_field.field_pb.default_value is string %}"{{req_field.field_pb.default_value }}"{% else %}{{ req_field.field_pb.default_value }}{% endif %}{# default is str #}
    {% endfor %}
}


def _{{ method.name | snake_case }}_get_unset_required_fields(message_dict):
    return {k, v for k, v in __{{ method.name | snake_case }}_required_fields_default_values.items() if k not in message_dict}

    ....
    query_params.update(self._{{ method.name | snake_case }}_get_unset_required_fields(query_params)
```

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -1770,7 +1837,7 @@ def test_{{ service.name|snake_case }}_rest_lro_client():
     # Ensure that we have a api-core operations client.
     assert isinstance(
         transport.operations_client,
-        operations_v1.OperationsClient,
+        operations_v1.AbstractOperationsClient,
```
vam-google, line 1840: Why was this changed?

vam-google (approved):
LGTM, with one question.

Also, can you please provide the links to the corresponding PRs that are expected to be released to unblock 100% coverage (i.e. for the 3 items in the PR description)

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -1210,6 +1212,61 @@ def test_{{ method_name }}_rest(transport: str = 'rest', request_type={{ method.
     {% endif %}
 
 
+    {% if method.input.required_fields %}
```
software-dov, line 1215: It might be reasonable to have a test or two just for the hidden required fields update method.

kbandes: I'm not sure what you mean.  Which method is hidden?

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
kbandes: I can make this change.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
kbandes: Oh, I see, once we refactor as you describe above.
This is quite tricky to test, which is a lot of motivation for these changes.  The problem is that typically, though not always, a required field is going to have an expected template in the http rule, so the default value will cause the transcoding to fail.  This can be worked around by mocking the transcoding function, but it gets convoluted and ugly.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
kbandes: This was at the request of software-dov.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
kbandes: Done.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
kbandes: I'm not sure what to do here.  No code is un-covered.  I would argue that testing this logic in the context of the api method itself doesn't necessarily add anything.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
kbandes: Regarding related PRs:  I'm not sure.  I believe other people are making these changes and releasing them, I'm not sure of the timeline or PR numbers.  The change to noxfile.py is small, but I won't know exactly how to do it until I know what the actual release numbers for showcase and api-core are.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
kbandes: Okay, I've added logic to test the actual api method with default-valued required fields.

software-dov (approved):

PR #1099, opened 2021-12-01
chore: remove actools@ from CODEOWNERS

noahdietz:
Removes actools@ from CODEOWNERS. Vadym, Aza, Sijun, Victor, and Dov should be able to approve things via actools-python.

This makes things a little less noisy for the rest of the team and ensures that we cannot approve things we shouldn't be able to.


noahdietz:
Looks like Dov still needs to be synced into the Team.

parthea (approved):

PR #1094, opened 2021-11-19
feat: add support for long-running operations with rest transport.

kbandes:
This PR adds support for long-running operations in the rest transport.  Some noteworthy points:

- the python gapic generator now reads the service yaml file for an api, using a new `service_yaml` parameter in `py_gapic_library` targets.  The content of this file is loaded into a `google.api.Service` proto and is available in the schema as the `API.service_yaml_config` property.  This can, in the future, support other features that rely on information in this file.  Note that the java, golang, php, and nodejs targets already have this parameter.

- In addition to unit tests, this change has been tested against a live server using the cloud translate api.
Note that this change is dependent upon https://github.com/googleapis/python-api-core/pull/311.

PR #1103, opened 2021-12-09
fix: syntax fix for required_fields struct in rest transport

### LINKED ISSUE #1102: Issues with gapic generator python version 0.58.0
There are 2 issues with gapic generator python version 0.58.0

1. A syntax error exists in the generated client for python compute here:

https://github.com/googleapis/python-compute/blob/73df2b5d16f67bf914a72772521003968d7880ac/owl-bot-staging/v1/google/cloud/compute_v1/services/addresses/transports/rest.py#L195-L196

See https://github.com/googleapis/python-compute/pull/168 where owlbot failed to run as a result of the syntax error.

2. Unused imports were added in many generated clients. For example, see https://github.com/googleapis/python-documentai/pull/251/files

software-dov:


kbandes (approved):
Thanks!

parthea (changes requested):
There is still an issue with unused imports in grpc based clients where [this line](https://github.com/googleapis/gapic-generator-python/blob/master/gapic/templates/tests/unit/gapic/%25name_%25version/%25sub/test_%25service.py.j2#L43) is importing `json_format` which is guarded by `{% if service.has_lro %}` regardless of grpc or rest based transport but the usage [here](https://github.com/googleapis/gapic-generator-python/blob/master/gapic/templates/tests/unit/gapic/%25name_%25version/%25sub/test_%25service.py.j2#L1176) is only for rest clients.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -40,8 +42,10 @@ from google.api_core import future
 from google.api_core import operation
```
parthea, line 42: PR #1094 also added this line. If it’s only used for rest transports, we should also guard it with

```
 {% if "rest" in opts.transport %}
```

parthea (commented):
Added one observation, otherwise LGTM

parthea (approved):

PR #1105, opened 2021-12-09
fix: syntax fix and test for multiple required fields

software-dov:


kbandes (approved):

tests/fragments/google/api/field_behavior.proto
```diff
@@ -0,0 +1,90 @@
+// Copyright 2018 Google LLC
```
vam-google, line 1: Isn't this file already in common protos?

vam-google (approved):
LGTM

tests/fragments/google/api/field_behavior.proto
software-dov: Yes it is, but I just copied it in here instead of trying to clone the repo as part of fragment testing.

PR #1107, opened 2021-12-10
fix: fix case for expected field names in required fields test.

kbandes:
The generated test for methods with required fields should have expected the field names to use camel case, but did not.  This should correct that.


vam-google (approved):
LGTM

PR #1108, opened 2021-12-11
fix: non-string required fields provide correct values

software-dov:


kbandes (approved):

gapic/schema/wrappers.py
```diff
@@ -757,17 +757,80 @@ class HttpRule:
...
+
+            Args:
+                  paths: a list of tuples, each with a (segmented) name and a pattern.
+            Returns:
+                  A new nested dict with the templates instantiated.
+            """
+
+            request: Dict[str, Any] = {}
+
+            def _sample_names() -> Generator[str, None, None]:
```
jameslynnwu, line 781: Iterator[str] instead?

[How to annotate a generator in python3?](https://stackoverflow.com/a/63237329)

Also, could we do something like naming the samples based on the last component of the path? It might make the sample names look a little better.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -170,7 +170,7 @@ class {{service.name}}RestTransport({{service.name}}Transport):
     {% if method.input.required_fields %}
     __{{ method.name | snake_case }}_required_fields_default_values =  {
     {% for req_field in method.input.required_fields if req_field.is_primitive %}
-        "{{ req_field.name | camel_case }}" : {% if req_field.field_pb.default_value is string %}"{{req_field.field_pb.default_value }}"{% else %}{{ req_field.field_pb.default_value }}{% endif %},{# default is str #}
+        "{{ req_field.name | camel_case }}" : {% if req_field.field_pb.type == 9 %}"{{req_field.field_pb.default_value }}"{% else %}{{ req_field.type.python_type(req_field.field_pb.default_value or 0) }}{% endif %},{# default is str #}
```
jameslynnwu, line 173: `req_field.field_pb.type == 9` feels strange to me. I assume it is from https://googleapis.dev/python/protobuf/latest/google/protobuf/descriptor_pb2.html#google.protobuf.descriptor_pb2.FieldDescriptorProto.TYPE_STRING. Can't we use that constant directly? is there any other way to do this?

Can't this python_type function do the work of passing returning the correct thing depending on what type we are working with?

gapic/schema/wrappers.py
software-dov: I would just a soon rip out all the type annotation. `Iterator[str]` does look more correct, will fix

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
software-dov: It's difficult and messy to do module import and access from the template logic. I agree that using magic numbers here is suboptimal, but it's the better solution to a nasty problem.

`python_type` returns a Python type, e.g. `float`, `int`, `bool`. What I'm using it here for is as a constructor and giving it a _real_ default of 0, which turns into `0`, `0.0`, or `False`.

gapic/schema/wrappers.py
```diff
@@ -757,17 +757,80 @@ class HttpRule:
...
+    def path_fields(self, method: "~.Method") -> List[Tuple[Field, str, str]]:
         """return list of (name, template) tuples extracted from uri."""
-        return [(match.group("name"), match.group("template"))
+        input = method.input
+        return [(input.get_field(*match.group("name").split(".")), match.group("name"), match.group("template"))
                 for match in path_template._VARIABLE_RE.finditer(self.uri)]
 
-    @property
-    def sample_request(self) -> str:
+    def sample_request(self, method: "~.Method") -> str:
```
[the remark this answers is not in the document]
software-dov: This code needs to be a method of `HttpRule` so that it can touch attributes of the instance; it's really unpleasant to interact with free functions from random modules in the template logic, and almost always better to move necessary logic into a method on the object you're working on.

vam-google (approved):
LGTM

PR #1110, opened 2021-12-13
fix: temporarily disable failing geneated unit tests for compute

software-dov:


vam-google (changes requested):
This PR disables more than expected. The idea was to rollback https://github.com/googleapis/gapic-generator-python/pull/1098/files

PR #1117, opened 2021-12-30
fix: handle message bodies

software-dov:
Some methods with http annotations have body fields that are message types.
Previously, generated unit tests did not handle this well.

This is a fix for that: generate a reasonable mock value for the
message, represented as a dict.

### DIFF
--- a/tests/fragments/test_non_primitive_body.proto
+++ b/tests/fragments/test_non_primitive_body.proto
@@ -1,0 +1,53 @@
+// Copyright (C) 2021  Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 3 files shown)


tests/fragments/test_non_primitive_body.proto
```diff
@@ -0,0 +1,53 @@
+// Copyright (C) 2021  Google LLC
```
vam-google, line 1: Is this file used? If yes, can you please explain  how (I could not find any referrences to it from other parts of this PR

vam-google (approved):
LGTM. Just one question and please fix the tests before pushing.

tests/fragments/test_non_primitive_body.proto
software-dov: It's used in the fragment tests. For each proto in this directory, we generate a GAPIC library and run its generated unit tests. This stresses both the templates and the actual logic of the generated surface. See noxfile.py:fragment for more detail on how the fragment tests work.
This particular fragment is the minimum required reproduction for the issues that compute was having.

Issue #950, opened 2021-07-09
gapic-generator-python fails to complete with a sufficiently large protobuf tree


tmc:
This generator fails when there is a type reference to a very large protobuf type tree. (For example, the types described at https://github.com/google/fhir/tree/master/proto/google/fhir/proto ).

Initial debugging appears to show that MessageType.with_context is where the majority of the time is spent.

tmc:
@software-dov Curious if you have anything to share regarding this.

tmc:
@software-dov I experimented and by moving away from FrozenSet and by ensuring the same set() instance is threaded through the service/message/field tree the generation time+memory goes down from non-terminating and consuming 32G to finishing within 2 minutes.

vam-google:
Waiting for https://github.com/googleapis/gapic-generator-python/pull/1125 to be merged. @tmc can you please follow up on it? Thanks!

tmc:
@vam-google I think that PR needs review.

vchudnov-g:
It seems like this was fixed by #1125 . If not, please file a new issue so we can follow up. Thanks!

PR #1125, opened 2022-01-08
fix: Change to Set vs FrozenSet and thread the same set through

tmc:
This improves the performance dramatically for large type hierarchies.

Addresses #950

### DIFF
--- a/gapic/schema/wrappers.py
+++ b/gapic/schema/wrappers.py
@@ -27,7 +27,6 @@ Reading of underlying descriptor properties in templates *is* okay, a
 Documentation is consistently at ``{thing}.meta.doc``.
 """
 
-import sys
 import collections
 import copy
 import dataclasses
--- a/gapic/schema/wrappers.py
+++ b/gapic/schema/wrappers.py
@@ -357,7 +357,7 @@ class Field:
             self,
             *,
             collisions: Set[str],
-            visited_messages: Set["MessageType"] = None,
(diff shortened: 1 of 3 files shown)


software-dov:
Thanks so much for making these changes! Performance on big, complex APIs has been a pain point for a while.
What APIs have you tested on? Do you have any before and after benchmark data? I'm a little bit concerned about changing `frozenset` to `set` because of the risk of mutating the caller's parameters. If you can show that the performance gained from that is substantial it would put my mind at ease.

vam-google:
@tmc PTAL

tmc:
@vam-google any example that reference the protobuf types from https://github.com/google/fhir will not effectively finish without a change such as this one.

parthea:
@tmc Please could you address the review comment and feedback from software-dov?

tmc:
@atulep @parthea PTAL.

parthea:
Converting to draft as presubmits are failing

tmc:
@parthea PTAL

gapic/schema/wrappers.py
```diff
@@ -27,6 +27,7 @@
 Documentation is consistently at ``{thing}.meta.doc``.
 """
 
+import sys
```
parthea, line 30:
```suggestion
```

parthea (dismissed):

parthea (approved):

PR #1129, opened 2022-01-11
feat: generate snippet metadata 

busunkim96:
Follow up to #1121.

**Main changes:**
* When snippets are generated, snippet metadata is also generated.
* If a method has a snippet, that snippet is included in the method docstring.

**Other changes:**
* Removed the method docstring in the code snippet file (line right below the method definition) since the same text is already in the comment block at the top of the file.
* Removed the concept of a "standalone" sample. All generated samples are expected to be standalone. When someone wants a smaller portion of the sample (e.g., request initialization only) they should fetch it from the file by looking up the line numbers in the snippet metadata file.

Other Notes:
* ~It doesn't look like it's possible to do type annotations with `_pb2` types, so those are annotated as `Any`.~ It is possible to do mypy checking with https://github.com/dropbox/mypy-protobuf, but I think it will be easier make that change in a separate PR.
* There are a lot of golden file updates, [this range of commits](https://github.com/googleapis/gapic-generator-python/pull/1129/files/872c156f5100f1de20631dd59d083206432db374) has _most_ of the generator and test changes.


gapic/generator/generator.py
```diff
@@ -190,36 +196,30 @@ def _generate_samples_and_manifest(
...
-        #         for fname, (spec, _) in fpath_to_spec_and_rendered.items()
-        #     ),
-        #     api_schema,
-        # )
-
-        # manifest_fname = os.path.join(out_dir, manifest_fname)
-        # output_files[manifest_fname] = CodeGeneratorResponse.File(
-        #     content=manifest_doc.render(), name=manifest_fname
-        # )
+        if len(index.metadata_index.snippets) > 0:
```
software-dov, line 211: Nit:
```python
if index.metadata_index.snippets:
```

gapic/generator/generator.py
```diff
@@ -190,36 +196,30 @@ def _generate_samples_and_manifest(
...
+            # NOTE(busunkim): Not all fields are yet populated in the snippet metadata.
+            # Expected filename: snippet_metadata_{metadata_schema_version}_{apishortname}_{apiversion}.json
+            snippet_metadata_path = str(pathlib.Path(
+                out_dir) / f"snippet_metadata_v1_{api_schema.naming.name}_{api_schema.naming.version}.json").lower()
```
software-dov, line 215: To clarify: the `v1` here is about the snippet metadata itself, correct?

gapic/generator/generator.py
```diff
@@ -190,36 +196,30 @@ def _generate_samples_and_manifest(
...
-        #     api_schema,
-        # )
-
-        # manifest_fname = os.path.join(out_dir, manifest_fname)
-        # output_files[manifest_fname] = CodeGeneratorResponse.File(
-        #     content=manifest_doc.render(), name=manifest_fname
-        # )
+        if len(index.metadata_index.snippets) > 0:
+            # NOTE(busunkim): Not all fields are yet populated in the snippet metadata.
+            # Expected filename: snippet_metadata_{metadata_schema_version}_{apishortname}_{apiversion}.json
```
software-dov, line 213: Is this comment still relevant, or is it cruft?

gapic/generator/generator.py
```diff
@@ -324,8 +325,8 @@ def _get_file(
         template_name: str,
         *,
         opts: Options,
-        api_schema=api.API,
-        **context: Mapping,
+        api_schema: api.API,
```
software-dov, line 328: Doh! Good catch.

gapic/samplegen/samplegen.py
```diff
@@ -927,11 +925,7 @@ def parse_handwritten_specs(sample_configs: Sequence[str]) -> Generator[Dict[str
                     raise types.InvalidConfig(
                         "Sample config is invalid", valid)
```
software-dov, line 926: Outside the scope of the review, but taking a second look, this isn't very helpful. Do you think it would be better to include the config or the filepath tothe config here?

gapic/samplegen/samplegen.py
```diff
@@ -1062,7 +1055,7 @@ def generate_sample_specs(api_schema: api.API, *, opts) -> Generator[Dict[str, A
                 yield spec
 
 
-def generate_sample(sample, api_schema, sample_template: jinja2.Template) -> str:
+def generate_sample(sample, api_schema, sample_template: jinja2.Template) -> Tuple[str, Any]:
```
software-dov, line 1058: Why use `Any` as the type of the second second element instead of the real type? Is it related to the `# type: ignore` a few lines down?

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
```diff
@@ -207,6 +207,13 @@ class {{ service.async_client_name }}:
             {% endif %}
         r"""{{ method.meta.doc|rst(width=72, indent=8) }}
 
+        .. code-block::
```
software-dov, line 210: Should the `.. code-block::` be inside the if statement a few lines below?

tests/integration/goldens/asset/google/cloud/asset_v1/services/asset_service/async_client.py
```diff
@@ -210,6 +210,31 @@ async def export_assets(self,
         the export operation result. For regular-size resource parent,
         the export operation usually finishes within 5 minutes.
 
+
+        .. code-block::
+        from google.cloud import asset_v1
```
software-dov, line 215: This indentation is a little unfortunate, but we can probably ignore it.

software-dov (approved):
Looks great! Just a few inline questions, no need to block submission.

tests/integration/goldens/asset/samples/generated_samples/snippet_metadata_v1_asset_v1.json
```diff
@@ -0,0 +1,1048 @@
+{
```
amanda-tarafa, line 1: For the time beeing, we dropped versioning of the metadata scheam so you can drop the first v1 here.
Also, we are doing, where possible, snippet_metadata_{protopackage}.json, although the real important part is snippet_metadata*.json as the rest of the name is not going to be used to identify anything automatically.

tests/integration/goldens/asset/samples/generated_samples/snippet_metadata_v1_asset_v1.json
```diff
@@ -0,0 +1,1048 @@
...
+  "snippets": [
+    {
+      "clientMethod": {
+        "async": true,
+        "method": {
+          "fullName": "AnalyzeIamPolicyLongrunning",
```
amanda-tarafa, line 7: fullName shuld be the RPC method name qualified by the proto package and the service name
This looks more like shortName.

tests/integration/goldens/asset/samples/generated_samples/snippet_metadata_v1_asset_v1.json
```diff
@@ -0,0 +1,1048 @@
...
+          "service": {
+            "shortName": "AssetService"
```
amanda-tarafa, line 9: And fullName would be the service name qualified by the proto package.
Fine if this is one the ones you are adding later.

tests/integration/goldens/asset/samples/generated_samples/snippet_metadata_v1_asset_v1.json
```diff
@@ -0,0 +1,1048 @@
...
+          }
+        }
+      },
+      "file": "samples/generated_samples/cloudasset_generated_asset_v1_asset_service_analyze_iam_policy_longrunning_async.py",
```
amanda-tarafa, line 13: This path should be relative to the location of this metadata file. Just double checking that's the case.

amanda-tarafa (approved):
I left a couple of comments on the generated metadata file, but it looks good!

gapic/generator/generator.py
busunkim96: Yes! Amanda left a note below saying the metadata will not be versioned for now though, so I'll go ahead and remove it.

gapic/generator/generator.py
busunkim96: Yep still relevant, there's a lot of fields that are not yet filled out.

gapic/samplegen/samplegen.py
busunkim96: Yes, since the type is ignored I cannot use it in an annotation. I learned about [mypy](https://github.com/dropbox/mypy-protobuf) from a PR with Tres though, so I'll open a separate PR to start type checking all the protobuf types.

tests/integration/goldens/asset/google/cloud/asset_v1/services/asset_service/async_client.py
busunkim96: Hrrm I didn't notice this earlier, I'll try poking at the template to see where this is coming from.

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
busunkim96: Ahh good catch, this would have created a lot of noise in the regen PRs.

tests/integration/goldens/asset/samples/generated_samples/snippet_metadata_v1_asset_v1.json
busunkim96: Ah it's currently relative to the library root.  I'll remove the `samples/generated_samples` since the metadata sits in the same directory as the snippets.

tests/integration/goldens/asset/samples/generated_samples/snippet_metadata_v1_asset_v1.json
busunkim96: Ack, I'll switch this out for `shortName`.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -363,6 +363,13 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
             {% endif %}
         r"""{{ method.meta.doc|rst(width=72, indent=8) }}
 
+        .. code-block::
```
software-dov, line 366: Same here about moving `.. code-block::` into the conditional

PR #1134, opened 2022-01-19
fix(snippetgen): remove special handling for resource paths

busunkim96:
There are some bugs in this implementation that stop generation
in some APIs and result in incorrect snippets getting generated
for others. Back out this feature for now so we can move forward
with enabling snippetgen for all APIs.


software-dov:
Which APIs are causing problems?

busunkim96:
I've opened an issue to track re-adding f-strings for resource paths https://github.com/googleapis/gapic-generator-python/issues/1144. Pub/Sub runs into the bug that halts generation and should be a good candidate for tests when I'm ready to work on this in the future.

busunkim96:
@software-dov Friendly ping

software-dov (approved):

PR #1142, opened 2022-01-20
feat: add interceptor-like functionality to REST transport

software-dov:
Interceptors are a gRPC feature that wraps rpcs in
continuation-passing-style pre and post method custom functions.
These can be used e.g. for logging, local caching, and tweaking
metadata.

This PR adds interceptor like functionality to the REST transport in
generated GAPICs.

The REST transport interceptors differ in a few ways:
1) They are not continuations. For each method there is a slot for a
"pre" function, and for each method with a non-empty return there is a
slot for a "post" function.
2) There is always an interceptor for each method. The default simply
does nothing.
3) Existing gRPC interceptors and the new REST interceptors are not
composable or interoperable.

### DIFF
--- a/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/rest.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/rest.py.j2
@@ -65,7 +65,7 @@ class {{ service.name }}RestInterceptor:
 
     .. code-block:
         class MyCustom{{ service.name }}Interceptor({{ service.name }}RestInterceptor):
-{% for _, method in service.methods|dictsort %}
+{% for _, method in service.methods|dictsort if not (method.server_streaming or method.client_streaming) %}
             def pre_{{ method.name|snake_case }}(request, metadata):
                 logging.log(f"Received request: {request}")
                 return request, metadata
@@ -81,7 +81,7 @@ class {{ service.name }}RestInterceptor:
 
 
     """
-    {% for method in service.methods.values()|sort(attribute="name") %}
+    {% for method in service.methods.values()|sort(attribute="name") if not(method.server_streaming or method.client_streaming) %}
     def pre_{{ method.name|snake_case }}(self, request: {{method.input.ident}}, metadata: Sequence[Tuple[str, str]]) -> Tuple[{{method.input.ident}}, Sequence[Tuple[str, str]]]:
         """Pre-rpc interceptor for {{ method.name|snake_case }}
 
--- a/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/rest.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/rest.py.j2
@@ -49,10 +49,67 @@ DEFAULT_CLIENT_INFO = gapic_v1.client_info.ClientInfo(
     rest_version=requests_version,
 )
 
+
+class {{ service.name }}RestInterceptor:
+    """Interceptor for {{ service.name }}.
+
+    Interceptors are used to manipulate requests, request metadata, and responses
+    in arbitrary ways.
+    Example use cases include:
+    * Logging
+    * Verifying requests according to service or custom semantics
+    * Stripping extraneous information from responses
+
+    These use cases and more can be enabled by injecting an
+    instance of a custom subclass when constructing the {{ service.name }}RestTransport.
+
+    .. code-block:
+        class MyCustom{{ service.name }}Interceptor({{ service.name }}RestInterceptor):
+{% for _, method in service.methods|dictsort %}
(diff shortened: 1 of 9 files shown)


gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -49,10 +49,67 @@ DEFAULT_CLIENT_INFO = gapic_v1.client_info.ClientInfo(
     rest_version=requests_version,
 )
 
+
+class {{ service.name }}RestInterceptor:
```
BenRKarl, line 53: Is the expectation that we subclass this and override the `pre_*` and `post_*` methods below? Or I suppose we could pass an arbitrary object that has the necessary methods.

gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -233,7 +292,7 @@ class {{service.name}}RestTransport({{service.name}}Transport):
             },
             {% endfor %}{# rule in method.http_options #}
             ]
-
+            request, metadata = self._interceptor.pre_{{ method.name|snake_case }}(request, metadata)
```
BenRKarl, line 295: At first I imagined we'd be able to pass a _list_ of interceptors that could be chained together, but understand that requires additional architecture here. We can handle the chaining within the body of the `pre_`/`post_` functions.

BenRKarl (approved):

gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/rest.py.j2
software-dov: Exactly, that's easily done by something like this:
```python
class InterceptorChainer:
    def __init__(self, pre_chain, post_chain):
        self.pre_chain = pre_chain
        self.post_chain = post_chain
        
    def __getattr__(self, name):
        if name.startswith("pre_"):
            def pre(self, request, metadata):
                for i in self.pre_chain:
                    request, metadata = getattr(i, name)(request, metadata)
                return request, metadata
                
            return pre
                
        elif name.startswith("post_"):
            def post(self, response):
                for i in self.post_chain:
                    response = getattr(i, name)(response)
                return response
            
            return post
            
        else:
            raise AttributeError(f"No such attribute: {name}")
```

gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/rest.py.j2
software-dov: Since python is duck typed, so if the transport is passed something that has the appropriate `pre_*` and `post_*` methods the actual type doesn't matter. I just added this generated class to make it easy to selectively intercept particular calls and have a reasonable, i.e. do nothing, default.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -39,6 +39,7 @@ from google.api_core import grpc_helpers_async
 from google.api_core import path_template
 {% if service.has_lro %}
 from google.api_core import future
+from google.api_core import operation
```
vam-google, line 42: Probably a stupid question, but whasn't this required before? I.e. it is logical to assume that `{% if service.has_lro %}` that operation should be imported.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -1515,6 +1516,57 @@ def test_{{ method_name }}_rest_unset_required_fields():
     {% endif %}{# required_fields #}
 
 
+{% if not (method.server_streaming or method.client_streaming) %}
```
vam-google, line 1519: just in case, do bidirectional streaming methods have their own flag or are signified by both server_streaming and client_streaming being true?

vam-google (commented):
LGTM

Did you have a chance to test the new interceptor logic besides unit tests? I.e. maybe generate a client (compute, for example) and try adding custom interceptors (which would output something to stdout) there and make a call? I.e. some sort of integration end-to-end test?

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
software-dov: It's a different operation module :P

software-dov:
Manual verification of the interceptor functionality with TextToSpeech has been done. Both pre and post interceptor calls fire.

vam-google (approved):
LGTM,

let's not forget to fix the showase CI

PR #1145, opened 2022-01-21
feat: expose extended operations annotations within generator

software-dov:
Add internal generator methods to identify custom operations services.

This is a stepping-stone commit to fully support idiomatic, ergonomic LROs that follow the extended operations semantics.

### DIFF
--- a/gapic/schema/api.py
+++ b/gapic/schema/api.py
@@ -476,15 +476,15 @@ class API:
         )
 
     def get_custom_operation_service(self, method: "wrappers.Method") -> "wrappers.Service":
-        if not method.output.is_diregapic_operation:
-            raise ValueError(f"Method is not a DIREGAPIC LRO: {method.name}")
+        if not method.output.is_extended_operation:
+            raise ValueError(f"Method is not an extended operation LRO: {method.name}")
 
         op_serv_name = self.naming.proto_package + "." + \
             method.options.Extensions[ex_ops_pb2.operation_service]
         op_serv = self.services[op_serv_name]
         if not op_serv.custom_polling_method:
             raise ValueError(
-                f"Service is not a DIREGAPIC operation service: {op_serv.name}")
+                f"Service is not an extended operation operation service: {op_serv.name}")
 
         return op_serv
 
--- a/gapic/schema/api.py
+++ b/gapic/schema/api.py
@@ -30,7 +30,7 @@ from google.api_core import exceptions
 from google.api import http_pb2  # type: ignore
 from google.api import resource_pb2  # type: ignore
 from google.api import service_pb2  # type: ignore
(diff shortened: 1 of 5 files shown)


tests/unit/schema/test_api.py
```diff
@@ -1595,3 +1596,151 @@ def test_http_options(fs):
                 method='get', uri='/v3/{name=projects/*/locations/*/operations/*}', body=None),
             wrappers.HttpRule(method='get', uri='/v3/{name=/locations/*/operations/*}', body=None)]
     }
+
+
+def generate_basic_diregapic_setup():
```
vam-google, line 1601: WDYT about using `extended_operations` as a suffix/prefix here? Because seems like this is what the feature is about and "diregapic" is technically an iternal term (unlike extended_operations which were published by this name in googleapis).

gapic/schema/api.py
```diff
@@ -474,6 +475,19 @@ def requires_package(self, pkg: Tuple[str, ...]) -> bool:
...
+    def get_custom_operation_service(self, method: "wrappers.Method") -> "wrappers.Service":
+        if not method.output.is_diregapic_operation:
+            raise ValueError(f"Method is not a DIREGAPIC LRO: {method.name}")
+
+        op_serv_name = self.naming.proto_package + "." + \
+            method.options.Extensions[ex_ops_pb2.operation_service]
+        op_serv = self.services[op_serv_name]
+        if not op_serv.custom_polling_method:
+            raise ValueError(
+                f"Service is not a DIREGAPIC operation service: {op_serv.name}")
```
vam-google, line 487: Please do not use DIREGAPIC as a term.

gapic/schema/api.py
```diff
@@ -474,6 +475,19 @@ def requires_package(self, pkg: Tuple[str, ...]) -> bool:
             for message in proto.all_messages.values()
         )
 
+    def get_custom_operation_service(self, method: "wrappers.Method") -> "wrappers.Service":
```
vam-google, line 478: Looks like this touches only one annotation among any (the operation_service) one. What about the others?

vam-google (commented):
LGTM, but a few general questions to understand the general apporach and what to expect in future PRs.

tests/unit/schema/test_api.py
software-dov: That seems reasonable. I'll make the change throughout.

gapic/schema/api.py
software-dov: The important ones are already expose, this one was just a special case.

gapic/schema/api.py
software-dov: Removed

parthea (approved):

PR #1148, opened 2022-01-21
chore: fix CI invocations to use Python 3.9

### LINKED ISSUE #901: Allow setting REST transport client connection options.
In particular, allow explicitly selecting `http` over `https`. #900 is doing this determination automatically based on the hostname (for use in local integration tests), but it would be cleaner to allow this to be done explicitly at client instantiation time.

### LINKED ISSUE #1147: Showcase CI tests are always skipped
An example of a such showcase skipped run: https://github.com/googleapis/gapic-generator-python/runs/4887995943?check_suite_focus=true.

```
Run nox -s showcase
nox > Running session showcase
nox > Session showcase skipped: Python interpreter 3.9 not found.
```

As a result showcase tests are always reported as green even though they never actually ran.

software-dov:


parthea (approved):

gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -1251,7 +1251,7 @@ def test_{{ method.name|snake_case }}_rest_flattened():
         assert len(req.mock_calls) == 1
         _, args, _ = req.mock_calls[0]
         {% with uri = method.http_options[0].uri %}
-        assert path_template.validate("https://%s{{ uri }}" % client.transport._host, args[1])
+        assert path_template.validate("%s{{ uri }}" % client.transport._host, args[1])
```
vam-google, line 1254: Is it related to the main goal of this PR? or a bug found independently?

gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -175,6 +176,14 @@ class {{service.name}}RestTransport({{service.name}}Transport):
         # TODO(yon-mg): resolve other ctor params i.e. scopes, quota, etc.
         # TODO: When custom host (api_endpoint) is set, `scopes` must *also* be set on the
         # credentials object
+        maybe_url_match = re.match("^(?P<scheme>http(?:s)?://)?(?P<host>.*)$", host)
```
vam-google, line 179: Why do we need this? Looks like a heavyweight regexp to do validation. Isn't it safe to assume that if code reaches this poing the url already must be correct?

gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
software-dov: It's related to getting showcase tests to pass. We need to pass "http://localhost" as the host, and this mechanism lets the hostname either already have the http:// (preserving the interface for existing users) or not, in which case we can either use https or http.

gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/rest.py.j2
software-dov: It is not safe to make that assumption. The host is not validated at any point before this.

vam-google: right, but the host itself is not coming as a user input, but is rather generated, so it must be correct and the correctness of that is expected to be tested by relevant gapic-generator tests. Context: I'm just trying to avoid extra logic if possible on every request (anticipating Ads being unhappy with any extra millisecond spend on client layer).

parthea (approved):

gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/rest.py.j2
software-dov: The host _may_ be passed in directly. There is a default, but it is overridable.
This logic happens once during transport construction, so I would be surprised if it shows up during their benchmark tests.

PR #1178, opened 2022-02-02
fix: disambiguate field headers whose names are reserved python words

parthea:
Fixes #1177

PR #1187, opened 2022-02-03
fix: disambiguate method names that are reserved in transport classes

### LINKED ISSUE #1149: Naming conflict with `create_channel` in 2 generated clients
Unit tests are failing in the generated client for google/cloud/video/livestream/v1 [here](https://github.com/googleapis/googleapis-gen/tree/master/google/cloud/video/livestream/v1/video-livestream-v1-py).

Steps to reproduce

- clone googleapis-gen and cd in to the directory `google/cloud/video/livestream/v1/video-livestream-v1-py`
- run the python post processor image in the root of  `google/cloud/video/livestream/v1/video-livestream-v1-py` using
```
docker run --user $(id -u):$(id -g) --rm -v $(pwd):/repo -w /repo gcr.io/cloud-devrel-public-resources/owlbot-python:latest
```
- run unit tests for python 3.8 with `nox -s unit-3.8`
- The majority of tests fail with `'property' object is not callable`

See stacktrace in screenshot below.

`create_channel` appears in the same class as both a [property](https://github.com/googleapis/googleapis-gen/blob/master/google/cloud/video/livestream/v1/video-livestream-v1-py/google/cloud/video/livestream_v1/services/livestream_service/transports/grpc.py#L249) and a [method](https://github.com/googleapis/googleapis-gen/blob/master/google/cloud/video/livestream/v1/video-livestream-v1-py/google/cloud/video/livestream_v1/services/livestream_service/transports/grpc.py#L183). I believe this is due to the fact that there is an rpc with the name `CreateChannel ` [here](https://github.com/googleapis/googleapis/blob/master/google/cloud/video/livestream/v1/service.proto#L48) which is converted to snake case as `create_channel`.

Screen Shot 2022-01-24 at 7 35 18 AM

software-dov:
In addition to the method specific stubs they provide, the generated
transports expose other methods to e.g. create a gRPC channel.
This presents the opportunity for a naming collision if an API has a
CreateChannel method.

This PR disambiguates colliding method names at the transport
level. Client level method names are unchanged for ergonomic reasons.

### DIFF
--- a/gapic/schema/wrappers.py
+++ b/gapic/schema/wrappers.py
@@ -949,6 +949,21 @@ class Method:
         return getattr(self.method_pb, name)
 
     @property
+    def transport_safe_name(self) -> str:
+        # These names conflict with other methods in the transport.
+        # We don't want to disambiguate the names at the client level
+        # because the disambiguated name is less convenient and user friendly.
+        #
+        # Note: this should really be a class variable,
+        # but python 3.6 can't handle that.
+        TRANSPORT_UNSAFE_NAMES = {
+            "CreateChannel",
+            "GrpcChannel",
+            "OperationsClient",
+        }
+        return f"{self.name}_" if self.name in TRANSPORT_UNSAFE_NAMES else self.name
+
+    @property
     def is_operation_polling_method(self):
         return self.output.is_extended_operation and self.options.Extensions[ex_ops_pb2.operation_polling_method]
 
--- a/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
@@ -443,7 +443,7 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
 
         # Wrap the RPC method; this adds retry and timeout information,
         # and friendly error handling.
-        rpc = self._transport._wrapped_methods[self._transport.{{ method.name|snake_case}}]
+        rpc = self._transport._wrapped_methods[self._transport.{{ method.transport_safe_name|snake_case}}]
         {% if method.field_headers %}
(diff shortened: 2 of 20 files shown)


gapic/schema/wrappers.py
```diff
@@ -948,6 +948,21 @@ class Method:
...
 
+    @property
+    def transport_safe_name(self) -> str:
+        # These names conflict with other methods in the transport.
+        # We don't want to disambiguate the names at the client level
+        # because the disambiguated name is less convenient and user friendly.
+        #
+        # Note: this should really be a class variable,
+        # but python 3.6 can't handle that.
+        TRANSPORT_UNSAFE_NAMES = {
```
vam-google, line 959: Looks like it is the second in a row disambiguating PR (the other one: https://github.com/googleapis/gapic-generator-python/pull/1178). Can we (should we) have some sort of centralized name-conflict resolution mechanism? I think it is fine for this PR to be as is, but in case more things like that happen, it would be  good to have a unified name-conflict resolution unility class or something like that. WDYT?

vam-google (approved):
LGTM

gapic/schema/wrappers.py
software-dov: Different names that refer to different data structures need to be disambiguated against different sets of names. It doesn't matter too much if there's a field named `create_channel`. I'd argue that there are costs to being too aggressive disambiguating names, and that it makes more sense to deal with them in the `Method` or `Field` class when collisions occur.

vam-google: I agree that they have to be disamgiguated against different sets of names, but that set of names can be a standardized parameter to the generic disambiguating mechanism. Basically some mechanism though which all names go through, and get disambiguated depending on the context. But yeah, this is beyond the scope of this PR and we have many more important tasks to work on.

PR #1234, opened 2022-03-09
feat: full LRO for Extended Operations

software-dov:
Implementation and tests for Extended Operations.

The full description of Extended Operations is beyond the scope of
this changelog, but three important differences from 'normal' LROs are
1) The type of the 'Operation' is not fixed but is determined by
specification.
2) The operation service is not fixed and is determined by specification.
3) The polling method in the operation service is not fixed and is
determined by specification.

Methods that meet the Extended Operations criteria will generate two
methods in the client class: a 'plain' version with the "_unary"
suffix that returns the custom Operation type as a raw proto message,
and a 'rich' version that returns google.api_core.ExtendedOperation,
which can be polled and waited on for completion.


software-dov:
The following smoke test passes with the generated client:
```python
#! /usr/bin/env python

import os

from google.cloud.compute import AddressesClient, InsertAddressRequest


def main():
    request = InsertAddressRequest(
        address_resource={
            "address_type": "INTERNAL",
            "ip_version": "IPV4",
            "name": "test-name",
        },
        project=os.environ["PROJECT"],
        region="us-west1",
    )
    client = AddressesClient(
        transport="rest",
    )

    response = client.insert(request)

    i = 0
    while True:
        print(f"Entering loop at {i}")
        if response.done():
            print("Done!")
            break
        print(f"Exiting loop at {i}")
        i += 1


if __name__ == "__main__":
    main()
```
and generated the following output:
```
Entering loop at 0
Exiting loop at 0
Entering loop at 1
Exiting loop at 1
Entering loop at 2
Exiting loop at 2
Entering loop at 3
Exiting loop at 3
Entering loop at 4
Done!
```

software-dov:
Also note: generated unit tests are in no way complete or even started.

gapic/schema/api.py
```diff
@@ -902,6 +924,70 @@ def _maybe_get_lro(
...
+        )
+        if not operation_polling_method_pb:
+            raise ValueError(
+                f"Could not find operation polling method for custom operation service: {op_service_name}"
+            )
+
+        operation_request_key = service_address.resolve(
+            operation_polling_method_pb.input_type.lstrip(".")
+        )
+        operation_request_message = self.api_messages[operation_request_key]
```
atulep, line 965: Are you certain this won't raise no key found exception ?

gapic/schema/api.py
```diff
@@ -902,6 +924,70 @@ def _maybe_get_lro(
...
+        )
+        method_output_type = service_address.resolve(
+            meth_pb.output_type.lstrip(".")
+        )
+        if operation_type != method_output_type:
+            raise ValueError(
+                f"Inconsistent return types between extended lro method '{meth_pb.name}'"
+                f" and extended lro polling method '{operation_polling_method_pb.name}':"
+                f" '{method_output_type}' and '{operation_type}'"
+            )
```
atulep, line 978: Does this imply some problem with the proto file ?

gapic/schema/api.py
software-dov: If it does, we have bigger problems, and we should fail anyway; I could see an argument for checking and explicitly raising the exception with more info, though.

gapic/schema/api.py
software-dov: Yes, exactly. If the initial method (the long running method) and the polling method have different return message types, that is a problem with the proto.

gapic/schema/wrappers.py
```diff
@@ -383,14 +395,59 @@ def oneof_fields(self, include_optional=False):
 
         return oneof_fields
 
+    @utils.cached_property
+    def extended_operation_request_fields(self) -> Sequence[Field]:
```
atulep, line 399: How is this different than `operation_request_field`? My question is mostly born from not understanding LRO and extended LRO. Like, what is the relationship between these two methods here.

gapic/schema/wrappers.py
```diff
@@ -383,14 +395,59 @@ def oneof_fields(self, include_optional=False):
...
+            return None
+
+        def canonical_name(field):
+            return OperationResponseMapping.Name(field.operation_field).lower()
+
+        OperationResponseMapping = ex_ops_pb2.OperationResponseMapping
+        default_field_names = [
+            k.lower()
+            # The first variant is UNKNOWN
+            for k in ex_ops_pb2.OperationResponseMapping.keys()[1:]
```
atulep, line 434: Is this always sorted?

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
```diff
@@ -409,6 +409,10 @@ class {{service.name}}RestTransport({{service.name}}Transport):
 
     {% endfor %}
 
+    @property
+    def kind(self) -> str:
```
atulep, line 413: Could you use isinstance to determine this ?

gapic/templates/%namespace/%name_%version/%sub/types/_message.py.j2
```diff
@@ -75,4 +75,22 @@ class {{ message.name }}({{ p }}.Message):
...
     {% endif %}{# field.map #}
     {% endfor %}{# for field in message.fields.values#}
+
+    {% if message.extended_operation_status_field %}
+    {% with status_field = message.extended_operation_status_field %}
+    @property
+    def done(self) -> bool:
+        """Return True if the backing extended operation is completed, False otherwise."""
+        {% if status_field.enum %}
+        return self.{{ status_field.name }}.name == "DONE"
```
atulep, line 85: is this .name intended?

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest.py.j2
software-dov: We're trying to do something here that I don't think `isinstance` is a good choice for.
1) After construction, the client doesn't know anything about the type of the transport.
a) There are situations where the client doesn't know the type of the transport _during_ construction: if the transport is injected, or if no transport is given and the default type needs to be used.
2) The first time an extended LRO method is called, the client needs to instantiate an extended operations client, and it **SHOULD** use a corresponding transport type: grpc if the normal client's transport is grpc, rest if the normal client's transport is rest.
a) Because of 1a, we can't solve this problem by making the extended operations client eagerly or by storing the transport "kind" during construction.
3) We therefore need a mechanism to determine the "kind" of transport the normal client has, i.e. grpc or rest. If we use `isinstance`, we still need a way to turn that information into a decision about what kind of transport to make or the extended operations client. Something like
```python
ex_op_transport = None
{% if "rest" in opts.transports %}
if isinstance(self._transport, {{ service.name }}RestTransport):
    ex_op_transport = "rest"
{% endif %}
{% if "grpc" in opts.transports %}
if isinstance(self._transport, {{ service.name }}GrpcTransport):
    ex_op_transport = "grpc"
{% endif %}
if not ex_op_transport:
    raise ValueError(f"Unexpected transport type: {type(self._transport)}")

self._extended_operations_client = {{ api.get_extended_operation_service(m).name }}(transport=ex_op_transport)
```
That looks a little clunky, IMO, compared to the logic where the transports self identify their "kind" and route that through.
```python
# Something like this, it's elsewhere in the diff
self._extended_operations_client = {{ api.get_extended_operation_service(m).name }}(transport=self._transport.kind)
```
4) If we ever need to add another transport "kind", having this `kind` property lets us handle that while making fewer code changes. Static analysis will detect whether the new transport class defines an override for `kind`, and none of the rest of the logic needs to be modified. If we had the `isinstance` code block above, new transport classes would need to be explicitly added to the logic. Heck, it should be possible for the user to define transport classes for both the normal client and the extended operations client and use them together without any code changes in the client class.

gapic/schema/wrappers.py
software-dov: I don't believe that's a safe assumption for us to make. Good catch!

gapic/schema/wrappers.py
software-dov: This is a method on `Message`; `operation_request_field` is a method on `Field`. They are complementary: for a message, we want to know all the fields that are `operation_request_fields`, and for a field, we want to know which field in the operation request message it maps to.

tests/integration/goldens/redis/google/cloud/redis_v1/services/cloud_redis/client.py
```diff
@@ -418,7 +418,7 @@ def list_instances(self,
 
         .. code-block:: python
 
-            from google.cloud import redis_v1
+                    from google.cloud import redis_v1
```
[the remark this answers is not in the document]
software-dov: Yeah, that's my bad :P

gapic/templates/%namespace/%name_%version/%sub/services/%service/client_macros.j2
```diff
@@ -0,0 +1,264 @@
...
+        class _CustomOperation(extended_operation.ExtendedOperation):
+            {% for default_name, field in method.output.differently_named_extended_operation_fields.items() %}
+            @property
+            def {{ default_name }}(self):
+                return self._extended_operation.{{ field.name }}
+
+            {% endfor %}
+        {% endif %}
+
+        {# TODO(dovs): make the appropriate subclass that defines appropriate accessors #}
```
atulep, line 256: Remove if done.

atulep (approved):

atulep:
Consider adding `_` to macros file to help readers understand it's not meant to be included in the final client code.

gapic/templates/%namespace/%name_%version/%sub/types/_message.py.j2
vam-google: Yes it is true. To me it seems the second variant that Dov provided is better, as it uses enums directly, but I'm fine with either one.

PR #1273, opened 2022-04-11
fix: correct import for request message type

### LINKED ISSUE #1263: snippetgen: resourcemanager_v3.GetIamPolicyRequest type is not defined in the package but is described in the public doc example
https://github.com/googleapis/python-resource-manager/blob/4694fea88edc731f637e1e6d6e2bd860ac06c3a1/google/cloud/resourcemanager_v3/services/projects/client.py#L1464

Public Doc: https://cloud.google.com/python/docs/reference/cloudresourcemanager/latest/google.cloud.resourcemanager_v3.services.projects.ProjectsClient#google_cloud_resourcemanager_v3_services_projects_ProjectsClient_get_iam_policy

dizcology:
Fixes: https://github.com/googleapis/gapic-generator-python/issues/1263

The generated snippet where the request message type is external to the client library package now becomes:
```
from google.cloud import resourcemanager_v3
from google.iam.v1 import iam_policy_pb2  # type: ignore


def sample_get_iam_policy():
    # Create a client
    client = resourcemanager_v3.ProjectsClient()

    # Initialize request argument(s)
    request = iam_policy_pb2.GetIamPolicyRequest(
        resource="resource_value",
    )

    # Make the request
    response = client.get_iam_policy(request=request)

    # Handle the response
    print(response)
```

Additional changes:
- pytest fixtures in tests/unit/samplegen/test_samplegen.py are refactored to make it easier to add new unit test code without duplicating much of the code.

### DIFF
--- a/gapic/samplegen/samplegen.py
+++ b/gapic/samplegen/samplegen.py
@@ -305,6 +305,14 @@ class Validator:
         # the MessageType of the request object passed to the rpc e.g, `ListRequest`
         sample["request_type"] = rpc.input
 
+        # We check if the request object is part of the service proto package.
+        # If not, it comes from a different module.
+        address = rpc.input.meta.address
+        if address.proto_package.startswith(address.api_naming.proto_package):
+            sample["request_module_name"] = sample["module_name"]
+        else:
+            sample["request_module_name"] = address.python_import.module
+
         # If no request was specified in the config
         # Add reasonable default values as placeholders
         if "request" not in sample:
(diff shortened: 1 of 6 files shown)


parthea (changes requested):
Thanks for fixing this! The changes look good. I noticed a formatting issue in the downstream repo where 2 import statements appear are on the same line. Please could you take a look?
```
+            from google.cloud import resourcemanager_v3from google.iam.v1 import iam_policy_pb2  # type: ignore
```

Steps to re-create the issue:
1) Modify the WORKSPACE file [here](https://github.com/googleapis/googleapis/blob/master/WORKSPACE#L238) to use `local_repository` instead of `http_archive` and point to a clone of gapic generator python with this branch.
2) Run `bazel build //google/cloud/resourcemanager/v3:resourcemanager-v3-py` in the root of googleapis where you modified the `WORKSPACE` file in step 1
3) Run the following in a clone of https://github.com/googleapis/python-resource-manager
```
docker run --rm --user $(id -u):$(id -g)   -v $(pwd):/repo   -v $HOME/git/googleapis/bazel-bin:/bazel-bin   gcr.io/cloud-devrel-public-resources/owlbot-cli:latest copy-bazel-bin   --source-dir /bazel-bin --dest /repo
```
```
docker run --user $(id -u):$(id -g) --rm -v $(pwd):/repo -w /repo gcr.io/cloud-devrel-public-resources/owlbot-python:latest
```
4) Check the diff

dizcology:
@parthea Thanks for catching that!  I think that is caused by https://github.com/googleapis/gapic-generator-python/blob/main/gapic/samplegen_utils/snippet_index.py#L91, but from the original PR containing that `strip` call I did not see any discussion about why it was needed.  Is there a different styling requirement for these snippets in client method docstrings to have only one blank line before function definition, while in the standalone sample.py file there should be two blank lines?

parthea:
@dizcology I only see the formatting issue with the imports when I apply the changes from this PR. I don't see the issue in the `main` branch. It could be that the changes in this PR exposed an underlying issue. I haven't debugged that part of the code yet. Would you be interested in adding a test to verify the output of the sample generator? I don't see change changes to the golden files based on the changes in this PR so it's possible that we don't have any checks for the generated sample code.

dizcology:
Looks like there is some magic in client method docstring rendering that suppresses multiple new lines in some cases, so that strip was not needed after all.

parthea (approved):

PR #1276, opened 2022-04-13
fix: use google-api-core==2.7.2

### LINKED ISSUE #1274: v0.65.0 causes sample tests to fail in python-asset
I merged a renovate bot PR https://github.com/googleapis/gapic-generator-python/pull/1227 to update google-api-core which included a change that incorrectly modified retry logic in client libraries. This appears to break samples tests in https://github.com/googleapis/python-asset/pull/409/files. The issue is that instead of getting `google_api.core.exceptions.NotFound` when a resource is not found we get `google.api_core.exceptions.RetryError`. 

In [this commit](https://github.com/googleapis/python-api-core/commit/ef6f0fcfdfe771172056e35e3c990998b3b00416#diff-9e3107d123dc6c10d09336071035de56d78969e69389d39486ccc9a1483f40e3L29) from google-api-core, `grpc` was not imported if `grpcio-status` is not available. The issue was fixed in https://github.com/googleapis/python-api-core/pull/355. 

The issue should be resolved with `google-api-core==2.7.2` https://github.com/googleapis/python-api-core/pull/358

parthea:
Fixes #1274

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 click==8.1.2
-google-api-core==2.7.1
+google-api-core==2.7.2
 googleapis-common-protos==1.56.0
 jinja2==3.1.1
 MarkupSafe==2.1.1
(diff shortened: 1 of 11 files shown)


vam-google (approved):
LGTM

PR #1282, opened 2022-04-19
fix: disambiguate method names

### LINKED ISSUE #1277: Syntax error in generated client for sqladmin v1
The generated client for sqladmin v1 has a syntax error [here](https://github.com/googleapis/googleapis-gen/blob/master/google/cloud/sql/v1/sql-v1-py/google/cloud/sql_v1/services/sql_instances_service/transports/base.py#L155) where a reserved term `import` is used as a property name.

parthea:
Fixes #1277


vam-google (approved):
LGTM

PR #1284, opened 2022-04-19
fix: fixes bug when an annotation path in an http rule does not have =.

atulep:
Fixes a bug when an annotation path in an http rule does not have "=". The generator would not include such annotations into its metadata before sending a gRPC call.

For example, 

`/v1b3/projects/{project_id}/locations/{location}/jobs` would not be included in the metadata.


gapic/schema/wrappers.py
```diff
@@ -1200,7 +1200,8 @@ def field_headers(self) -> Sequence[FieldHeader]:
 
         http = self.options.Extensions[annotations_pb2.http]
 
-        pattern = re.compile(r'\{([a-z][\w\d_.]+)=')
+        # Copied from Node generator.
```
davidcavazos, line 1203: Thanks! Can we add a link to the GitHub file where this regex was copied from?

davidcavazos (approved):

parthea:
Aza, Please could you take a look at the test failure?

gapic/schema/wrappers.py
atulep: Done

vam-google (approved):
LGTM

PR #1003, opened 2021-09-30
feat: port REST transport to Ads templates

software-dov:
Add REST transport tests to Ads generated unit tests
Boost google-api-core dependency version
Minor bugfixes

PR #1291, opened 2022-04-30
fix: don't use stale session in rest transport

parthea:
Partial revert of https://github.com/googleapis/gapic-generator-python/pull/1003
Fixes b/230914094


vam-google (approved):
LGTM

PR #1297, opened 2022-05-01
chore: update setup.py

parthea:


setup.py
```diff
@@ -1,4 +1,6 @@
-# Copyright 2018 Google LLC
+# -*- coding: utf-8 -*-
+#
+# Copyright 2020 Google LLC
```
dandhlee, line 3: The copyright year shouldn't be updated.

dandhlee (changes requested):

setup.py
parthea: Fixed

dandhlee (approved):
LGTM.

PR #1304, opened 2022-05-06
feat: adds LRO mixin.

atulep:
Adds support for LRO to existing MixIns implementation. It extends client interface with `get_operation`, `wait_operation`, `cancel_operation`, `delete_operation`, `list_operations` methods (if they were defined in service yaml).


atulep:
@vam-google PTAL

tests/integration/goldens/redis/tests/unit/gapic/redis_v1/test_cloud_redis.py
```diff
@@ -3022,7 +3022,7 @@ def test_cloud_redis_base_transport():
         'export_instance',
         'failover_instance',
         'delete_instance',
-    )
+       )
```
vam-google, line 3025: Please make sure these unnecessary whitepace changes are removed

vam-google (approved):
Please check on showcase all 5 new methods (do actual network call) before pushing.

tests/integration/goldens/redis/tests/unit/gapic/redis_v1/test_cloud_redis.py
atulep: Done

PR #1305, opened 2022-05-07
ci: run unit tests for golden files

parthea:
This PR updates the presubmit checks to also verify the unit tests and coverage for the generated golden files.


gapic/templates/noxfile.py.j2
```diff
@@ -24,9 +24,9 @@ CURRENT_DIRECTORY = pathlib.Path(__file__).parent.absolute()
 LOWER_BOUND_CONSTRAINTS_FILE = CURRENT_DIRECTORY / "constraints.txt"
 PACKAGE_NAME = subprocess.check_output([sys.executable, "setup.py", "--name"], encoding="utf-8")
 
-BLACK_VERSION = "black==19.10b0"
+BLACK_VERSION = "black==22.3.0"
```
vchudnov-g, line 27: Is there an alias for the name of this and similar settings? Using the term "black" goes against our efforts to be more inclusive and counteract implicit bias.

PR #1330, opened 2022-05-30
fix: Break up big files in `types` directory

### LINKED ISSUE #1252: Code completion failing to work in PyCharm with generated client for python-compute
This issue was originally reported in python-compute [here](https://github.com/googleapis/python-compute/issues/245). The generated client for python-compute has a [types.py](https://github.com/googleapis/googleapis-gen/blob/master/google/cloud/compute/v1/compute-v1-py/google/cloud/compute_v1/types/compute.py) file which has ~60,000 lines of code and is greater than 2.5 MB. This exceeds the default maximum size limit of 2.5 MB defined [here](https://www.jetbrains.com/help/pycharm/file-idea-properties.html).

see `idea.max.intellisense.filesize` 
```
idea.max.intellisense.filesize | 2500 | Sets the maximum size of files (in kilobytes) for which PyCharm provides code assistance. Code assistance for large files can decrease editor performance and increase memory consumption.
```

### LINKED ISSUE #16436: Code completion failing to work in PyCharm with generated client for python-compute
This issue was originally reported in python-compute [here](https://github.com/googleapis/python-compute/issues/245). The generated client for python-compute has a [types.py](https://github.com/googleapis/googleapis-gen/blob/master/google/cloud/compute/v1/compute-v1-py/google/cloud/compute_v1/types/compute.py) file which has ~60,000 lines of code and is greater than 2.5 MB. This exceeds the default maximum size limit of 2.5 MB defined [here](https://www.jetbrains.com/help/pycharm/file-idea-properties.html).

see `idea.max.intellisense.filesize` 
```
idea.max.intellisense.filesize | 2500 | Sets the maximum size of files (in kilobytes) for which PyCharm provides code assistance. Code assistance for large files can decrease editor performance and increase memory consumption.
```

m-strzelczyk:
Fixes googleapis/google-cloud-python#16436

Dividing the classes for each proto into 3 separate files:

 * %proto/requests.py.j2
 * %proto/responses.py.j2
 * %proto/__init__.py.j2

The division is pretty simple. All "...Request" classes go to `requests.py` file, all "...Response" classes go to `responses.py` and everything else goes to `other.py`. This way, the content should be spread across multiple files and will move us away from the 2.5Mb PyCharm limit.

For comparison, in current version of `python-compute`, which I suspect might be the biggest GAPIC API library, the `types/compute.py` file is 2.8Mb big. After the change, we have:

 * %proto/requests.py - 1.5Mb
 * %proto/responses.py - 15Kb
 * %proto/__init__.py - 1.4Mb

This change gives us 1Mb distance from the 2.5Mb limit and should last relatively long.

Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/gapic-generator-python/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Fixes googleapis/google-cloud-python#16436 🦕


m-strzelczyk:
I tested this by replacing `google/cloud/compute/types` directory in my local `python-compute` installation by the `types` directory generated with this change. It worked fine.

vchudnov-g:
Thanks for doing this, @m-strzelczyk ! I have one overall question: what was your rationale for choosing to do a separation based on requests/responses/other as opposed to a separation based on proto service name? Naively, it seems to me the latter keeps related things closer together and is (arguably) more robust if "other" becomes really big itself, or if there is a change to naming conventions.

PR #1343, opened 2022-06-28
fix: resolve issue where rest test appears in grpc-only client

atulep:
Some APIs are failing with errors related to rest transport.

This is because one of the tests is missing conditional check on `opts.transport`.

See example of the failing test: https://github.com/googleapis/python-resource-settings/pull/118.

### DIFF
--- a/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
+++ b/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
@@ -929,7 +929,16 @@ def test_{{ service.name|snake_case }}_transport_auth_adc(transport_class):
         )
 
 
-{% if 'grpc' in opts.transport %}
+{% if 'grpc' in opts.transport and 'rest' in opts.transport %}
+@pytest.mark.parametrize(
+    "transport_class",
+    [
+        transports.{{ service.name }}GrpcTransport,
+        transports.{{ service.name }}GrpcAsyncIOTransport,
+        transports.{{ service.name }}RestTransport,
+    ],
+)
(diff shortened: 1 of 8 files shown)


m-strzelczyk (approved):

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -937,6 +938,22 @@ def test_{{ service.name|snake_case }}_transport_auth_adc(transport_class):
...
+    ],
+)
+{% elif 'rest' in opts.transport %}
+@pytest.mark.parametrize(
+    "transport_class",
+    [
+        transports.{{ service.name }}RestTransport,
+    ],
+)
+{% endif %}
```
parthea, line 956: The following could also work. WDYT?

```
suggestion
@pytest.mark.parametrize(
    "transport_class",
    [
    {% if 'grpc' in opts.transport %}
        transports.{{ service.name }}GrpcTransport,
        transports.{{ service.name }}GrpcAsyncIOTransport,
     {% endif %}
     {% if 'rest' in opts.transport %}
         transports.{{ service.name }}RestTransport,
     {% endif %}
    ],
)

```

parthea (approved):
LGTM, added one observation.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
parthea: I'm going to merge this as is since the generated output is the same.

Issue #1359, opened 2022-07-20
Type field mistakenly called as `type_` in generated unit tests


vam-google:
When generating logging python client (https://github.com/googleapis/googleapis/blob/master/google/logging/v2/BUILD.bazel#L191), the generated unit tests fails with:
```python
            response = await client.write_log_entries(
                log_name='log_name_value',
>               resource=monitored_resource_pb2.MonitoredResource(type_='type__value'),
                labels={'key_value': 'value_value'},
                entries=[log_entry.LogEntry(log_name='log_name_value')],
            )
E           ValueError: Protocol message MonitoredResource has no "type_" field.
```

This is currently being fixed by `owlbot.py` post-processing logic https://github.com/googleapis/python-logging/blob/main/owlbot.py#L29 but should be fixed in the generator itself

Issue #1358, opened 2022-07-20
Asterisk resource names result into broken resource name helper method


vam-google:
For example this asterisk resource name pattern:
https://github.com/googleapis/googleapis/blob/master/google/cloud/eventarc/v1/trigger.proto#L37

Causes the generated `def test_parse_service_path()` method fail with `nothing to repeat at position 1` error due to  a broken regexp (`r"^*$"`) in the corresponding generated helper method:
```python
    @staticmethod
    def parse_service_path(path: str) -> Dict[str,str]:
        """Parses a service path into its component segments."""
        m = re.match(r"^*$", path)
        return m.groupdict() if m else {}
```

PR #1360, opened 2022-07-20
feat: Introduce proper py_library and py_test support in bazel rules

vam-google:
These changes allow running the generated python gapic unit tests on the fly by bazel right after generation in our CI.

As a result of this change some of the issues with the generated code became immediatelly visible (because now we ran the generated tests) and will be addressed separately in these issues:

https://github.com/googleapis/gapic-generator-python/issues/1358
https://github.com/googleapis/gapic-generator-python/issues/1359

The tests can be now added in BUILD.bazel files as follows:
```bzl
py_test(
    name = "credentials_py_gapic_test",
    srcs = [
        "credentials_py_gapic_pytest.py",
        "credentials_py_gapic_test.py",
    ],
    legacy_create_init = False,
    deps = [
        ":credentials_py_gapic",
    ],
)
```


tests/integration/BUILD.bazel
```diff
@@ -67,6 +67,18 @@ py_gapic_library(
     ],
 )
 
+py_test(
```
parthea, line 70: could we add a placeholder for asset as well?
```
# uncomment once...
#py_test(
#    name = "asset_py_gapic_test",
#    srcs = [
#        "asset_py_gapic_pytest.py",
#        "asset_py_gapic_test.py",
#    ],
#    legacy_create_init = False,
#    deps = [
#        ":asset_py_gapic",
#    ],
#)
```

rules_python_gapic/pytest.py
```diff
@@ -0,0 +1,11 @@
+import sys
+import pytest
+import os
+
+
+if __name__ == '__main__':
+    sys.exit(pytest.main([
+        '--disable-pytest-warnings',
```
parthea, line 8: Should we keep warnings?

parthea (approved):
Added minor observations, otherwise LGTM

rules_python_gapic/pytest.py
vam-google: It emits a lot of garbage warnings, contaminating the output and those warnings seem to be more related to the execution environment instead of the tests themselves, so I decided to turn them off. It won't affect execution much, but it will make our log even longer (the current full build log accross whole googleapis is multple megabytes already)

tests/integration/BUILD.bazel
vam-google: Probably. Asset has other issues though (not that much about generator but about the wierdeness of the API itself, as it depends on many other APIs)

PR #1378, opened 2022-07-28
feat: Implement REST support for MixIns

atulep:
Implements MixIns support in REST. The implementation is different than what was used for GRPC version. Special wrappers for MixIns are created to make the code less repetitive. In a separate PR, it's possible to refactor GRPC implementation too.

NOTE: Support for IAM `opts` flag will be removed in the future PR.


atulep:
The relevant MixIns changes have not been ported over to ads yet.

gapic/templates/tests/unit/gapic/%name_%version/%sub/_test_mixins.py.j2
```diff
@@ -1,3 +1,70 @@
...
+    {{ sig.request_type }},
+    dict,
+])
+def test_{{ name|snake_case }}_rest(request_type):
+    client = {{ service.client_name }}(
+        credentials=ga_credentials.AnonymousCredentials(),
+        transport="rest",
+    )
+    request = request_type()
+    if isinstance(request_type, {{ sig.request_type }}):
```
vam-google, line 32: There is no need to call ParseDict explicitly.
Plesse replace this isinstance block with

```
    request_init = api.mixin_http_options["{}".format(name)][0].sample_request
    request = request_type(**request_init)
```

This will let us get rid of isincance, which reduces test coverage and also just  much more verbose. Plus, the `request_init` trick is used in the non-mixin tests already, so it will make the tests more consistent.

gapic/schema/mixins.py
```diff
@@ -0,0 +1,68 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from gapic.schema import wrappers
+
+MIXINS_MAP = {
```
vam-google, line 17: Why was this list map not necessary for grpc mixins?

vam-google (commented):
looks good, but please fix the test coverage first (the isinstance thing)

tests/unit/schema/test_api.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import collections
+from textwrap import wrap
```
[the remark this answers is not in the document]
atulep: Done

gapic/templates/tests/unit/gapic/%name_%version/%sub/_test_mixins.py.j2
atulep: Done

gapic/schema/mixins.py
atulep: Because GRPC mixins use a copy-pasting approach. Ideally, need to be refactored to use this map instead.

vam-google (approved):

PR #1399, opened 2022-08-12
feat: Encode numeric enums parameter with REST requests

vchudnov-g:


vchudnov-g:
@parthea  LMK where would be a good place to test this using our existing framework. IMO, the best place would be an integration test against Showcase, but we're still building that framework.

parthea (approved):

PR #1420, opened 2022-08-22
fix: Fix rest test generation for non-primitive values in query params

vam-google:
This is a partial rollback of https://github.com/googleapis/gapic-generator-python/pull/1389/commits/73b1373e4cc4a7ce05e7abf726521c58c2fb881e

### DIFF
--- a/gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
+++ b/gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
@@ -1109,7 +1109,7 @@ def test_{{ method_name }}_rest_required_fields(request_type={{ method.input.ide
         ))
 
     # verify fields with default values are dropped
-    {% for req_field in method.input.required_fields if req_field.name in method.query_params %}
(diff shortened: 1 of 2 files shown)


gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -1109,7 +1109,7 @@ def test_{{ method_name }}_rest_required_fields(request_type={{ method.input.ide
         ))
 
     # verify fields with default values are dropped
-    {% for req_field in method.input.required_fields if req_field.name in method.query_params %}
+    {% for req_field in method.input.required_fields if req_field.is_primitive and req_field.name in method.query_params %}
```
atulep, line 1112: This drops coverage when `is_primitive` is false. How would you test it?

vam-google: This is a rollback. It gets tested the same way as before. This affects the py_test taregets at minimum (the whole reason why I'm changing this is because it fails the generted unit tests)

atulep (approved):

PR #1421, opened 2022-08-22
fix: Fix REST tests generation for repeated enums

vam-google:


vchudnov-g (approved):
I don't grok the existing code 100%, but this looks fine AFAICT

PR #1428, opened 2022-09-01
fix: Fix remaining REST transport issues

vam-google:
This change depends on https://github.com/googleapis/python-api-core/pull/428 and needs api-core-2.9.0 to be released

This includes:
1) Use message-based transcoding logic instead of dictionary-based. This fixes multiple bugs caused by dict-json-object representation mismatches fro protobuf objects. This also removes a lot of `json->dict->object->proto-plus` object conversion overhead, which was slow and very error-prone. Now all transcodding and rest-transport-specific logic is executed in terms of native protobuf (`_pb`) stubs objects. No dicts, no proto-plus, no unnecessary conversions.
2) Fix for non-proto-plus objects in rest.py transport and in tests. This essential for APIs which  have IAM methods defined on their surface.
3) Fix support for Any field in tests
4) Fix creation of mock_values (do `mock_value1` instead of `mock_value_1`, since `mock_value_1` causes issues for `FieldMask` mocks


vam-google:
The showcase tests need to wait till api-core-2.9.0 gets released to pypi

alexander-fenster (approved):

parthea (approved):

PR #1436, opened 2022-09-08
fix(rest): Use strict encoding to lowercase query string bools

### LINKED ISSUE #1407: ensure fields are JSON encoded correctly
The best way of verifying correct behavior is probably to implement integration tests against the Showcase compliance suite, which test various data types.

Context: boolean fields seem to be not serialized correctly. I assigned `True` to a proto.Field as a kwarg in the message constructor:
```
request = showcase_v1beta1.EnumRequest(unknown_enum = True)
```

and the Showcase server returns with an error. The Showcase server logs read:
```
error reading query params: terminal field "unknownEnum" of field path "unknownEnum" is of type "bool" with value string "True", which could not be parsed: could not parse "True" as a bool
```
which is mostly due to True being sent capitalized rather than lower-cased, as the JSON spec requires.

This is a blocker for launching REGAPIC.

vchudnov-g:
- depend on google-api-core >= 2.10.0 for REST transport
- use `strict` to force query string booleans to be JSON encoded in lower case

Fixes #1407

### DIFF
--- a/gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
+++ b/gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
@@ -1214,7 +1214,7 @@ def test_{{ method_name }}_rest_required_fields(request_type={{ method.input.ide
                     {% if req_field.field_pb.type == 9 %}
                     "{{ req_field.field_pb.default_value }}",
                     {% else %}
-                    {{ req_field.type.python_type(req_field.field_pb.default_value or 0) }},
+                    str({{ req_field.type.python_type(req_field.field_pb.default_value or 0) }}),
                     {% endif %}{# default is str #}
                 ),
               {% endfor %}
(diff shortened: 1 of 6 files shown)


parthea (approved):

gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -1214,7 +1214,7 @@ def test_{{ method_name }}_rest_required_fields(request_type={{ method.input.ide
                     {% if req_field.field_pb.type == 9 %}
                     "{{ req_field.field_pb.default_value }}",
                     {% else %}
-                    {{ req_field.type.python_type(req_field.field_pb.default_value or 0) }},
+                    str({{ req_field.type.python_type(req_field.field_pb.default_value or 0) }}),
```
vam-google, line 1217: This does not seem right, and it fails generated unit tests for compute

vchudnov-g: Now that you mention it, yes, I think you're right. This "else" should be an "elseif" specific to bools. Yikes! That was quite an oversight on my part. Thanks for catching it.

PR #1440, opened 2022-09-09
fix: remove grpc only rpcs from rest.py

parthea:
This PR fixes an issue where rpcs which do not have the `google.api.http` annotation are included in `rest.py`


vam-google (approved):
LGTM assuming generated unit tests work

vchudnov-g (approved):

PR #1447, opened 2022-09-13
fix: unit test generation for boolean query prams

vam-google:
Without this fix the current compute engine tests are failing. Also 3 other APIs in googleapis would have failed if grpc+rest transport was enabled on them.


vchudnov-g (approved):
Thank you so much for catching and fixing this!

PR #1480, opened 2022-10-27
fix: Fix timeout default values

### LINKED ISSUE #1477: Missing default deadlines in generated stubs
#### Code example

```python
from google.cloud.bigtable_admin_v2 import BigtableTableAdminClient
from google.cloud.bigtable_admin_v2.types import ModifyColumnFamiliesRequest, ColumnFamily,GcRule

c = BigtableTableAdminClient()
c.modify_column_families(
    ModifyColumnFamiliesRequest(
        name="projects/my-project/instances/my-instance/tables/my-table",
        modifications=[
            ModifyColumnFamiliesRequest.Modification(
                id="family",
                create=ColumnFamily(
                    gc_rule=GcRule(
                        max_num_versions=1
                    )
                )
            )
        ]
    )
)
```

This should send the RPC using the default deadline configured here in service json file, which generates the defaults base.py:
https://github.com/googleapis/python-bigtable/blob/0841824c7b9539df2530370ea128c140eda0f2b2/google/cloud/bigtable_admin_v2/services/bigtable_table_admin/transports/base.py#L199-L203

However this deadline is not respected. I believe the problem comes from the generator using `None` as the default in the stub:

https://github.com/googleapis/python-bigtable/blob/0841824c7b9539df2530370ea128c140eda0f2b2/google/cloud/bigtable_admin_v2/services/bigtable_table_admin/client.py#L1229

Which will end up being pushed down into determine_timeout:
https://github.com/googleapis/python-api-core/blob/ddcc2499fa756f98268eb552edba6f25ddbe0c26/google/api_core/gapic_v1/method.py#L63-L101

Which seems to treat `None` to mean no deadline and wants Method.DEFAULT to mean the default deadline

Specifically this:
```python
   if specified_timeout is DEFAULT:
        specified_timeout = default_timeout
```

This effectively means that gapic generated clients will never send a default deadline

vam-google:
Also fix LRO for REST transport. This PR makes generated gapics appreciate timeout values from grpc_service_config.json instead of overriding them with None (which means no timeout)

It is basically a direct fix for https://github.com/googleapis/gapic-generator-python/issues/1477.

This PR depends on https://github.com/googleapis/python-api-core/pull/462, and expects `setup.py.j2` templates to be updated after https://github.com/googleapis/python-api-core/pull/462 gets pushed and released with new version.

BEGIN_COMMIT_OVERRIDE
fix: Fix timeout default values
fix(deps): require google-api-core >=1.34.0, >=2.11.0
END_COMMIT_OVERRIDE

Fixes https://github.com/googleapis/gapic-generator-python/issues/1477.


vam-google:
@igorbernstein2 PTAL

igorbernstein2 (commented):
the default looks correct, but maybe you want to use type-extensions to narrow the type annotation to a Literal instead of object?

parthea (approved):

PR #1488, opened 2022-11-11
chore: fix docs build

parthea:
This PR removes the `undoc-members` option of sphinx to address the error `duplicate object description` seen in the docs build for downstream clients. See build log [here](https://github.com/googleapis/python-access-approval/actions/runs/3438234164/jobs/5734028967) 

See the definition for `undoc-members` here:
> :undoc-members: (no value)[¶](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-option-automodule-undoc-members)
If set, autodoc will also generate document for the members not having docstrings:

https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-option-automodule-undoc-members


dandhlee (approved):
This should not be a problem for autogenerated clients. For handwritten libraries the maintainers have choice to add this themselves, so LGTM.

PR #1495, opened 2022-11-15
fix: snippetgen should call await on the operation coroutine before calling result

dizcology:
Currently the generated snippet contains the following code for **async** client that returns a **long running operation**, such as Vision API's `import_product_sets`:

```
    ...
    operation = client.import_product_sets(request=request)
    ...
    response = await operation.result()
    ...
```

This is incorrect since `operation` is a coroutine that needs to be `await`ed to return an actual operation (which has the `result` attribute).

The proposed fix is to change the generated code to:
```
     response = (await operation).result()
```

### DIFF
--- a/tests/integration/goldens/asset/google/cloud/asset_v1/services/asset_service/async_client.py
+++ b/tests/integration/goldens/asset/google/cloud/asset_v1/services/asset_service/async_client.py
@@ -239,7 +239,7 @@ class AssetServiceAsyncClient:
 
                 print("Waiting for operation to complete...")
 
-                response = await operation.result()
+                response = (await operation).result()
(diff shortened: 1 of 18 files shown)


tests/integration/goldens/asset/google/cloud/asset_v1/services/asset_service/async_client.py
```diff
@@ -239,7 +239,7 @@ async def sample_export_assets():
```
dizcology, line 239: An alternative is to change the line above to:
```
operation = await client.export_assets(request=request)
```

parthea (approved):

PR #1525, opened 2022-12-07
chore: configured snippet generates empty sample function

dizcology:
This PR illustrates the main mechanism of generating a configured snippet: Start with a minimal libcst object (the empty code module), and iteratively add to it (adding the sample function into the module's body).  Most of the TODO items in the `ConfiguredSnippet` class are similar "local" changes that can be tested and updated in isolation.  The two recursive cases (type annotation and statements) will be more complex.

Because the sample function is where most of the code will be added, we add a `ConfiguredSnippet._sample_function_def` attribute to simplify access.


gapic/configurable_snippetgen/configured_snippet.py
```diff
@@ -78,3 +78,26 @@ def filename(self) -> str:
...
+        """Generates the snippet.
+
+        This is the main entrypoint of a ConfiguredSnippet instance, calling
+        other methods to update self._module.
+        """
+        self._build_sample_function()
+        self._add_sample_function()
+        # TODO: add imports.
+        # TODO: add region tag.
+        # TODO: add header.
```
parthea, line 103: Please create issues in this repo for these Todo comments and the ones above

parthea (approved):

amanda-tarafa (approved):
LGTM!

gapic/configurable_snippetgen/configured_snippet.py
dizcology: Done.

PR #1546, opened 2022-12-14
chore: add service client initialization

dizcology:
Add service client initialization call to sample function's body, but without `client_options` (for setting `api_endpoint`), which will be added in the next PR.

This PR also includes the main helper method `_append_to_sample_function_def_body` for appending statements to the sample function.

### DIFF
--- a/tests/unit/configurable_snippetgen/test_configured_snippet.py
+++ b/tests/unit/configurable_snippetgen/test_configured_snippet.py
@@ -17,6 +17,7 @@ from pathlib import Path
 
 from google.protobuf import json_format
 from google.protobuf.compiler import plugin_pb2
+import libcst
 import pytest
 
 from gapic import utils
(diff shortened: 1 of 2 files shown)


amanda-tarafa (approved):
LGTM

tests/unit/configurable_snippetgen/test_configured_snippet.py
```diff
@@ -81,21 +82,66 @@ def snippet():
...
+    snippet = _make_configured_snippet(
+        SPEECH_V1_REQUEST_PATH, CONFIG_JSON_PATH, api_version="v1", is_sync=is_sync
     )
+    assert snippet.filename == expected
+
+
+def test_AppendToSampleFunctionBody():
+    # Start with a function def with nonempty body to we can be sure the
+    # transformer appends the statement.
+    function_def = libcst.parse_statement("def f():\n    'hello'")
```
quartzmo, line 138: Just curious, why does `'hello'` end up in this libcst statement? Shouldn't it be in the next one?

quartzmo (approved):

tests/unit/configurable_snippetgen/test_configured_snippet.py
dizcology: The behavior I hope to verify with this test is that the new statement is appended to the end of the function body (as opposed to, say, _prepended_), so the original function def has `'hello'` as placeholder representing nonempty existing statements.

atulep (approved):

PR #1547, opened 2022-12-25
chore: add client_options to service client initialization

dizcology:
Add `client_options` to service client initialization.

The signature of the helper function `convert_py_dict` intentionally does **not** accept a Python dict, to separate (A) data structures used internally in the generator code, and (B) pieces of data that need to be assembled into the code snippet.

Additional change:
- Remove a layer of method call "add to sample function", which is not necessary.


gapic/configurable_snippetgen/configured_snippet.py
```diff
@@ -119,6 +120,23 @@ def filename(self) -> str:
         sync_or_async = "sync" if self.is_sync else "async"
         return f"{self.gapic_module_name}_generated_{service_name}_{snake_case_rpc_name}_{config_id}_{sync_or_async}.py"
 
+    @property
+    def api_endpoint(self) -> Optional[str]:
```
amanda-tarafa, line 124: If the schema and port are set, you should not ignore them. Or is it that those are not configurable in Python GAPICs.

dizcology: I feel I haven't seen the schema or the port used in Python.  Could you point me to some examples (in any language) where this is configured?

dizcology: Added handling of port.  The schema is not required for Python GAPIC libraries ("https" is added internally unless the caller specified "http"), for snippets I think it is safe to ignore that.

parthea (approved):

gapic/configurable_snippetgen/configured_snippet.py
amanda-tarafa: Oh, sorry for missing your earlier reply.

The thing is that if a sample producer specified "http" then you'd be generating the wrong sample (same as with the port). They may be specifiying any of that, for instance, for producing a sample that uses an emulator.

How I'm doing this is for .NET is, I check each of the parts, for the ones that have value different from default, I use those values, for the ones that don't I let the library handle with defaults, etc.

dizcology: Thanks - I clicked merge before seeing your message here.  It makes sense to respect the config (even if from the perspective of public single method code snippet, I don't see why http would be used).  I will add this in a future PR: https://github.com/googleapis/gapic-generator-python/issues/1565

PR #1561, opened 2023-01-16
docs: sync README with synthtool templates

parthea:
This PR updates the README.md in gapic-generator-python to match the templated README in synthtool [here](https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/python_library/README.rst).

This PR also adds the ability to read and override `api_description`, `documentation_uri` and `title` from the service yaml file.

### DIFF
--- a/gapic/schema/naming.py
+++ b/gapic/schema/naming.py
@@ -37,11 +37,14 @@ class Naming(abc.ABC):
     An concrete child of this object is made available to every template
     (as ``api.naming``).
     """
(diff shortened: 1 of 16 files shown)


vchudnov-g:
AFAICT, this doesn't change the `README.rst` in `gapic-generator-python`, but just aims to change the `README`s for generated APIs. Is that right? If so, we should edit the PR description for accuracy.

Filed issue #1641 to track updating the `README` in this repo.

PR #1566, opened 2023-01-18
chore: support url schema in api endpoint

### LINKED ISSUE #1565: Generated snippet should use the correct url schema
The current implementation ignores the schema part of the snippet configuration: https://github.com/googleapis/gapic-generator-python/blob/main/gapic/configurable_snippetgen/configured_snippet.py#L133

dizcology:
Fixes https://github.com/googleapis/gapic-generator-python/issues/1565


gapic/configurable_snippetgen/configured_snippet.py
```diff
@@ -141,9 +141,24 @@ def api_endpoint(self) -> Optional[str]:
...
+            host_maybe_with_port_and_region = host_maybe_with_port
+
+        if (
+            schema
+            == snippet_config_language_pb2.Snippet.ClientInitialization.ServiceEndpoint.HTTP
+        ):
+            return f"http://{host_maybe_with_port_and_region}"
+        elif (
+            schema
+            == snippet_config_language_pb2.Snippet.ClientInitialization.ServiceEndpoint.HTTPS
```
amanda-tarafa, line 155: If HTTPS is used in python by default, I think you can treat HTTPS the same as language default.  That's what I'm doing in .NET.

This was included because some languages require the schema to be specified.

dizcology: The generated Python library would inject https if not specified in client_options, and it seems the common convention for Python to pass in the endpoint without the schema: https://googleapis.dev/python/google-api-core/latest/client_options.html

This does affect the shape of the generated snippet, so I am interpreting the config as:
1. if the config explicitly specifies http(s), then explicitly include it in the generated snippet.
2. if the config does not explicitly specify http(s), go with the more familiar no-schema api_endpoint in the generated snippet.

atulep (approved):

gapic/configurable_snippetgen/configured_snippet.py
amanda-tarafa: The thing is HTTPS may have been specified on the config only for the benefit of languages that require the schema to be specified. So, it will end up in say, Python and C# samples needlessly, and will give users the impression that they always need to specify it in those languages, where that's not the case (in C# libraries, same as python, specifying the endpoint without the schema is the norm).
Basically, in C# I'm only setting the schema explicitly if it is HTTP, as anything else means the schema is HTTPS and that's the default for C# anyways.
I'm doing the same for port, i.e. I only specify it if it's set to a value different than 443, when unset or set to 443, I'm not setting the port as 443 is the default for C#.

dizcology: UPDATE:  I just ran some quick tests with the aiplatform API.  Specifying the API endpoint as the following works fine:
- `dns:us-central1-aiplatform.googleapis.com`
- `us-central1-aiplatform.googleapis.com`

However using the following resulted in `google.api_core.exceptions.ServiceUnavailable: 503 DNS resolution failed for https:us-central1-aiplatform.googleapis.com: C-ares status is not ARES_SUCCESS qtype=AAAA name=https is_balancer=0: Domain name not found`:

- https:us-central1-aiplatform.googleapis.com
- https://us-central1-aiplatform.googleapis.com

In this sense the changes in this PR would actually result in broken snippets, prefixing with `https://`.

amanda-tarafa: Oh that's interesting. Still a point towards only specifying schema if it's set to HTTP, and leaving the library to do its thing otherwise. HTTP won't work with the real service, of course, the only usage I'm aware of is for products that have emulators, and possibly with TPC as well.
I'll run some tests with .NET AIPlatform library later and come back here with my findings.

dizcology: Thanks for the further clarification!  I updated the code accordingly, omitting the schema in both https and language default cases.

PR #1568, opened 2023-01-19
docs: add documentation for enums

### LINKED ISSUE #1567: Documentation for enums is missing
See the enum for `PartitionKey` in the golden file below. 
https://github.com/googleapis/gapic-generator-python/blob/6cc95de000f4ec74e49749fc7c87d854054c3ad2/tests/integration/goldens/asset/google/cloud/asset_v1/types/asset_service.py#L746-L755

The documentation for enums `PARTITION_KEY_UNSPECIFIED`, `READ_TIME` and `REQUEST_TIME` [appear in the proto file ](https://github.com/googleapis/googleapis/blob/9f04c988778321376f6d9612cf0d2c101cce8d4c/google/cloud/asset/v1/asset_service.proto#LL770-L785C4) but not in the generated code.

```
  enum PartitionKey {
    // Unspecified partition key. If used, it means using non-partitioned table.
    PARTITION_KEY_UNSPECIFIED = 0;

    // The time when the snapshot is taken. If specified as partition key, the
    // result table(s) is partitoned by the additional timestamp column,
    // readTime. If [read_time] in ExportAssetsRequest is specified, the
    // readTime column's value will be the same as it. Otherwise, its value will
    // be the current time that is used to take the snapshot.
    READ_TIME = 1;

    // The time when the request is received and started to be processed. If
    // specified as partition key, the result table(s) is partitoned by the
    // requestTime column, an additional timestamp column representing when the
    // request was received.
    REQUEST_TIME = 2;
  }
```

parthea:
Fixes https://github.com/googleapis/gapic-generator-python/issues/1567

### DIFF
--- a/gapic/templates/%namespace/%name_%version/%sub/types/_enum.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/types/_enum.py.j2
@@ -1,5 +1,12 @@
 class {{ enum.name }}({{ p }}.Enum):
-    r"""{{ enum.meta.doc|rst(indent=4) }}"""
+    r"""{{ enum.meta.doc|rst(indent=4) }}
+
+    Values:
+    {% for enum_value in enum.values %}
+        {{ enum_value.name }}:
+            {{ enum_value.meta.doc|rst(width=72, indent=12, nl=False) }}
+    {% endfor %}
+    """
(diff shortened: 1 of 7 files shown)


dizcology (approved):
Looks great!

Issue #835, opened 2021-04-06
Addition of `policy` and `options` to reserved names in #824 results in breaking changes to stable libraries


busunkim96:
These APIs have field renames resulting from this #824 which expanded the `RESERVED_NAMES` list. Some of them are GA, and even for beta APIs I'd prefer not to take changes that looks spurious from the user's perspective. 

https://github.com/googleapis/python-artifact-registry/pull/12
https://github.com/googleapis/python-cloudbuild/pull/81
https://github.com/googleapis/python-iot/pull/87
https://github.com/googleapis/python-functions/pull/45
https://github.com/googleapis/python-securitycenter/pull/120
https://github.com/googleapis/python-bigquery-connection/pull/57
https://github.com/googleapis/python-containeranalysis/pull/112
https://github.com/googleapis/python-bigtable/pull/280
https://github.com/googleapis/python-tasks/pull/85
https://github.com/googleapis/python-policy-troubleshooter/pull/8
https://github.com/googleapis/python-asset/pull/171
https://github.com/googleapis/python-org-policy/pull/32
https://github.com/googleapis/python-binary-authorization/pull/18
https://github.com/googleapis/python-spanner/pull/303
https://github.com/googleapis/python-secret-manager/pull/100/
https://github.com/googleapis/python-firestore/pull/336
https://github.com/googleapis/python-billing/pull/55
https://github.com/googleapis/python-datacatalog/pull/143

`policy` and `options` are the two that show up in these APIs.

From looking at the regen, it doesn't look like these need to be added to the reserved names list? https://github.com/googleapis/gapic-generator-python/commit/f5e9f367ec6a72b4272f559a93f6fbb3d7e54b8b seems to result in the imports being renamed to not conflict. I see `from google.iam.v1 import policy_pb2 as giv_policy` ([example](https://github.com/googleapis/python-bigtable/pull/280/files#diff-aa015723229f8d7fd386f79dab53254d30ea8e10c880a6d12627f4d60ec3441fR42)) instead of `from google.iam.v1 import policy_pb2 as policy`.

Can `policy` and `options` be removed from the list? @software-dov

software-dov:
Options cannot, policy may be. @parthea can chime in for context, I think he's working with the APIs that require these names.

parthea:
The issue I encountered was with keyword `auth` in API `servicecontrol` `v1`. We could certainly limit the impact of this change if we only add `auth` as a reserved name.

[...]

PR #1575, opened 2023-01-23
feat: freezes reserved names list.

atulep:
This PR implements a solution proposed to reduce number of breaking changes occurring in downstream GAPIC libraries. It addresses the common denominator between https://github.com/googleapis/gapic-generator-python/issues/835, https://github.com/googleapis/python-video-transcoder/pull/121#discussion_r780659412 and https://github.com/googleapis/python-dataflow-client/pull/143.

Concretely, this PR freezes the list of reserved names. The list values come from a careful analysis on Google APIs. Specifically, the old list was filtered based on its usage in Google APIs protos: if the value was used in any of the API protos, that value was retained in the new list. This ensures backwards compatibility.


atulep:
@parthea PTAL

gapic/utils/reserved_names.py
```diff
@@ -12,22 +12,81 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import builtins
-import itertools
-import keyword
```
parthea, line 17: Please could you also check other usages of keywords in the code?

https://github.com/googleapis/gapic-generator-python/blob/13b6c6c1da9b75c171cc53167cd048aba2dc9481/gapic/samplegen/samplegen.py#L44-L47

https://github.com/googleapis/gapic-generator-python/blob/13b6c6c1da9b75c171cc53167cd048aba2dc9481/gapic/schema/api.py#L269-L271

https://github.com/googleapis/gapic-generator-python/blob/13b6c6c1da9b75c171cc53167cd048aba2dc9481/gapic/schema/wrappers.py#L1141-L1146

atulep: Thanks for the find.

* [gapic-generator-python/gapic/schema/api.py](https://github.com/googleapis/gapic-generator-python/blob/13b6c6c1da9b75c171cc53167cd048aba2dc9481/gapic/schema/api.py#L269-L271)
This is not related to protoplus reserved names. The keyword list is included to avoid syntax errors from Python and "metadata" etc are added because these are "reserved" inside of client methods. For example, "request" is hardcoded in templates as an argument to all client methods.
* [gapic-generator-python/gapic/schema/wrappers.py](https://github.com/googleapis/gapic-generator-python/blob/13b6c6c1da9b75c171cc53167cd048aba2dc9481/gapic/schema/wrappers.py#L1141-L1146) this seems to only deal with Python keywords for the same reason as above. Again, unrelated to protoplus reserved names.

`samplegen` looks legit. Looking into it.

atulep: `RESERVED_WORDS` in samplegen also doesn't relate to protoplus. Their list is used in https://github.com/googleapis/gapic-generator-python/blob/13b6c6c1da9b75c171cc53167cd048aba2dc9481/gapic/samplegen/samplegen.py#L710-L720, where they try to generate sample code, and want to avoid invalid Python. Actually, their list is also too big (e.g. builtins not need to be there), but it's a separate problem.

atulep:
Spoke with Tony, and he wanted to run some internal diagnostics before LGTM, but got busy with other tasks. I'll run them on his behalf.

atulep:
There is no diff between generated client libs with these changes applied and the main branch.

parthea (approved):

PR #1576, opened 2023-01-26
fix: fix install issue for packages without google namespace

parthea:
Fixes b/263306233

### DIFF
--- a/tests/integration/goldens/credentials/setup.py
+++ b/tests/integration/goldens/credentials/setup.py
@@ -55,9 +55,7 @@ packages = [
     if package.startswith("google")
 ]
 
-namespaces = ["google"]
-if "google.cloud" in packages:
-    namespaces.append("google.cloud")
+namespaces = ["google", "google.iam"]
 
 setuptools.setup(
     name=name,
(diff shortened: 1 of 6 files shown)


dizcology (approved):

tests/integration/goldens/credentials/setup.py
```diff
@@ -55,9 +55,7 @@
     if package.startswith("google")
 ]
 
-namespaces = ["google"]
-if "google.cloud" in packages:
-    namespaces.append("google.cloud")
+namespaces = ["google", "google.iam"]
```
vchudnov-g, line 58: I don't quite grok everything this PR implies, but here we used to be adding `"google.cloud"` but now you're adding `"google.iam"`. Is this intentional?

parthea: `google.iam` is the correct namespace as the source lives in [google/iam](https://github.com/googleapis/googleapis/blob/master/google/iam). Historically, we have used the `python-gapic-namespace` option to override the default namespace when the API is part of cloud
https://github.com/googleapis/googleapis/blob/master/google/iam/v2/BUILD.bazel#L181. I'll push a commit to add that option.

vchudnov-g: OK, thanks for clarifying.

PR #1582, opened 2023-02-06
fix: fix generated unit tests with repeated double

### LINKED ISSUE #1581: Generated unit tests fail with repeated double in response
https://github.com/googleapis/gapic-generator-python/pull/1541 fixed an issue with a float comparison in a unit test however the fix does not work for repeated fields.

An example of the error is 
```
>       assert math.isclose(response.repeatedfield, [0.1809], rel_tol=1e-6)
E       TypeError: must be real number, not Repeated
```

I can recreate the issue by adding a fragment test with a `repeated double` field in the response.
```
=================================== FAILURES ===================================
________________________ test_my_method[MethodRequest] _________________________

request_type = <class 'google.fragment.types.test_response_has_repeated_double.MethodRequest'>
transport = 'grpc'

    @pytest.mark.parametrize("request_type", [
      test_response_has_repeated_double.MethodRequest,
      dict,
    ])
    def test_my_method(request_type, transport: str = 'grpc'):
        client = MyServiceClient(
            transport=transport,
        )
    
        # Everything is optional in proto3 as far as the runtime is concerned,
        # and we are mocking out the actual API, so just send an empty request.
        request = request_type()
    
        with mock.patch.object(
                type(client.transport.my_method),
                '__call__') as call:
            # Designate an appropriate return value for the call.
            call.return_value = test_response_has_repeated_double.MethodResponse(
                result=[0.671],
            )
            response = client.my_method(request)
    
            # Establish that the underlying gRPC stub method was called.
            assert len(call.mock_calls) == 1
            _, args, _ = call.mock_calls[0]
            assert args[0] == test_response_has_repeated_double.MethodRequest()
    
        # Establish that the response is the type that we expect.
        assert isinstance(response, test_response_has_repeated_double.MethodResponse)
>       assert math.isclose(response.result, [0.671], rel_tol=1e-6)
E       TypeError: must be real number, not Repeated

/tmp/tmpq09qwuah/tests/unit/gapic/fragment/test_my_service.py:572: TypeError
```

parthea:
I added a new fragment test `test_repeated_double.proto` that I used to re-create the issue.

Fixes googleapis/gapic-generator-python#1581 🦕

### DIFF
--- a/gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
+++ b/gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
@@ -587,7 +587,16 @@ def test_{{ method_name }}(request_type, transport: str = 'grpc'):
     {% for field in method.output.fields.values() | rejectattr('message') %}
     {% if not field.oneof or field.proto3_optional %}
     {% if field.field_pb.type in [1, 2] %}{# Use approx eq for floats #}
+    {% if field.repeated %}
+    for index in range(len(response.{{ field.name }})):
+        assert math.isclose(
+            response.{{ field.name }}[index],
+            {{ field.mock_value }}[index],
+            rel_tol=1e-6,
+        )
+    {% else %}{# field.repeated #}
     assert math.isclose(response.{{ field.name }}, {{ field.mock_value }}, rel_tol=1e-6)
+    {% endif %}{# field.repeated #}
     {% elif field.field_pb.type == 8 %}{# Use 'is' for bools #}
     assert response.{{ field.name }} is {{ field.mock_value }}
     {% else %}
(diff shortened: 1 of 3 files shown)


gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -587,7 +587,16 @@ def test_{{ method_name }}(request_type, transport: str = 'grpc'):
...
     {% if not field.oneof or field.proto3_optional %}
     {% if field.field_pb.type in [1, 2] %}{# Use approx eq for floats #}
+    {% if field.repeated %}
+    for index in range(len(response.{{ field.name }})):
+        assert math.isclose(
+            response.{{ field.name }}[index],
+            {{ field.mock_value }}[index],
+            rel_tol=1e-6,
+        )
+    {% else %}{# field.repeated #}
```
dizcology, line 597: Is it feasible to refactor the same code?

dizcology (approved):
Looks good!  Does this change affect generated golden files?

gapic/ads-templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
parthea: This is a much larger issue as duplication exists throughout the templates. We should tackle the duplicate code as a while separate from this bug fix. I've created https://github.com/googleapis/gapic-generator-python/issues/1592 to track the issue.

parthea:
We won't see any change to the golden files as the golden APIs listed [here](https://github.com/googleapis/gapic-generator-python/tree/main/tests/integration/goldens) don't have a repeated double in the response.

PR #1596, opened 2023-02-14
feat: add generator option proto-plus-dep

### LINKED ISSUE #1611: Incorrect assumption in gapic-generator-python for dependencies
By default, python clients (gapics) are generated with [proto-plus](https://github.com/googleapis/proto-plus-python) types.

There is an incorrect assumption in the generator where it assumes that dependencies never have proto-plus types.

For example, see the comment below which shows the problem. All dependencies are treated as non-proto plus (with `_pb2`) extension. When gapics depend on other gapics, the generator produces broken code.

https://github.com/googleapis/gapic-generator-python/blob/0ef9e62a64457bc9ca710524cad3993a58721e0b/gapic/schema/metadata.py#L92-L96

The code for `is_external_type` is below
https://github.com/googleapis/gapic-generator-python/blob/0ef9e62a64457bc9ca710524cad3993a58721e0b/gapic/schema/metadata.py#L105-L107

Googlers see b/266837434

### LINKED ISSUE #10828: cannot run maps-fleetengine-delivery-v1
in the beta libary pip install https://storage.googleapis.com/fleetengine-gapic/dist/793/2022-07-19/maps-fleetengine-delivery-v1-py.tar.gz I am getting error as below :+1: 

ModuleNotFoundError: No module named 'google.geo.type'

when in install google-geo-type 0.3.0 module then I get the next error as 

ImportError: cannot import name 'viewport_pb2' from 'google.geo.type

Reference :
[https://developers.google.com/maps/documentation/transportation-logistics/last-mile-fleet-solution/fleet-performance/fleet-engine/gapic_client](https://developers.google.com/maps/documentation/transportation-logistics/last-mile-fleet-solution/fleet-performance/fleet-engine/gapic_client)

parthea:
fix: resolve ImportError when GAPIC depends on another GAPIC

Fixes b/266837434
Fixes https://github.com/googleapis/google-cloud-python/issues/10828
Fixes https://github.com/googleapis/gapic-generator-python/issues/1611

Googlers see go/fix-gapic-depends-on-gapic-with-proto-plus-types


gapic/schema/metadata.py
```diff
@@ -188,6 +192,28 @@ def python_import(self) -> imp.Import:
                 alias=self.module_alias,
             )
 
+        if self.is_proto_plus_type:
+            # We need to change the import statement to use an
+            # underscore between the module and the version. For example,
+            # change google.cloud.documentai.v1 to google.cloud.documentai_v1.
+            # Check if the package name contains a version.
+            version_regex = "^v\d(alpha|beta)?\d?"
```
vchudnov-g, line 200:
```suggestion
            version_regex = "^v\d+(alpha|beta)?\d*"
```

vchudnov-g: We could have double digit versions on either side of the preview labels.

Also, we have seen some more complicated version numbers like [these](https://github.com/googleapis/googleapis/tree/master/google/cloud/asset) (v1p5beta1). So maybe the regex should be `"^v\d[^/]*$"` ? WDYT?

gapic/templates/setup.py.j2
```diff
@@ -33,10 +33,16 @@ dependencies = [
     "protobuf>=3.19.5,<5.0.0dev,!=3.20.0,!=3.20.1,!=4.21.0,!=4.21.1,!=4.21.2,!=4.21.3,!=4.21.4,!=4.21.5",
     {# TODO: Remove after https://github.com/googleapis/gapic-generator-python/pull/1240 is merged. #}
     {% if api.requires_package(('google', 'iam', 'v1')) or opts.add_iam_methods or api.has_iam_mixin %}
-        'grpc-google-iam-v1 >= 0.12.4, < 1.0.0dev',
+    "grpc-google-iam-v1 >= 0.12.4, < 1.0.0dev",
```
vchudnov-g, line 36: The TODO comment above this `{%if}` should be updated, since https://github.com/googleapis/gapic-generator-python/pull/1240 has been merged.

gapic/templates/setup.py.j2
```diff
@@ -33,10 +33,16 @@ dependencies = [
...
     {% endif %}
-    {% if api.requires_package(('google', 'cloud', 'documentai', 'v1')) %}
-        'google-cloud-documentai >= 1.2.1, < 3.0.0dev',
+    {% if api.requires_package(('google', 'cloud', 'documentai', 'v1')) and api.naming.warehouse_package_name != 'google-cloud-documentai' %}
```
vchudnov-g, line 38: Could you file an issue and refer to it in a TODO to have a more elegant way of handling these cases? It seems wrong to hard-code packages and versions in the template (brittle, technical debt).

gapic/templates/testing/constraints-3.7.txt.j2
```diff
@@ -13,3 +13,6 @@ grpc-google-iam-v1==0.12.4
 {% if api.requires_package(('google', 'cloud', 'documentai', 'v1')) %}
 google-cloud-documentai==1.2.1
 {% endif %}
+{% if api.requires_package(('google', 'cloud', 'kms', 'v1')) %}
```
vchudnov-g, line 16: Ditto for issue/TODO here.

gapic/schema/metadata.py
```diff
@@ -188,6 +192,28 @@ def python_import(self) -> imp.Import:
...
+            regex_match = re.match(version_regex, self.package[-1])
+
+            if regex_match:
```
vchudnov-g, line 203: Do we have a test for this import regex rewriting? We should, for at least cases like `v1`, `v10`,`v2beta`,`v20beta``,v20p1`,`v20p1alpha3p5`, don't you think?

vchudnov-g (changes requested):
Don't know much about proto-plus so I can't comment on those aspects of it, but otherwise this looks good.

Two important issues: making the regex more robust, and capturing an item to reduce the hard-coding tech debt for specific API×version.

gapic/schema/metadata.py
parthea: Fixed

gapic/templates/setup.py.j2
parthea: Fixed

gapic/templates/setup.py.j2
parthea: I've moved these cases to a separate file and extracted the common code.

gapic/templates/testing/constraints-3.7.txt.j2
parthea: Fixed

gapic/schema/metadata.py
parthea: Done

gapic/templates/setup.py.j2
```diff
@@ -31,13 +31,14 @@ dependencies = [
     "proto-plus >= 1.22.0, <2.0.0dev",
     "proto-plus >= 1.22.2, <2.0.0dev; python_version>='3.11'",
     "protobuf>=3.19.5,<5.0.0dev,!=3.20.0,!=3.20.1,!=4.21.0,!=4.21.1,!=4.21.2,!=4.21.3,!=4.21.4,!=4.21.5",
-    {# TODO: Remove after https://github.com/googleapis/gapic-generator-python/pull/1240 is merged. #}
-    {% if api.requires_package(('google', 'iam', 'v1')) or opts.add_iam_methods or api.has_iam_mixin %}
-        'grpc-google-iam-v1 >= 0.12.4, < 1.0.0dev',
+    {% for package_tuple, package_info in pypi_packages.items() %}
+    {# Quick check to make sure the package is different from this setup.py #}
```
vchudnov-g, line 35: Could you clarify `the package is different from this setup.py`? This to check we're not processing specifying a version range for the GAPIC we're generating itself, only for others we may depend on?

vchudnov-g (approved):
One test suggestion and one question, but otherwise LGTM

gapic/templates/setup.py.j2
parthea: Correct. There is an existing test that shows the `requires_package` [method](https://github.com/googleapis/gapic-generator-python/blob/13b6c6c1da9b75c171cc53167cd048aba2dc9481/gapic/schema/api.py#L487) will return true for the API that is currently being generated.
https://github.com/googleapis/gapic-generator-python/blob/13b6c6c1da9b75c171cc53167cd048aba2dc9481/tests/unit/schema/test_api.py#L93-L111

parthea:
Waiting on approval in go/fix-gapic-depends-on-gapic-with-proto-plus-types

gapic/schema/metadata.py
```diff
@@ -103,8 +102,13 @@ def __str__(self) -> str:
         return '.'.join(self.parent + (self.name,))
 
     @property
-    def is_external_type(self):
-        return not self.proto_package.startswith(self.api_naming.proto_package)
+    def is_proto_plus_type(self):
+        if self.proto_package.startswith(self.api_naming.proto_package) or (
```
atulep, line 106: Simplify by returning the condition directly.

return if ....

gapic/schema/metadata.py
```diff
@@ -103,8 +102,13 @@ def __str__(self) -> str:
         return '.'.join(self.parent + (self.name,))
 
     @property
-    def is_external_type(self):
-        return not self.proto_package.startswith(self.api_naming.proto_package)
+    def is_proto_plus_type(self):
```
atulep, line 105: This is non-trivial function and will benefit from adding a docstring explaining what and why you are checking for these things. Please add.

gapic/schema/metadata.py
```diff
@@ -188,6 +192,29 @@ def python_import(self) -> imp.Import:
...
+        if self.is_proto_plus_type:
+            # We need to change the import statement to use an
+            # underscore between the module and the version. For example,
+            # change google.cloud.documentai.v1 to google.cloud.documentai_v1.
+            # Check if the package name contains a version.
+            version_regex = "^v\d[^/]*$"
+            regex_match = re.match(version_regex, self.package[-1])
+
+            if regex_match:
+                versioned_module = f"{self.package[-2]}_{regex_match[0]}"
```
atulep, line 204: Can a package have less than 2 elements? (so [-2] raise IndexError)

gapic/schema/metadata.py
parthea: Fixed in 
https://github.com/googleapis/gapic-generator-python/pull/1596/commits/99a1ac8da58713d758b508f5980f7a940d228773

gapic/schema/metadata.py
parthea: Done in https://github.com/googleapis/gapic-generator-python/pull/1596/commits/47a2067690019bde6e17e0025ced16325ec72692

gapic/schema/metadata.py
parthea: Good catch! Fixed in https://github.com/googleapis/gapic-generator-python/pull/1596/commits/3f95c0af570cd68db0dc634f79a5760573d3f406

PR #1605, opened 2023-02-18
add http annotation to fragment proto

parthea:
Updating fragment proto to reproduce error in https://github.com/googleapis/python-debugger-client/pull/151. 

See build log [here](https://github.com/googleapis/python-debugger-client/actions/runs/4129085698/jobs/7134304785).

From https://github.com/googleapis/python-debugger-client/pull/151,
```
E       ValueError: Invalid request.
E       Some of the fields of the request message are either not initialized or initialized with an invalid value.
E       Please make sure your request matches at least one accepted HTTP binding.
E       To match a binding the request message must have all the required fields initialized with values matching their patterns as listed below:
E       	URI: "/v2/debugger/debuggees/{debuggee_id}/breakpoints/set"
E       	Required request fields:
E       		field: "debuggee_id", pattern: "*"
```

### DIFF
--- a/tests/fragments/test_rest_streaming.proto
+++ b/tests/fragments/test_rest_streaming.proto
@@ -17,11 +17,16 @@ syntax = "proto3";
 package google.fragment;
 
 import "google/api/client.proto";
+import "google/api/annotations.proto";
 
 service MyService {
   option (google.api.default_host) = "my.example.com";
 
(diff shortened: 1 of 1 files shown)


parthea:
I've proposed a fix in https://github.com/googleapis/python-api-core/pull/489

PR #1608, opened 2023-02-23
fix: Resolve errors from annotating containers with non-local enums

### LINKED ISSUE #1607: annotations subscripting containers cause TypeErrors
As per an internal doc,  subscripting  a container that does not inherit from `typing.Generic` can cause errors. In particular, a `MutableSequence` indexed by  enums causes an error of the form
`TypeError: Parameters to generic types must be types`. This can be triggered, for example, by code of the form

```
        days_of_week: MutableSequence[dayofweek_pb2.DayOfWeek] = proto.RepeatedField(
            proto.ENUM,
            number=2,
            enum=dayofweek_pb2.DayOfWeek,
        )
```

which we are seeing in the internal part of our publishing pipeline.

Either of these fixes by itself works:
1. Quoting the type: `MutableSequence['dayofweek_pb2.DayOfWeek']`
2. Adding `from __future__ import annotations ` at the top of the file. As per an internal reference, this seems to be related to [PEP563](https://peps.python.org/pep-0563/), which has a status of `Accepted` as of 2017, so it's not clear why we would need a fix in the first place.

We should decide and implement the right fix.

vchudnov-g:
The key was `from __future__ import annotations`, implementing [PEP
563](https://www.python.org/dev/peps/pep-0563/), which, though approved in 2017, is now planned for inclusion in Python 3.11 (see this [Reddit discussion](https://www.reddit.com/r/Python/comments/muyz5h/pep_563_getting_rolled_back_from_python_310/) and [Python.org discussion](https://mail.python.org/archives/list/python-dev@python.org/thread/CLVXXPQ2T2LQ5MP2Y53VVQFCXYWQJHKZ/)).

Fixes #1607.


atulep (approved):

tests/fragments/google/type/dayofweek.proto
```diff
@@ -0,0 +1,50 @@
+// Copyright 2021 Google LLC
```
Mariatta, line 1:
```suggestion
// Copyright 2023 Google LLC
```

PR #1613, opened 2023-03-03
fix: fix bug with quote replacement.

atulep:
MixIn methods currently do `json.loads(json.dumps(transcoded_request["body"]))`. `json.loads` converts `"` into `'`, which is an invalid JSON. So when we pass it to the server later, we get a bad request error; i.e. [this issue](https://github.com/googleapis/python-pubsub/pull/863). 

Apparently, AuthorizedSession, used for sending the request, can accept JSON strings in the body parameter. We should remove json.loads from the templates.

I tested this change on pubsub and notebook clients.


vchudnov-g (changes requested):
- Could you add tests that fail without this change and pass with it?
- The Ads templates probably need a corresponding change. Could you implement that (and tests)?
- Do you recall why we had `json.loads` initially? Doing a bit of `git blame` ([example](https://github.com/googleapis/gapic-generator-python/blame/120f19eef34ca3d722f079683222be99931363da/gapic/ads-templates/%25namespace/%25name/%25version/%25sub/services/%25service/transports/rest.py.j2)), it seems we had this for query parameters. My vague recollection is that it did work there. Is that still the case? (I'm concerned that there may be a reason why we had this construct for the body, too, and I don't want to miss that.)

atulep:
> Could you add tests that fail without this change and pass with it?
This bug only occurs in generated code for services with REGAPIC + MixIns. It's hard to isolate this case with existing infra. We need an integration test that accepts service.yaml, and existing setup disallows that. 

> The Ads templates probably need a corresponding change. Could you implement that (and tests)?
Ads does not have MixIns in neither grpc nor rest.

> Do you recall why we had json.loads initially?
This was an oversight during initial implementation of MixIns in REGAPIC.

vchudnov-g:
> > Could you add tests that fail without this change and pass with it?
> 
> This bug only occurs in generated code for services with REGAPIC + MixIns. It's hard to isolate this case with existing infra. We need an integration test that accepts service.yaml, and existing setup disallows that.


I was thinking more of refactoring the transcoding into a (non-j2) function we could test in the generator unit tests. So for example the generated code would read something like this

```
import gapic_utils
...

body = gapic_utils.json_encode(transcoded_request['body'])
```

and `gapic_utils` would be a static (non-templated) file we always include in the GAPIC and contains:

```
def json_encode(input):
  return json.dumps(input)
```

and which we could unit-test in the generator, at a single point of truth.

I think that's worth doing in general...but it looks like we're not including such "static" files right now. Is that right? In that case, it's not worth blocking this PR on that. But if we are....

vchudnov-g (approved):
See my comment. If it's easy to add the function in a unit-testable way, let's. Otherwise, I'll file a separate issue for that and we can unblock this in the meantime.

Thanks for doing this work!

atulep:
This will require additional infra work not related to the bug at hand. I'll merge this as it blocks several releases. Ideally, we actually test this in Showcase.

PR #1618, opened 2023-03-10
fix: ignore_unknown_fields parsing service config

noahdietz:
Enables [`ignore_unknown_fields`](https://googleapis.dev/python/protobuf/latest/google/protobuf/json_format.html#google.protobuf.json_format.ParseDict) behavior while parsing the `google.api.Service` config YAML file. This will make the Python generator resilient to parsing a Service config with fields that it hasn't received the updated proto for yet.

cc: @jskeet @alexander-fenster

### DIFF
--- a/gapic/schema/api.py
+++ b/gapic/schema/api.py
@@ -341,7 +341,11 @@ class API:
 
         # Parse the google.api.Service proto from the service_yaml data.
         service_yaml_config = service_pb2.Service()
-        ParseDict(opts.service_yaml_config, service_yaml_config, True)
+        ParseDict(
+            opts.service_yaml_config,
+            service_yaml_config,
+            ignore_unknown_fields=True
+        )
 
         # Done; return the API.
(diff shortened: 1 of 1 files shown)


gapic/schema/api.py
```diff
@@ -341,7 +341,7 @@ def disambiguate_keyword_sanitize_fname(
 
         # Parse the google.api.Service proto from the service_yaml data.
         service_yaml_config = service_pb2.Service()
-        ParseDict(opts.service_yaml_config, service_yaml_config)
+        ParseDict(opts.service_yaml_config, service_yaml_config, True)
```
alexander-fenster, line 344: I think Python should allow named parameter here?
```suggestion
        ParseDict(opts.service_yaml_config, service_yaml_config, ignore_unknown_fields=True)
```

noahdietz:
@parthea @atulep is this OK to do?

parthea (approved):

PR #1632, opened 2023-03-24
[DRAFT] Add pooled transport class

daniel-sanche:
Fixes https://github.com/googleapis/gapic-generator-python/issues/1635

> Individual gRPC channels are limited to [100 concurrent streams](https://cloud.google.com/bigtable/docs/connection-pools#how_connection_pools_work). This puts a cap on max throughput, which may be run up against in an asyncio application. One way to get around this is to maintain a pool of channels, and rotate requests between them. This has been implemented in gapic in some languages (example: [go](https://github.com/googleapis/gapic-generator-go/blob/34124f14cdc18208b3bfcdf6602b3c8b25f9d21a/internal/gengapic/gengrpc.go#L268)), but not Python.

This PR adds a new PooledGrpcAsyncIOTransport template. Rather than maintaining a single grpc channel, it maintains a list of channels. Requests are rotated between channels in a round robin fashion. It also provides a `replace_channel` method, for expiring and swapping out a channel in the pool with a new one.

Update: April 3:
- I moved the pooling logic into a grpc.aio.Channel implementation, to match how they do it in Java. This points to a future direction where we don't need a separate Transport class for pooled channels, we can make it an argument for the existing grpc transport classes (maybe using a ClientOptions setting?). Let me know what you think
- For now, I intend to keep the changes isolated to a custom experimental transport though, to limit the scope

TODO:
- allow dynamic pool sizes?
- create a mechanism to only generate PooledGrpcAsyncIOTransport classes for clients that opt-in
- update documentation
- update golden files with changes


gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
dizcology, line 1: We will need to update the golden files too, for instance https://github.com/googleapis/gapic-generator-python/blob/main/tests/integration/goldens/asset/tests/unit/gapic/asset_v1/test_asset_service.py

This is typically done by running `bazel run //tests/integration:asset_update` for each service here: https://github.com/googleapis/gapic-generator-python/tree/main/tests/integration/goldens

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/pooled_grpc_asyncio.py.j2
```diff
@@ -0,0 +1,445 @@
...
+            client_cert_source_for_mtls: Optional[Callable[[], Tuple[bytes, bytes]]] = None,
+            quota_project_id: Optional[str] = None,
+            client_info: gapic_v1.client_info.ClientInfo = DEFAULT_CLIENT_INFO,
+            always_use_jwt_access: Optional[bool] = False,
+            api_audience: Optional[str] = None,
+            ) -> None:
+        """Instantiate the transport.
+
+        Args:
+            pool_size (int): the number of grpc channels to maintain in a pool
```
dizcology, line 131: Capitalize "the" and add period at the end to be consistent with the rest of the docstring.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/pooled_grpc_asyncio.py.j2
```diff
@@ -0,0 +1,445 @@
...
+            host=host,
+            credentials=credentials,
+            credentials_file=credentials_file,
+            scopes=scopes,
+            quota_project_id=quota_project_id,
+            client_info=client_info,
+            always_use_jwt_access=always_use_jwt_access,
+            api_audience=api_audience,
+        )
+        self._quota_project_id = quota_project_id
```
dizcology, line 222: Most of the code in `__init__` above this line seems to be the same as that in the regular (non-pooled) `GrpcAsyncIOTransport`.  Does this suggest the possibility for the pool transport to be a subclass of the regular one?

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/pooled_grpc_asyncio.py.j2
```diff
@@ -0,0 +1,445 @@
...
+        self._prep_wrapped_messages(client_info)
+
+    def next_channel(self) -> aio.Channel:
+        """Returns the next channel in the round robin pool."""
+        # Return the channel from cache.
+        channel = self.channel_pool[self._next_idx]
+        self._next_idx = (self._next_idx + 1) % len(self.channel_pool)
+        return channel
+
+    async def replace_channel(self, channel_idx, grace=None, new_channel=None) -> aio.Channel:
```
dizcology, line 252: Is this method to be called by the client internally, or by the user?

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/pooled_grpc_asyncio.py.j2
```diff
@@ -0,0 +1,445 @@
...
+
+        Args:
+          channel_idx(int): the channel index in the pool to replace
+          grace(Optional[float]): The time to wait until all active RPCs are 
+            finished. If a grace period is not specified (by passing None for
+            grace), all existing RPCs are cancelled immediately.
+          new_channel(grpc.aio.Channel): a new channel to insert into the pool
+            at `channel_idx`. If `None`, a new channel will be created.
+        """
+        if channel_idx >= len(self.channel_pool) or channel_idx < 0:
```
dizcology, line 267: Would it work well enough here if we just use `channel_idx % len(self.channel_pool)`?

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/pooled_grpc_asyncio.py.j2
```diff
@@ -0,0 +1,445 @@
...
+            self._operations_client = operations_v1.OperationsAsyncClient(
+                self.grpc_channel
+            )
+
+        # Return the client from cache.
+        return self._operations_client
+    {% endif %}
+    {% for method in service.methods.values() %}
+
+    def {{ method.transport_safe_name|snake_case }}(self, *args, **kwargs) -> Awaitable[{{ method.output.ident }}]:
```
dizcology, line 314: In the regular transport these are properties returning callables: https://github.com/googleapis/gapic-generator-python/blob/3bd2f8703e4a1a67e6a4c281890c4f99eda13fe7/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/transports/grpc_asyncio.py.j2#L260

But the return type here is Awaitable (while the docstring still says this returns Callable).

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/pooled_grpc_asyncio.py.j2
daniel-sanche: From the transport side, this will be user-exposed, but it will probably be client code calling this, not end users.

For BigTable, we need to call this to refresh channels before they expire, which causes a latency spike. We also need to pass in custom channels which have called cache warm-up RPCs, and will also likely use this to add emulator channels. 

I think if other libraries end up using pooling, they may also have other reasons for swapping channels in the pool

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/pooled_grpc_asyncio.py.j2
daniel-sanche: Yeah that could be an option. I'd be a bit worried about accidentally changing channels I didn't intent to though

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/pooled_grpc_asyncio.py.j2
daniel-sanche: I changed the implementation, so this is out of date, but if you're interested in the explaination, the reason for this is that I was running into issues with caching. 

The client would grab a grpc callable here once at init time, and use it to make future calls. But we wanted it to rotate calls between channels, so I needed to make sure it was actually calling my wrapper, and not just the unary callable directly. The signatures look different, but they should actually be the same, since I dropped the `@propery`. (A method with signature X should be functionally equivalent to a property that returns a function with signature X)

But TL;DR, it got a little ugly for my liking, and I moved the pooling implementation to a channel wrapper, which should make maintenance a bit easier

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/pooled_grpc_asyncio.py.j2
```diff
@@ -0,0 +1,445 @@
...
+                '/{{ '.'.join(method.meta.address.package) }}.{{ service.name }}/{{ method.name }}',
+                request_serializer={{ method.input.ident }}.{% if method.input.ident.python_import.module.endswith('_pb2') %}SerializeToString{% else %}serialize{% endif %},
+                response_deserializer={{ method.output.ident }}.{% if method.output.ident.python_import.module.endswith('_pb2') %}FromString{% else %}deserialize{% endif %},
+            )
+            self._stubs[stub_key] = stub_func
+        # call stub
+        return stub_func(*args, **kwargs)
+    {% endfor %}
+
+    {# TODO: Remove after https://github.com/googleapis/gapic-generator-python/pull/1240 is merged. #}
```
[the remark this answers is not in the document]
daniel-sanche: Yeah, for now I'm leaving it in place because (last I checked) this is still present in the other transport, and I wanted to minimize the diff during development. but I will make sure this is fixed before merging

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/pooled_grpc_asyncio.py.j2
daniel-sanche: Yeah, I'm still considering how best to go about that. The regular GrpcAsyncIOTransport will attempt to create its own grpc channel and wrapped messages, which I'd like to avoid.

I think eventually we can try to remove the pooled transport entirely (see update in description)

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/pooled_grpc_asyncio.py.j2
```diff
@@ -0,0 +1,378 @@
...
+class PooledMultiCallable():
+
+    def __init__(self, channel_pool:"PooledChannel", *args, **kwargs):
+        self._init_args = args
+        self._init_kwargs = kwargs
+        self.next_channel_fn = channel_pool.next_channel
+
+class PooledUnaryUnaryMultiCallable(PooledMultiCallable, aio.UnaryUnaryMultiCallable):
+    def __call__(self, *args, **kwargs) -> aio.UnaryUnaryCall:
+        return self.next_channel_fn().unary_unary(*self._init_args, **self._init_kwargs)(*args, **kwargs)
```
daniel-sanche, line 50: todo: add caching dict for unary calls for each channel, instead of re-creating each time

Issue #1644, opened 2023-04-13
Code coverage check failing in python-assured-workloads


parthea:
The code coverage for python-assured-workloads is 99% as can be see in https://github.com/googleapis/python-assured-workloads/actions/runs/4666486762/jobs/8261147131?pr=241.  We require 100% code coverage in the mono repository.

There are 2 issues that need to be addressed to bring code coverage to 100%.

1. Remove REST methods for rpcs which do not have http annotations. Such as [this one](https://github.com/googleapis/python-assured-workloads/blob/54877e4c61a6718e4442e88782992617e1c3d0ea/google/cloud/assuredworkloads_v1/services/assured_workloads_service/transports/rest.py#L470).

2. Ensure that code formatting is turned off when `pragma: no branch` is used  because `pragma: no branch` does not work across multiple lines. https://github.com/nedbat/coveragepy/issues/754

parthea:
Both issues have been addressed.

PR #1645, opened 2023-04-13
fix: remove REST methods for rpcs which do not have http annotations

parthea:
Towards https://github.com/googleapis/gapic-generator-python/issues/1644


vchudnov-g (approved):
Could you mention this PR in internal issue 262044440?

parthea:
Done

PR #1174, opened 2022-01-29
chore: track coverage in tests

parthea:
This PR is needed so we can avoid a repeat of the coverage issue introduced in https://github.com/googleapis/gapic-generator-python/pull/1159.

Issue #663, opened 2020-10-15
feat: add async client support for Ads templates


software-dov:
The Google Ads API uses a parallel set of templates to generate its client libraries. Features added to the default templates also should be added to the Ads templates to maintain a common interface for all Google GAPICs.

Feature request to add Async client support to the Ads templates.

PR #1646, opened 2023-04-13
fix: fix code coverage in async test

parthea:
Towards https://github.com/googleapis/gapic-generator-python/issues/1644

See https://github.com/googleapis/gapic-generator-python/pull/1174#issuecomment-1025132372 for background on why `# pragma: no branch` is used.

### DIFF
--- a/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
+++ b/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
@@ -848,6 +848,8 @@ async def test_{{ method_name }}_async_pages():
         {% endif %}
         )
         pages = []
+        # Workaround issue in python 3.9 related to code coverage by adding `# pragma: no branch`
+        # See https://github.com/googleapis/gapic-generator-python/pull/1174#issuecomment-1025132372
         # Turn off code formatting as `# pragma: no branch` does not work on multilines
         # fmt: off
         async for page_ in (await client.{{ method_name }}(request={})).pages: # pragma: no branch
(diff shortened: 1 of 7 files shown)


gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
```diff
@@ -848,8 +848,13 @@ async def test_{{ method_name }}_async_pages():
         {% endif %}
         )
         pages = []
+        # Workaround issue in python 3.9 related to code coverage by adding `# pragma: no branch`
+        # See https://github.com/googleapis/gapic-generator-python/pull/1174#issuecomment-1025132372
+        # Turn off code formatting as `# pragma: no branch` does not work on multilines
```
vchudnov-g, line 853: As per https://stackoverflow.com/a/54134770, it may work if you put it before the line break,
so maybe something like this would work?
```
async for page_ in ( #pragma no branch
   await client.{{ method_name }}(request={})).pages:
  pages.append(page_)
```

It would be nice if we could get this working instead, but if it's not quick to do right now, I'm OK filing a follow-up issue and proceeding with this change, since we're disabling formatting rather than coverage.

vchudnov-g (approved):
See note about a possible workaround, or filing a follow-up issue. LGTM otherwise.

vchudnov-g (commented):
Do we need to do this for ads templates as well?

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
parthea: Nice find! This worked.

parthea:
These changes are specific to async and Ads doesn't support async so no changes are needed for Ads. There is a separate issue to track adding async support for Ads. https://github.com/googleapis/gapic-generator-python/issues/663

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
parthea: Fixed in a8463aa

PR #1652, opened 2023-04-18
chore: set CODEOWNER to @googleapis/python-core-client-libraries

parthea:
Same as `google-cloud-python` [here](https://github.com/googleapis/google-cloud-python/blob/main/.github/CODEOWNERS)


.github/CODEOWNERS
```diff
@@ -1,8 +1,4 @@
 # Code owners file.
 # This file controls who is tagged for review for any given pull request.
 
-# For syntax help see:
-# https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
-
-* @googleapis/actools-python @googleapis/yoshi-python
-*.yaml @googleapis/yoshi-python @googleapis/actools-python
+*                         @googleapis/python-core-client-libraries
```
vchudnov-g, line 4: This team does not have write access to the repository, so we're getting a GitHub warning on this file.

The team settings are [here](https://github.com/orgs/googleapis/teams/python-core-client-libraries/repositories), but I don't see how to add a new repo. Maybe I don't have the needed permissions?

@parthea

vchudnov-g (changes requested):

.github/CODEOWNERS
parthea: Good catch! I granted write permission to the team via the Github Setting UI (Settings->Collaborators and Teams)

.kokoro/release/common.cfg
```diff
@@ -38,12 +38,3 @@ env_vars: {
   key: "SECRET_MANAGER_KEYS"
   value: "releasetool-publish-reporter-app,releasetool-publish-reporter-googleapis-installation,releasetool-publish-reporter-pem"
 }
-
-# Store the packages we uploaded to PyPI.  That way, we have a record of exactly
```
vchudnov-g: Is this removal intentional? I'm asking because the comment about SBOMs and attestations seems important.

parthea: It looks like this change was from a commit made by owlbot (https://github.com/googleapis/gapic-generator-python/pull/1652/commits/b2f902db1f4facce01bfa68627735b976bb7fcc5), which was followed by another commit from owlbot (https://github.com/googleapis/gapic-generator-python/pull/1652/commits/3d78cb8396e86bc1db4f4d20049b35451b9ca5a3) which reverted the change. Closing as obsolete.

vchudnov-g (approved):

PR #1665, opened 2023-05-29
chore: set initial gapic version to 0.0.0

parthea:


vchudnov-g (changes requested):
The internal doc makes it seem that release-please is looking at the current "default version value" of 0.1.0, so I think we should do the following:
- in this PR, link to the release-please changes that make the default version value 0.0.0 (or fix the design doc otherwise)
- mention in the design doc that one of the changes we're going to make is changing the default version value to 0.0.0, and why

vchudnov-g (approved):
Per offline conversation, there needs to be a version change to trigger the release job, so I'm approving to unblock. We should still describe this clearly in the design/info doc.

PR #1671, opened 2023-06-12
Update Ads templates to set minimum Python version to 3.8.

BenRKarl:
This is in preparation for the deprecation of Python 3.7, which will happen before our next release so we will want 3.8 set as the minimum next time we're generating code.

### DIFF
--- a/gapic/ads-templates/%namespace/%name/%version/__init__.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/__init__.py.j2
@@ -6,8 +6,8 @@ import importlib
 import sys
 
 
-if sys.version_info < (3, 7):
-    raise ImportError('This module requires Python 3.7 or later.')
+if sys.version_info < (3, 8):
+    raise ImportError('This module requires Python 3.8 or later.')
 
 
 _lazy_type_to_package_map = {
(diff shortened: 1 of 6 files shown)


gapic/ads-templates/%namespace/%name/%version/__init__.py.j2
```diff
@@ -6,8 +6,8 @@ import importlib
 import sys
 
 
-if sys.version_info < (3, 7):
-    raise ImportError('This module requires Python 3.7 or later.')
+if sys.version_info < (3, 8):
+    raise ImportError('This module requires Python 3.8 or later.')
```
BenRKarl, line 10: It might be more user-friendly to raise a warning instead of an error here.

parthea (approved):

gapic/ads-templates/%namespace/%name/%version/__init__.py.j2
parthea: In setup.py, we require python 3.8 so most folks on python 3.7 won't be able to install the library.

BenRKarl: We don't use the setup.py from the generator, but we also plan to set the minimum to 3.8 in our library, so 3.7 users will probably be blocked on install as well.

vchudnov-g:
Is there a way we can get the language version from a single place rather than have to hard-code repeatedly?  Not saying we need to block this PR, but it would be be easier, more elegant, and more solid to only have to change this in one place in the future and have all other uses derive from it.

BenRKarl:
Closing this because I'm going to put the deprecation warning in a different location.

PR #1682, opened 2023-07-04
fix: add `help` as a reserved word

### LINKED ISSUE #1681: `help` should be added as a reserved word to avoid a breaking change in compute
In PR https://github.com/googleapis/gapic-generator-python/pull/1575, `help` was removed from the reserved names list. Removing it from the list would result in a breaking change for `compute`. See `help_` [here](https://raw.githubusercontent.com/googleapis/python-compute/main/google/cloud/compute_v1/types/compute.py) which would change to `help` if we don't have `help` as a reserved term . 

See the breaking change in https://github.com/googleapis/python-compute/pull/417/files

parthea:
`help` was removed in #1575 

Fixes #1681 🦕

### DIFF
--- a/gapic/utils/reserved_names.py
+++ b/gapic/utils/reserved_names.py
@@ -87,6 +87,7 @@ RESERVED_NAMES = frozenset(
         "try",
         "type",
         "exec",
+        "help",
         # Comes from Protoplus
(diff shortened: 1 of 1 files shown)


chalmerlowe (approved):
LGTM

PR #1699, opened 2023-07-17
feat: allow Callables for transport and channel init

daniel-sanche:
Our gapic libraries support passing in custom [transport](https://github.com/googleapis/python-bigtable/blob/7f747e10d5c4182d95327321872b4d78519de818/google/cloud/bigtable_v2/services/bigtable/client.py#L368) and [channel](https://github.com/googleapis/python-bigtable/blob/7f747e10d5c4182d95327321872b4d78519de818/google/cloud/bigtable_v2/services/bigtable/transports/grpc.py#L54) objects when initializing clients and transports respectively. 

In theory, this allows full customization of the transport stack to support uncommon or experimental use-cases, like channel pooling. In practice, customizing transports and channels in this way is messy, because the user needs to provide a set of parameters when constructing the transport or channel. These parameters depend on complex initialization logic (e.g., local credential parsing), which we already perform when a transport or channel is not provided. The user should not have to duplicate that code.

This change addresses the issue by allowing the user to provide a Callable in place of a custom transport or channel instance. The Callable will receive all the arguments that are already computed in the generator when the transport or channel is not provided, and is expected to return a usable instance (subclass of XXX or YYY).

We recommend that henceforth users generally provide such a callable rather than already instantiated transport or channel objects.

---

**Example**:

Instead of:
```
client_options = client_options_lib.ClientOptions()
api_endpoint, client_cert_source_func = self.get_mtls_endpoint_and_cert_source(client_options)
api_key_value = getattr(client_options, "api_key", None)

custom_transport = MyCustomTransport(
  credentials=None,
  credentials_file=client_options.credentials_file,
  host=api_endpoint,
  scopes=client_options.scopes,
  client_cert_source_for_mtls=client_cert_source_func,
  quota_project_id=client_options.quota_project_id,
  always_use_jwt_access=True,
  api_audience=client_options.api_audience,
)
client = BigtableAsyncClient(transport=custom_transport)
```

We can do:
```
client = BigtableAsyncClient(transport=MyCustomTransport)
```

or:
```
client = BigtableAsyncClient(transport=lambda *args, **kwargs: MyCustomTransport(*args,  pool_size=5, **kwargs))
```

---

Let me know if you have any feedback. If we decide to go forward with his, I can add some tests and polish


parthea:
Converting to draft as tests are failing. Please convert back to review once tests are passing.

daniel-sanche:
> Converting to draft as tests are failing. Please convert back to review once tests are passing.

I have most of the tests passing, but I am having trouble interpreting the last "tests / integration". Any pointers?

daniel-sanche:
tests are now passing

tests/integration/goldens/credentials/google/iam/credentials_v1/services/iam_credentials/client.py
```diff
@@ -376,8 +377,12 @@ def __init__(self, *,
...
                 credentials = google.auth._default.get_api_key_credentials(api_key_value)
 
-            Transport = type(self).get_transport_class(transport)
-            self._transport = Transport(
+            transport_init: Union[Type[IAMCredentialsTransport], Callable[..., IAMCredentialsTransport]] = (
+                type(self).get_transport_class(transport)
+                if isinstance(transport, str) or transport is None
+                else transport
+            )
+            self._transport = transport_init(
```
vchudnov-g, line 385: Here were invoking the callable with a specific set of arguments. The documentation above simply says "with the set of initialization arguments". Where are these arguments listed/described for the user? Maybe we should amend the parameters to say "the same set of initialization arguments as the provided transport classes [LINK]"

WDYT?

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
```diff
@@ -173,7 +175,8 @@ class {{ service.name }}GrpcTransport({{ service.name }}Transport):
         )
 
         if not self._grpc_channel:
-            self._grpc_channel = type(self).create_channel(
+            channel_init = channel or type(self).create_channel
```
vchudnov-g, line 178: nit: Here and in similar places below, I would consider adding a comment `# initialize with the provided callable or the default channel`. That would make it more immediately clear what's going on here.

vchudnov-g (commented):
The functional changes appear good. I have one non-trivial in-code documentation suggestion.

Also, is this part of a greater initiative described in an internal doc? I couldn't find it in the one I was looking at. (Not that this needs lengthy documentation; it's straightforward enough that a one-liner would do, but it would be good to document why we need this.)

tests/integration/goldens/credentials/google/iam/credentials_v1/services/iam_credentials/client.py
daniel-sanche: Yeah makes sense, although I'm not really sure what a proper link would look like in this case. I improved the docstrings, let me know if that looks better

parthea:
Marking as draft as presubmits are failing

daniel-sanche:
fixed presubmits

parthea (approved):

parthea:
LGTM, but please wait for review from @vchudnov-g

parthea:
@vchudnov-g Please can you take a look?

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -444,9 +444,11 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
                 credentials identify the application to the service; if none
                 are specified, the client will attempt to ascertain the
                 credentials from the environment.
-            transport (Union[str, {{ service.name }}Transport]): The
-                transport to use. If set to None, a transport is chosen
-                automatically.
+            transport (Optional[Union[str,{{ service.name }}Transport,Callable[..., {{ service.name }}Transport]]]):
+                The transport to use, or a Callable that constructs and returns a new transport.
+                If a Callable is given, it will be called with the same set of initialization
+                arguments as used in the {{ service.name }}Transport constructor.
```
vchudnov-g, line 450: Can we hyperlink this somehow? Ditto elsewhere, and ditto for the channels.

vchudnov-g (approved):
Could you expand the PR description to include something like the following, if it is accurate? This is just to be very clear in the PR description and the commit message.

===BEGIN

... In practice, customizing `transport`s and `channel`s in this way is messy, because the user needs to provide a set of parameters when constructing the transport or channel. These parameters depend on  complex initialization logic (e.g., local credential parsing), which we already perform when a `transport` or `channel` is not provided. The user should not have to duplicate that code.

This change addresses the issue by allowing the user to provide a `Callable` in place of a custom `transport` or `channel` instance. The Callable will receive all the arguments that are already computed in the generator when the `transport` or `channel` is not provided, and is expected to return a usable instance (subclass of XXX or YYY).

We recommend that henceforth users generally provide such a callable rather than already instantiated `transport` or `channel` objects.

=== END

I wonder whether we should incorporate that last paragraph into code comments somehow. Maybe mention the callable is preferred? Maybe as a usage example in the `__init__` docstring?

PR #1705, opened 2023-07-28
fix: resolve docs issue when API has a service called services

### LINKED ISSUE #825: Error during client generation `self referenced toctree found` due to service named `services`
The `appengine_admin` client has a service called `services` . Unfortunately, `services` is already an implicitly reserved keyword in `gapic-generator-python` templates here:
https://github.com/googleapis/gapic-generator-python/blob/master/gapic/templates/docs/%25name_%25version/services.rst.j2

As a result of having `service.rst` referencing another `service.rst`, I get a Sphinx error `self referenced toctree found` when generating the client.

```
sphinx.errors.SphinxWarning: /usr/local/google/home/partheniou/git/GOB/python-appengine-admin/docs/appengine_admin_v1/services.rst:self referenced toctree found. Ignored.
```

I'd like to propose to rename `docs/<api>/services` to `docs/<api>/services_` when there is a service called `services`. For example,

image

You can see that there is a service called `services` below:
image

parthea:
This PR removes the need for workarounds [here](https://github.com/googleapis/python-run/blob/082e5630fd574aa467cf03b601d5166203d64052/owlbot.py#L40) and [here](https://github.com/googleapis/python-appengine-admin/blob/b66d7f5a75b39e5b82cb0228fddd0ae0a72be5d3/owlbot.py#L42) where the autogenerated `docs/index.rst` and `docs/<API>/services.rst` are excluded.

Fixes https://github.com/googleapis/gapic-generator-python/issues/825 🦕

### DIFF
--- a/gapic/generator/generator.py
+++ b/gapic/generator/generator.py
@@ -410,6 +410,18 @@ class Generator:
         # (e.g. no %version); handle this.
         filename = re.sub(r"/+", "/", filename)
 
+        # Check if the API contains a service called "Services" so we can avoid a
+        # conflict with the templated RST file which lives at
+        # `gapic/templates/docs/%name_%version/services.rst.j2`.
+        # Use `services_.rst.j2` instead of `services.rst.j2` when there is a service
+        # called `services`.
+        # See https://github.com/googleapis/gapic-generator-python/issues/825
+        if template_name.endswith("services.rst.j2"):
+            if any(
+                [service.name == "Services" for service in api_schema.services.values()]
+            ):
+                filename = filename.replace("services.rst", "services_.rst")
+
         # Done, return the filename.
         return filename
 
--- a/gapic/templates/docs/index.rst.j2
+++ b/gapic/templates/docs/index.rst.j2
@@ -2,6 +2,16 @@ API Reference
 -------------
 .. toctree::
     :maxdepth: 2
-    
+
+    {% set index_ns = namespace(has_services_api=false) %}    
+    {% for service in api.services.values() | selectattr('name') %}
+    {% if service.name == "Services" %}
+    {% set index_ns.has_services_api = True %}
+    {% endif %}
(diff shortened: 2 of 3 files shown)


gapic/generator/generator.py
```diff
@@ -410,6 +410,18 @@ def _get_filename(
...
 
+        # Check if the API contains a service called "Services" so we can avoid a
+        # conflict with the templated RST file which lives at
+        # `gapic/templates/docs/%name_%version/services.rst.j2`.
+        # Use `services_.rst.j2` instead of `services.rst.j2` when there is a service
+        # called `services`.
+        # See https://github.com/googleapis/gapic-generator-python/issues/825
+        if template_name.endswith("services.rst.j2"):
+            if any(
+                [service.name == "Services" for service in api_schema.services.values()]
```
vchudnov-g, line 421: Should we be proactive and handle any service name that will class with our docs? I'm thinking of a possible service names `Types`

gapic/templates/docs/index.rst.j2
```diff
@@ -2,6 +2,16 @@ API Reference
 -------------
 .. toctree::
     :maxdepth: 2
-    
+
+    {% set index_ns = namespace(has_services_api=false) %}    
+    {% for service in api.services.values() | selectattr('name') %}
+    {% if service.name == "Services" %}
```
vchudnov-g, line 8: Should we also check for `service.name == "Types"` and set `index_ns.has_types_api`?

gapic/templates/docs/index.rst.j2
```diff
@@ -2,6 +2,16 @@ API Reference
...
+    {% set index_ns.has_services_api = True %}
```
vchudnov-g, line 9: nit: since "API" is so overloaded, maybe `.has_service_named_services` and `.has_service_named_types`?

PR #1726, opened 2023-08-09
fix: remove duplicate import statement for `google.longrunning.operations_pb2`

### LINKED ISSUE #1727: Duplicate import statement for `google.longrunning.operations_pb2` in `python-logging`
In PR https://github.com/googleapis/python-logging/pull/746, there is a duplicate import statement for `google.longrunning.operations_pb2`. See the review comment [here](https://github.com/googleapis/python-logging/pull/746#discussion_r1190481195).

parthea:
Fixes googleapis/gapic-generator-python#1727 🦕


gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/base.py.j2
```diff
@@ -18,9 +18,14 @@ from google.auth import credentials as ga_credentials  # type: ignore
 from google.oauth2 import service_account # type: ignore
 
 {% filter sort_lines %}
+{% set import_ns = namespace(has_operations_mixin=false) %}
```
vchudnov-g, line 21: It seems like this may need to be generalized to the other mixins. WDYT?

I don't think we should block on that, but I would add a TODO in the code and reference this PR.

vchudnov-g (approved):

PR #1740, opened 2023-08-14
fix: fix docs build for numbered lists

### LINKED ISSUE #1723: Docs build in google-cloud-datalabeling fails with `Enumerated list ends without a blank line; unexpected unindent.`
See build log [here](https://github.com/googleapis/google-cloud-python/actions/runs/5719226812/job/15496680091) which shows the docs build failure with error message `Enumerated list ends without a blank line; unexpected unindent.`. The issue is that lines in some list items are not indented. The list exists [here](https://github.com/googleapis/googleapis/blob/53cd9ad1b48e40cdd44e0c13e96ac0281b32828f/google/cloud/datalabeling/v1beta1/human_annotation_config.proto#L131-L135) , but it is not indented properly in the generated output [here](https://github.com/googleapis/googleapis-gen/blob/7e8867efbed7dbfe5ef6ec3c2c92a4bce4280f7a/google/cloud/datalabeling/v1beta1/datalabeling-v1beta1-py/google/cloud/datalabeling_v1beta1/types/human_annotation_config.py#L245-L251). For example, see this error below

```
sphinx.errors.SphinxWarning: /usr/local/google/home/partheniou/git/google-cloud-python/packages/google-cloud-datalabeling/google/cloud/datalabeling_v1beta1/types/human_annotation_config.py:docstring of google.cloud.datalabeling_v1beta1.types.human_annotation_config.VideoClassificationConfig:5:Enumerated list ends without a blank line; unexpected unindent.
```

https://github.com/googleapis/googleapis-gen/blob/7e8867efbed7dbfe5ef6ec3c2c92a4bce4280f7a/google/cloud/datalabeling/v1beta1/datalabeling-v1beta1-py/google/cloud/datalabeling_v1beta1/types/human_annotation_config.py#L245-L251

We have
```
class VideoClassificationConfig(proto.Message):
    r"""Config for video classification human labeling task.
    Currently two types of video classification are supported:


    1. Assign labels on the entire video.
    2. Split the video into multiple video clips based on camera
    shot, and assign labels on each video clip.
```

but sphinx expects the text in the final list item to be indented like this
```
class VideoClassificationConfig(proto.Message):
    r"""Config for video classification human labeling task.
    Currently two types of video classification are supported:


    1. Assign labels on the entire video.
    2. Split the video into multiple video clips based on camera
       shot, and assign labels on each video clip.
```

### LINKED ISSUE #1724: Docs build in google-cloud-build fails with `Bullet list ends without a blank line; unexpected unindent`
See build log [here](https://github.com/googleapis/google-cloud-python/actions/runs/5719226812/job/15496680091) which shows the docs build failure with error message `Enumerated list ends without a blank line; unexpected unindent.`. The issue is that lines in some list items are not indented. The list exists [here](https://github.com/googleapis/googleapis/blob/53cd9ad1b48e40cdd44e0c13e96ac0281b32828f/google/devtools/cloudbuild/v1/cloudbuild.proto#L1705-L1710) , but it is not indented properly in the generated output [here](https://github.com/googleapis/googleapis-gen/blob/7e8867efbed7dbfe5ef6ec3c2c92a4bce4280f7a/google/devtools/cloudbuild/v1/devtools-cloudbuild-v1-py/google/cloud/devtools/cloudbuild_v1/types/cloudbuild.py#L2273-L2281). For example, see this error below

```
sphinx.errors.SphinxWarning: /usr/local/google/home/partheniou/git/python-cloudbuild/google/cloud/devtools/cloudbuild_v1/types/cloudbuild.py:docstring of google.cloud.devtools.cloudbuild_v1.types.cloudbuild.BuildTrigger:40:Bullet list ends without a blank line; unexpected unindent.
```

https://github.com/googleapis/googleapis-gen/blob/7e8867efbed7dbfe5ef6ec3c2c92a4bce4280f7a/google/devtools/cloudbuild/v1/devtools-cloudbuild-v1-py/google/cloud/devtools/cloudbuild_v1/types/cloudbuild.py#L2273-L2281

We have
```
        name (str):
            User-assigned name of the trigger. Must be
            unique within the project. Trigger names must
            meet the following requirements:

            + They must contain only alphanumeric characters
            and dashes. + They can be 1-64 characters long.
            + They must begin and end with an alphanumeric
            character.
```

but sphinx expects the text in the list items to be indented like this
```
        name (str):
            User-assigned name of the trigger. Must be
            unique within the project. Trigger names must
            meet the following requirements:

            + They must contain only alphanumeric characters
              and dashes. 
            + They can be 1-64 characters long.
            + They must begin and end with an alphanumeric
              character.
```

parthea:
Fixes googleapis/gapic-generator-python#1723 🦕
Fixes googleapis/gapic-generator-python#1724 🦕


gapic/utils/lines.py
```diff
@@ -40,6 +40,47 @@ def sort_lines(text: str, dedupe: bool = True) -> str:
...
+
+    Here subsequent lines should be indented by 3
+
+    1. The quick brown fox jumps over the lazy dog. The quick brown fox jumps
+       over the lazy dog
+
+    """
+    if len(list_item) >= 2 and list_item[0:2] in ['- ', '+ ']:
+        indentation_level = 2
+    elif len(list_item) >= 3 and list_item[0].isdigit() and list_item[1:3] == '. ':
```
vchudnov-g, line 67: This isn't robust if we get into double digits. How about searching that anything preceding the first space is a series of digits followed by a period?
```suggestion
    elif re.match("^\d+\. ", list_item):
```

vchudnov-g, line 67: This isn't robust if we get into double digits. How about searching that anything preceding the first space is a series of digits followed by a period?
```suggestion
    elif re.match("^\d+\. ", list_item):
```

Haven't tested this, and we may want to pre-define the pattern for efficiency.

gapic/utils/lines.py
```diff
@@ -40,6 +40,47 @@ def sort_lines(text: str, dedupe: bool = True) -> str:
...
+    return indentation_level
+
+
+def is_list_item(list_item: str) -> bool:
+    """
+    Given a string return a boolean indicating whether a list is identified.
+    """
+    if len(list_item) < 3:
+        return False
+    return list_item.startswith('- ') or list_item.startswith('+ ') or (list_item[0].isdigit() and list_item[1:].startswith('. '))
```
vchudnov-g, line 81: Again, issue with double-digits.

```suggestion
    return list_item.startswith('- ') or list_item.startswith('+ ') or re.match("^\d+\. ", list_item)
```

tests/unit/utils/test_lines.py
```diff
@@ -246,3 +246,37 @@ def test_list_with_multiple_paragraphs():
...
 ante non, vestibulum erat. Donec mollis ultricies nisl."""
     assert lines.wrap(input, width=60) == expected
+
+
+def test_list_with_numbered_list():
+    input = """Config for video classification human labeling task.
+Currently two types of video classification are supported:
+1. Assign labels on the entire video.
+2. Split the video into multiple video clips based on camera shot, and
+assign labels on each video clip."""
```
vchudnov-g, line 256: Include double-digit entries too.

vchudnov-g (changes requested):
Looks good, but let's make this robust to numbered list entries whose numbers are more than one digit.

gapic/utils/lines.py
parthea: Based on local testing, the sphinx docs build will fail with double digits. This could be a bug in sphinx. I don't think it's worth while supporting double digits if sphinx can't handle it. We don't have any APIs using numbered lists with double digits so I'm not sure if it's worthwhile to go down the rabbit hole.

tests/unit/utils/test_lines.py
parthea: See response above

gapic/utils/lines.py
vchudnov-g: > Based on local testing, the sphinx docs build will fail with double digits.

I'm surprised. The [Sphinx spec](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#lists-and-quote-like-blocks) does not mention double-digit items but implies they are supported. Is it a problem with our testing?

I think we should still implement my suggestions about regexes, because then at least _this_ code you're changing will be future-proof and work when we encounter a double-digit list. This way we don't accrue more technical debt. All that remains will be figuring out why the docs tests fail with double-digit items, which we can track and fix separately (and lower priority for now since we haven't encountered it in the field, as you say).

tests/unit/utils/test_lines.py
parthea: Fixed

tests/integration/goldens/eventarc/google/cloud/eventarc_v1/types/channel.py
```diff
@@ -102,9 +102,9 @@ class State(proto.Enum):
                 possible cases this state can happen:
 
                 1. The SaaS provider disconnected from this
-                Channel. 2. The Channel activation token has
-                expired but the SaaS provider    wasn't
-                connected.
+                    Channel.
+                2. The Channel activation token has expired but
+                    the SaaS provider    wasn't connected.
```
vchudnov-g, line 107: Weird whitespace here (was here before). Separate issue to be aware of.

gapic/utils/lines.py
```diff
@@ -56,16 +56,18 @@ def get_subsequent_line_indentation_level(list_item: str) -> int:
...
+        over the lazy dog
 
+    22. The quick brown fox jumps over the lazy dog. The quick brown fox jumps
+        over the lazy dog
     """
     if len(list_item) >= 2 and list_item[0:2] in ['- ', '+ ']:
         indentation_level = 2
-    elif len(list_item) >= 3 and list_item[0].isdigit() and list_item[1:3] == '. ':
-        indentation_level = 3
+    elif len(list_item) >= 4 and re.match("^\d+\. ", list_item):
```
vchudnov-g, line 69: Considering factoring out the regex itself into a constant, so we have a single source of truth, rather than repeating it here and below. (But granted, it's not terribly complicated.)

vchudnov-g (approved):

gapic/utils/lines.py
parthea: Fixed in [73dd48f](https://github.com/googleapis/gapic-generator-python/pull/1740/commits/73dd48f25319e085fc878409289b0073fcbadf47)

tests/integration/goldens/eventarc/google/cloud/eventarc_v1/types/channel.py
parthea: Filed https://github.com/googleapis/gapic-generator-python/issues/1754

PR #1799, opened 2023-10-12
fix(revert): partial revert of #1125

parthea:
This PR partially reverts googleapis/gapic-generator-python#1125 to fix the issue captured in googleapis/google-cloud-python#16369. I wasn't able to create a minimal reproduction. We should keep googleapis/google-cloud-python#16369 open until we can create a minimal test for the issue.

### DIFF
--- a/gapic/schema/wrappers.py
+++ b/gapic/schema/wrappers.py
@@ -713,7 +713,7 @@ class MessageType:
         references.
         """
         visited_messages = visited_messages or set()
-        visited_messages.add(self)
+        visited_messages = visited_messages | {self}
         return dataclasses.replace(
             self,
             fields={


ohmayr (approved):

gapic/schema/wrappers.py
```diff
@@ -713,7 +713,7 @@ def with_context(self, *,
         references.
         """
         visited_messages = visited_messages or set()
-        visited_messages.add(self)
+        visited_messages = visited_messages | {self}
```
vchudnov-g, line 716: @parthea How is adding an element `b` to a set `A` different than updating `A` to be the union of `A` and `{b}`?

parthea: Great question. I don't have the answer. Issue https://github.com/googleapis/gapic-generator-python/issues/1794 is still open to try to create a minimal reproduction of the issue to dig deeper.

PR #1804, opened 2023-10-16
fix: add missing dependencies

parthea:
This PRs adds the necessary dependencies for generating `google-cloud-asset`. See setup.py here: https://github.com/googleapis/python-asset/blob/7c1f5fec48364889da95d21042278a9d944688c4/setup.py#L41-L43


ohmayr (approved):

parthea:
The current failure is from `mypy`
```
nox > mypy --explicit-package-bases google
google/cloud/asset_v1/types/assets.py:266: error: Variable "google.cloud.orgpolicy.v1.orgpolicy_pb2.Policy" is not valid as a type  [valid-type]
google/cloud/asset_v1/types/assets.py:266: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
```

parthea:
I removed `google-cloud-org-policy` which caused mypy to fail. I'll create a separate PR once I get mypy to pass.

PR #1834, opened 2023-10-30
Fix: wrap method in async client

### LINKED ISSUE #11415: Async clients do not wrap RPC errors for all functions
Some methods in `*AsyncClients` use `google.api_core.gapic_v1.method.wrap_method` instead of `google.api_core.gapic_v1.method_async.wrap_method`. For example, the `google.cloud.artifactregistry_v1.ArtifactRegistryAsyncClient` class has:

```python
    async def list_locations(...):
        [...]
        rpc = gapic_v1.method.wrap_method(
            self._client._transport.list_locations,
            default_timeout=None,
        )
        [...]
        response = await rpc(
            request,
            retry=retry,
            timeout=timeout,
            metadata=metadata,
        )
```

The issue is whether `google.api_core.grpc_helpers.wrap_errors` or `google.api_core.grpc_helpers_async.wrap_errors` is called. Only the latter decorator will wrap `grpc.RpcError` instances in `google.api_core.exceptions.GoogleAPIError` instances during awaiting of a coroutine, while the former will effectively just catch exceptions when obtaining a coroutine (when no exceptions are raised, anyway) and return it, only for exceptions to be raised (and not wrapped) later.

The end result is that the `*AsyncClient`'s method will raise grpc.RpcError instances, instead of `google.api_core.exceptions.GoogleAPIError`. This doesn't follow the documentation or the convention and will result in default `AsyncRetry` instances not working (because they expect the latter type of exception).

Interestingly, not all methods within the same `*AsyncClient` class have this issue. For example, `google.cloud.artifactregistry_v1.ArtifactRegistryAsyncClient`'s `list_repositories` method correctly uses `google.api_core.gapic_v1.method_async.wrap_method`.

Searching for packages using the naming and coding convention yields a fair few with issues:

```bash
$ find packages -name async_client.py -exec grep 'method\.wrap_method' '{}' + | cut -d/ -f2 | sort -u
google-cloud-alloydb
google-cloud-api-keys
google-cloud-apigee-registry
google-cloud-artifact-registry
google-cloud-bare-metal-solution
google-cloud-beyondcorp-appconnections
google-cloud-beyondcorp-appconnectors
google-cloud-beyondcorp-appgateways
google-cloud-beyondcorp-clientconnectorservices
google-cloud-beyondcorp-clientgateways
google-cloud-bigquery-data-exchange
google-cloud-certificate-manager
google-cloud-channel
google-cloud-confidentialcomputing
google-cloud-contact-center-insights
google-cloud-contentwarehouse
google-cloud-datacatalog
google-cloud-datacatalog-lineage
google-cloud-dataform
google-cloud-dataplex
google-cloud-dataproc-metastore
google-cloud-datastream
google-cloud-deploy
google-cloud-discoveryengine
google-cloud-dms
google-cloud-documentai
google-cloud-edgecontainer
google-cloud-eventarc
google-cloud-functions
google-cloud-gke-multicloud
google-cloud-life-sciences
google-cloud-memcache
google-cloud-migrationcenter
google-cloud-network-connectivity
google-cloud-network-security
google-cloud-notebooks
google-cloud-orchestration-airflow
google-cloud-private-ca
google-cloud-rapidmigrationassessment
google-cloud-redis
google-cloud-resource-manager
google-cloud-securitycenter
google-cloud-storageinsights
google-cloud-talent
google-cloud-tpu
google-cloud-vm-migration
google-cloud-vmwareengine
google-cloud-vpc-access
google-cloud-webrisk
google-cloud-workstations
```

For those that are affected by this issue, you can use a workaround that will continue to work once this bug is addressed: just wrap the `grpc.RpcError` instances in `google.api_core.exceptions.GoogleAPIError` instances in a custom predicate. For example, to use the default behaviour of retrying for "transient" errors as determined by `google.api_core.retry`, you can use something like:

```python
import grpc
import google.api_core.exceptions
import google.api_core.retry
import google.api_core.retry_async

def check_error(exc):
    if isinstance(exc, grpc.RpcError):
        exc = google.api_core.exceptions.from_grpc_error(exc)
    return google.api_core.retry.if_transient_error(exc)

retry = google.api_core.retry_async.AsyncRetry(predicate=check_error)
response = await client.list_locations(..., retry=retry)
```

ohmayr:
Fixes: googleapis/google-cloud-python#11415 🦕

### DIFF
--- a/tests/system/test_retry.py
+++ b/tests/system/test_retry.py
@@ -52,3 +52,14 @@ if os.environ.get("GAPIC_PYTHON_ASYNC", "true") == "true":
                     'message': 'This service is not available.',
                 },
             })
+
+    # Note: This test verifies that:
+    # Using gapic_v1.method.wrap_method in *AsyncClient raises a RPCError (Incorrect behaviour).
+    # Using gapic_v1.method_async.wrap_method in *AsyncClient raises a google.api_core.exceptions.GoogleAPIError.
+
+    @pytest.mark.asyncio
+    async def test_method_async_wrapper_for_async_client(async_echo):
+        with pytest.raises(exceptions.NotFound):
+            await async_echo.get_operation({
+                'name': "operations/echo"
+            })
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/_async_mixins.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/_async_mixins.py.j2
@@ -34,7 +34,7 @@
 
         # Wrap the RPC method; this adds retry and timeout information,
         # and friendly error handling.
-        rpc = gapic_v1.method.wrap_method(
+        rpc = gapic_v1.method_async.wrap_method(
             self._client._transport.list_operations,
             default_timeout=None,
             client_info=DEFAULT_CLIENT_INFO,
@@ -88,7 +88,7 @@
 
         # Wrap the RPC method; this adds retry and timeout information,
         # and friendly error handling.
-        rpc = gapic_v1.method.wrap_method(
+        rpc = gapic_v1.method_async.wrap_method(
             self._client._transport.get_operation,
             default_timeout=None,
             client_info=DEFAULT_CLIENT_INFO,
@@ -145,7 +145,7 @@
 
         # Wrap the RPC method; this adds retry and timeout information,
         # and friendly error handling.
-        rpc = gapic_v1.method.wrap_method(
+        rpc = gapic_v1.method_async.wrap_method(
             self._client._transport.delete_operation,
             default_timeout=None,
             client_info=DEFAULT_CLIENT_INFO,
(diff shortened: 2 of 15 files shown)


tests/system/test_retry.py
```diff
@@ -45,10 +45,10 @@ def test_retry_bubble(echo):
 
     @pytest.mark.asyncio
     async def test_retry_bubble_async(async_echo):
-        with pytest.raises(exceptions.DeadlineExceeded):
+        with pytest.raises(exceptions.RetryError):
```
vchudnov-g, line 48: Does this mean that the exception the user sees in case of an error changes? If so, we should flag this as a breaking change, since users may be catching the previously raised exception.

ohmayr: It's more like this test case was defined incorrectly. Both `exceptions.DeadlineExceeded` and `exceptoins.RetryError` are raised but we're checking for latter because that confirms that async retry is working as expected.

ohmayr: Note that `exceptions.RetryError` was not raised before.

parthea (approved):

PR #1861, opened 2023-11-11
feat: Add support for python 312 (ads templates)

### LINKED ISSUE #1858: Remove usage of `pkg_resources` in ads templates
https://github.com/googleapis/gapic-generator-python/blob/746466b49d46059375e66f713befb39d718d5c4f/gapic/ads-templates/%25namespace/%25name/%25version/%25sub/services/%25service/client.py.j2#L9

https://github.com/googleapis/gapic-generator-python/blob/746466b49d46059375e66f713befb39d718d5c4f/gapic/ads-templates/%25namespace/%25name/%25version/%25sub/services/%25service/client.py.j2#L809-L816

### LINKED ISSUE #1859: Add support for python 3.12 in ads templates
Once https://github.com/googleapis/gapic-generator-python/issues/1858 is fixed, we should add support for python 3.12 in ads templates: https://github.com/googleapis/gapic-generator-python/tree/main/gapic/ads-templates

parthea:
Fixes #1858 🦕
Fixes #1859 🦕

This PR: 
- Removes usage of `exec` in both ads and standard templates
- Updates the `setup.py` of Ads to be similar to the standard templates
- Removes usage of `pkg_resources`
- Add support for python 3.12 in Ads templates
- Adds version module `gapic_version.py` to Ads templates
- Adds `__version__` to `__init__.py` for Ads templates

### DIFF
--- a/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
@@ -6,7 +6,6 @@ from collections import OrderedDict
 import os
 import re
 from typing import Callable, Dict, Mapping, MutableMapping, MutableSequence, Optional, {% if service.any_server_streaming %}Iterable, {% endif %}{% if service.any_client_streaming %}Iterator, {% endif %}Sequence, Tuple, Type, Union, cast
-import pkg_resources
 {% if service.any_deprecated %}
 import warnings
 {% endif %}
@@ -20,6 +19,9 @@ from google.auth.transport import mtls                            # type: ignore
 from google.auth.transport.grpc import SslCredentials             # type: ignore
 from google.auth.exceptions import MutualTLSChannelError          # type: ignore
 from google.oauth2 import service_account                         # type: ignore
+
+{% set package_path = api.naming.module_namespace|join('.') + "." + api.naming.versioned_module_name %}
+from {{package_path}} import gapic_version as package_version
 
 try:
     OptionalRetry = Union[retries.Retry, gapic_v1.method._MethodDefault]
@@ -806,14 +808,7 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
     {% endif %}
 
 
-try:
-    DEFAULT_CLIENT_INFO = gapic_v1.client_info.ClientInfo(
-        gapic_version=pkg_resources.get_distribution(
-            "{{ api.naming.warehouse_package_name }}",
-        ).version,
-    )
-except pkg_resources.DistributionNotFound:
-    DEFAULT_CLIENT_INFO = gapic_v1.client_info.ClientInfo()
+DEFAULT_CLIENT_INFO = gapic_v1.client_info.ClientInfo(gapic_version=package_version.__version__)
 
 
 __all__ = (
--- a/gapic/ads-templates/setup.py.j2
+++ b/gapic/ads-templates/setup.py.j2
@@ -2,36 +2,58 @@
 
 {% block content %}
 
+import os
+import re
+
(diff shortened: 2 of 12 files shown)


ohmayr (approved):

gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
```diff
@@ -806,14 +808,7 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
 
-try:
-    DEFAULT_CLIENT_INFO = gapic_v1.client_info.ClientInfo(
-        gapic_version=pkg_resources.get_distribution(
-            "{{ api.naming.warehouse_package_name }}",
-        ).version,
-    )
-except pkg_resources.DistributionNotFound:
-    DEFAULT_CLIENT_INFO = gapic_v1.client_info.ClientInfo()
+DEFAULT_CLIENT_INFO = gapic_v1.client_info.ClientInfo(gapic_version=package_version.__version__)
```
BenRKarl, line 811: It looks like this is grabbing the version that's defined in the above `gapic_version` file, is that correct? 

The implementation is actually to access the version of the Ads client library itself, see the logic [here](https://github.com/googleads/google-ads-python/blob/main/google/ads/googleads/v15/services/services/google_ads_service/client.py#L4467)

If you look at go/ads-client-libs-user-agent-design you can see that we opted to use the `gccl` user agent k/v for our own logging purposes, which is why the `gapic_version` keyword arg is used here instead of something else. 

Is it possible to retrieve the client library version without pkg_resources? If not, then we might need to add that logic in our interceptors before this change can be made.

gapic/ads-templates/setup.py.j2
```diff
@@ -2,55 +2,84 @@
...
+        "Programming Language :: Python :: 3.11",
+        "Programming Language :: Python :: 3.12",
+        "Operating System :: OS Independent",
+        "Topic :: Internet",
     ],
+    platforms="Posix; MacOS X; Windows",
+    packages=packages,
+    python_requires=">=3.7",
+    install_requires=dependencies,
+    include_package_data=True,
```
BenRKarl, line 82: We don't use the generated setup.py file, so I'm not sure it makes sense to make these changes, and perhaps you could remove it from the templates.

gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/client.py.j2
parthea: > It looks like this is grabbing the version that's defined in the above gapic_version file, is that correct?

Yes, that is correct. In [google-cloud-python](https://github.com/googleapis/google-cloud-python), we use [release-please](https://github.com/googleapis/release-please) to update the version in `gapic_version.py` prior to releasing. As an example see https://github.com/googleapis/google-cloud-python/pull/12010.

> The implementation is actually to access the version of the Ads client library itself, see the logic [here](https://github.com/googleads/google-ads-python/blob/main/google/ads/googleads/v15/services/services/google_ads_service/client.py#L4467)

We've moved away from this in GAPIC, and chose to populate `gapic_version.py` instead of a hardcoded version number in setup.py. Customers can obtain the version number programatically using `__version__`.

>Is it possible to retrieve the client library version without pkg_resources? If not, then we might need to add that logic in our interceptors before this change can be made.

It is possible to use something else like https://docs.python.org/3/library/importlib.metadata.html but we chose to have a version module instead.

gapic/ads-templates/setup.py.j2
parthea: This file is still needed so that we can run tests for the ads templates. We'll need to maintain it as long as we support ads templates.

PR #1898, opened 2024-01-05
feat: Allow users to explicitly configure universe domain.

### LINKED ISSUE #1866: Update mixins to support TPC
With Client Libraries supporting TPC, users can configure a new property i.e. `client_options.universe_domain` to explicitly configure a universe domain for a TPC universe. see: googleapis/python-api-core/issues/547

Mixin functions need to verify that the universe domain configured via client options matches the universe domain in credentials.

```python3
if self._universe_domain != self._transport._credentials.universe_domain:
            raise ValueError("The configured universe domain (googleapis.com) does not match the universe domain found in the credentials (foo.com). If you haven't configured the universe domain explicitly, `googleapis.com` is the default.")
```

Also make sure that the above change is accompanied and verified by appropriate unit tests.

### LINKED ISSUE #1900: Allow users to explicitly configure universe domain in GAPICs
Allow users to specify a universe domain that they want to connect to instead of using the google default universe (GDU) i.e. `"googleapis.com"`. Note that the universe domain will default to GDU if it is not explicitly configured by the user.

Users will have the option to provide a universe domain via the following options:

1. `Client_Options.universe_domain` client option.
2. `GOOGLE_CLOUD_UNIVERSE_DOMAIN` environment variable.

In order to support universe domain, GAPICs will:
- Deprecate the hardcoded `DEFAULT_ENDPOINT` class variable.
- Introduce `DEFAULT_ENDPOINT_TEMPLATE` which will have the format: `service_name.{UNIVERSE_DOMAIN}`.
- Resolve the template using the universe domain configured by the user.

GAPICs will need ensure that:
- Universe domain cannot be an empty string
- MTLS is only supported for GDU and any other configured universe should raise an exception.
- The configured universe domain is consistent with the one in credentials.

ohmayr:
Fixes #1900 🦕
Fixes #1866


gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -131,12 +131,17 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
 
         return api_endpoint.replace(".googleapis.com", ".mtls.googleapis.com")
 
+    # Note: DEFAULT_ENDPOINT is deprecated. Use DEFAULT_ENDPOINT_TEMPLATE instead.
```
ohmayr, line 134: Deprecation note for `DEFAULT_ENDPOINT`. This should no longer be used anywhere in the code except for `_get_mtls_endpoint_and_cert_source` tests which will eventually be removed once we clean up the code.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
```diff
@@ -133,8 +133,8 @@ class {{ service.name }}GrpcTransport({{ service.name }}Transport):
             warnings.warn("client_cert_source is deprecated", DeprecationWarning)
 
         if channel:
-            # Ignore credentials if a channel was passed.
-            credentials = False
+            # Ignore credentials if a channel was passed without credentials.
+            credentials = False if not credentials else credentials
```
ohmayr, line 137: This wasn't a part of the design so adding some clarification. Validation step (for universe domain) requires us to access the universe domain from the credentials, hence we can no longer ignore credentials if a channel is explicitly passed in (mainly for testing purposes).
i.e. the condition:
```python3
if self.universe_domain != self.transport._credentials.universe_domain
```
fails without this change.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
```diff
@@ -6,7 +6,7 @@
 ])
 def test_{{ method_name }}(request_type, transport: str = 'grpc'):
     client = {{ service.client_name }}(
-        credentials=ga_credentials.AnonymousCredentials(),
+        credentials=AnonymousCredentialsWithUniverseDomain(),
```
ohmayr, line 9: Since validation step is a part of each RPC call, we need to pass in credentials with universe domain property to each test. Otherwise, they will complain about missing `universe_domain` property in the validation step.

tests/system/test_universe_domain.py
```diff
@@ -0,0 +1,20 @@
+import pytest
+
+# Note(TODO): Uncomment these test cases once showcase library is updated
```
parthea, line 3: Please fix. We can update the minimum version of `google-api-core` if necessary.

parthea (commented):
LGTM. Please update the version of `google-api-core` if required. Note, we still support `google-api-core` 1.x so we will need to cherry-pick the necessary change in the `google-api-core` 1.x branch.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
vchudnov-g: Would this be a breaking change? Before, if the user was passing a channel with credentials, we were clearing the credentials. Now we're not. So that will affect the behavior of this constructor, it would seem.

One idea for not changing the code is that when we try to access the universe domain from the credentials, we first check to see whether the credentials are set, and if they're not, consider the universe domain to be the empty string or `None`. Would this help the testing scenario?

vchudnov-g, line 137:
```suggestion
            credentials = credentials or False
```
if you really need a Boolean value, or even better, just delete the line entirely: if `credentials` is not passed in, it is `None` so it evaluates as `False`. In other words, if you don't need a strict Boolean type, this line is a no-op and can be removed. But again, it changes the behavior of this method.

vchudnov-g (commented):
Mostly minor comments, but two non-trivial ones
- we should determine whether the change to the channels+credentials check is breaking
- suggestions for consolidating the check for valid universe domain

vchudnov-g:
@ohmayr Please re-assign back to me once it's ready for the next round of review.

tests/integration/goldens/logging/google/cloud/logging_v2/services/logging_service_v2/client.py
```diff
@@ -308,11 +312,12 @@ def _read_environment_variables():
         """
         use_client_cert = os.getenv("GOOGLE_API_USE_CLIENT_CERTIFICATE", "false").lower()
         use_mtls_endpoint = os.getenv("GOOGLE_API_USE_MTLS_ENDPOINT", "auto").lower()
+        universe_domain_env = os.getenv("GOOGLE_CLOUD_UNIVERSE_DOMAIN")
         if use_client_cert not in ("true", "false"):
             raise ValueError("Environment variable `GOOGLE_API_USE_CLIENT_CERTIFICATE` must be either `true` or `false`")
         if use_mtls_endpoint not in ("auto", "never", "always"):
             raise MutualTLSChannelError("Environment variable `GOOGLE_API_USE_MTLS_ENDPOINT` must be `never`, `auto` or `always`")
-        return use_client_cert == "true", use_mtls_endpoint
+        return use_client_cert == "true", use_mtls_endpoint, universe_domain_env
```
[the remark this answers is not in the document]
ohmayr: shouldn't we keep the logic for figuring out the `mtls_endpoint` in the `get_api_endpoint function`? The purpose of this function is just to read variables as the name suggests i.e. `read_environment_variables`.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -316,26 +322,56 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+            universe_domain_env (str): The universe domain configured via the "GOOGLE_CLOUD_UNIVERSE_DOMAIN" environment variable.
+
+        Returns:
+            str: The universe domain to be used by the client.
+        """
+        universe_domain = {{ service.client_name }}.GOOGLE_DEFAULT_UNIVERSE
+        if client_universe_domain is not None:
+            universe_domain = client_universe_domain
+        elif universe_domain_env is not None:
+            universe_domain = universe_domain_env
```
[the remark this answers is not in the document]
ohmayr: This is a much concise alternative but I run into some issues when I make this change:
- `""` is treated as None in the `or` statement and the statement will end up evaluating to the default GDU value. As a result, the below exception for checking the empty string will never be raised (the exception is part of the design.)
- If I were to split the `or` statement to two, the `len` check on the None type would lead to an error.
- If I do the error checking before assigning the `universe_domain` value for each individual parameter, then I end up raising an exception for the following scenario i.e. `client_universe_domain="foo.com"` and `universe_domain_env=""`. In this case, `client_universe_domain` is a valid universe and we shouldn't bother if a universe is configured (even incorrectly) via the environment variable.

There are obviously workarounds to make this work but that seem to defeat the purpose of what's proposed here. 

Feel free to suggest more on this!

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -131,12 +131,17 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+    # Note: DEFAULT_ENDPOINT is deprecated. Use DEFAULT_ENDPOINT_TEMPLATE instead.
     DEFAULT_ENDPOINT = {% if service.host %}"{{ service.host }}"{% else %}None{% endif %}
 
     DEFAULT_MTLS_ENDPOINT = _get_default_mtls_endpoint.__func__(  # type: ignore
         DEFAULT_ENDPOINT
     )
 
+    DEFAULT_ENDPOINT_TEMPLATE = {% if service.host %}"{{ service.host.replace("googleapis.com", "{UNIVERSE_DOMAIN}") }}"{% else %}None{% endif %}
+    
+    GOOGLE_DEFAULT_UNIVERSE = "googleapis.com"
```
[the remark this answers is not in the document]
ohmayr: `GOOGLE_DEFAULT_UNIVERSE` can definitely be moved to `api_core` but not `DEFAULT_ENDOINT_TEMPLATE` since the value for it is dynamically resolved.

Note that we're using an older version of `api_core` for `python 3.7`. As a result we'll have to have the check `hasattr(Client_options, "universe_domain")` every time we use code from `api_core` in gapic that we recently add.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -316,26 +322,56 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+        universe_domain = {{ service.client_name }}.GOOGLE_DEFAULT_UNIVERSE
+        if client_universe_domain is not None:
+            universe_domain = client_universe_domain
+        elif universe_domain_env is not None:
+            universe_domain = universe_domain_env
+        if universe_domain == "":
+            raise ValueError("Universe Domain cannot be an empty string.")
+        return universe_domain
+    
+    @staticmethod
```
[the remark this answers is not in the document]
ohmayr: 1. and 2. can definitely be done (having workarounds in place so we don't break code if an older version of `api_core` is being used). [I'm working on this.]

3. This is decided in the design for TPC and needs to be consistent across platforms i.e.
- We should wait for validating the universe domain until the first call to an API is made.

Also, even if a user provides an incorrect universe domain value that does not match the one in credentials, they should be able to initialize a client and be able to read the value for universe domain in it. The following should work:

```python3

client = EchoClient(ClientOptions{"universe_domain":"foo.com"})
client.universe_domain

File: gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2 [hunk 9]
```diff
@@ -345,7 +381,17 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+    
+    @property
+    def universe_domain(self) -> str:
+        """Return the universe domain used by the client instance.
+        
+        Returns:
+            str: The universe domain used
+                by the client instance.
+        """
+        return self._universe_domain
```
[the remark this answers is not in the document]
ohmayr: Users should be able to access the value that they've set even if it's incorrect? The validation will be delayed until an API call is made (as part of the design).

tests/integration/goldens/logging/google/cloud/logging_v2/services/logging_service_v2/client.py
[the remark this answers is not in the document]
vchudnov-g: Sorry, I forgot to delete this comment before sending out the review. Ignore.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
ohmayr: Addressed!

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -78,13 +78,29 @@ from google.iam.v1 import policy_pb2  # type: ignore
 def client_cert_source_callback():
     return b"cert bytes", b"key bytes"
 
-
+{#TODO: Remove this function as part of cleanup when DEFAULT_ENDPOINT is no longer used.#}
```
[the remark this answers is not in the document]
ohmayr: addressed!

tests/system/test_universe_domain.py
ohmayr: addressed!

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
```diff
@@ -54,8 +54,11 @@ class {{ service.async_client_name }}:
 
     _client: {{ service.client_name }}
 
+    # Note: DEFAULT_ENDPOINT is deprecated. Use DEFAULT_ENDPOINT_TEMPLATE instead.
```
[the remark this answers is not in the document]
ohmayr: addressed!

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -316,26 +322,56 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
         return client_cert_source
 
-    def _get_api_endpoint(api_override, client_cert_source, use_mtls_endpoint):
+    def _get_api_endpoint(api_override, client_cert_source, universe_domain, use_mtls_endpoint):
         """Return the API endpoint used by the client.
 
         Args:
             api_override (str): The API endpoint override. If specified, this is always the return value of this function.
             client_cert_source (bytes): The client certificate source used by the client.
+            universe_domain (str): The universe domain used by the client.
```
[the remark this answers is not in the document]
ohmayr: addressed!

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -316,26 +322,56 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
             use_mtls_endpoint (str): How to use the MTLS endpoint, which depends also on the other parameters.
                 Possible values are "always", "auto", or "never".
 
         Returns:
             str: The API endpoint to be used by the client.
         """
-
+        use_mtls_endpoint = use_mtls_endpoint == "always" or (use_mtls_endpoint == "auto" and client_cert_source)
```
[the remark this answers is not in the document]
ohmayr: addressed!

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -316,26 +322,56 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+            raise ValueError("Universe Domain cannot be an empty string.")
+        return universe_domain
+    
+    @staticmethod
+    def _validate_universe_domain(client_universe, credentials_universe):
+        """Validates the universe domain used by the client instance against
+            the universe domain in the credentials.
+        """
+        if client_universe != credentials_universe:
+            raise ValueError("The configured universe domain (" + client_universe + ") does not match the universe domain found in the credentials (" + credentials_universe + "). If you haven't configured the universe domain explicitly, `googleapis.com` is the default.")
```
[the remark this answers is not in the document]
ohmayr: addressed!

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -581,6 +631,10 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
                 (("resource", request.resource),)),
         )
 
+        # Validate the universe domain.
+        self._validate_universe_domain(self.universe_domain, self.transport._credentials.universe_domain) if not self.is_universe_domain_valid else None
```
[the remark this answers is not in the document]
ohmayr: addressed!

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -316,26 +322,56 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+        if client_universe_domain is not None:
+            universe_domain = client_universe_domain
+        elif universe_domain_env is not None:
+            universe_domain = universe_domain_env
+        if universe_domain == "":
+            raise ValueError("Universe Domain cannot be an empty string.")
+        return universe_domain
+    
+    @staticmethod
+    def _validate_universe_domain(client_universe, credentials_universe):
```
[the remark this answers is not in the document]
ohmayr: addressed.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
ohmayr: addressed!

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -345,7 +381,17 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
             str: The API endpoint used 
                 by the client instance.
         """
-        return self._api_endpoint    
+        return self._api_endpoint
```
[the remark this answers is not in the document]
ohmayr: addressed!

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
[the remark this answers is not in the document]
vchudnov-g: Oh, good point.

I think this should work, though it's less clear than the non-working version

```suggestion
EMPTY_TOKEN=-9999 # since empty strings are falsy, but we want to distinguish them from None
universe_domain = (((client_universe_domain is not None) and (client_universe_domain or EMPTY_TOKEN)) or
                   ((universe_domain_env is not None) and (universe_domain_env or EMPTY_TOKEN)) or
                   {{ service.client_name }}.GOOGLE_DEFAULT_UNIVERSE)
 if universe_domain == EMPTY_TOKEN or len(universe_domain.strip()) == 0:
     raise ValueError(...)
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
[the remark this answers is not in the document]
vchudnov-g: So then we would not have custom universes for Python 3.7? Does that go against the requirements of this project, or is that acceptable?

If that would be a problem, you can keep the code as is but add a TODO outlining two options I see (do you see others?):
*  when we stop supporting Python 3.7, we can move this symbol to api_core; or
* we could add the symbol to the latest version of `api_core` and at the same time:
  *  take the version of `api_core` we support for 3.7, make a branch for that, and there only add what we need for this feature. But we'd have to be clear on how to release it and on making sure we pin the dependency correctly. 


(Probably not worth investing the time doing all that coding work at the moment, which is why I'm suggesting a TODO if we need to support the feature in 3.7)

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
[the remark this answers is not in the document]
vchudnov-g: Hmmm, OK. I'll have to refresh myself on the motivation for that design decision. 

But OK, if that's the feature requirement, we can't do 3. (And I'm not seeing any other similar lazy initialization we can lump together with this one.)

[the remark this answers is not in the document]
vchudnov-g: Also:

> (having workarounds in place so we don't break code if an older version of api_core is being used). 

If the workarounds make the generated code less clear or less readable, we may just have to make do with a TODO for now saying when we can do what simplifications.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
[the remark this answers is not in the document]
vchudnov-g: Your call on which of the two alternatives is easier to follow. I slightly prefer my version, but there's a tradeoff between conciseness and thinking through the code---not a big burden either way in this case.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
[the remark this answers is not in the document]
parthea: > So then we would not have custom universes for Python 3.7? Does that go against the requirements of this project, or is that acceptable?

The issue is not specific to python 3.7 but rather whether the user has an older version of `google-api-core` installed and wants to use the new feature which requires a newer version of `google-api-core`

Possible solution : If the user sets `universe_domain` and has an older version of `google-api-core` installed, we can raise an error that a newer version of `google-api-core`  is required (we should provide the users with specific minimum version).

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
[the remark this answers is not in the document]
ohmayr: I think let's keep it as is since it looks easier to follow through for someone who might not be familiar with what's happening here.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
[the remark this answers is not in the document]
ohmayr: addressed.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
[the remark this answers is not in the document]
ohmayr: I've addressed this by implementing [1.] and [2.] in the following fashion:
- I have exactly what you've proposed except I've moved the logic for raising the exception to a new private function `compare_universes` which is a place holder for `api_core.validate_universe_domain`. This way, when we eventually plan to move code to `api_core`, we can just move this function directly without requiring much refactor.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -316,36 +322,94 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+    def _compare_universes(client_universe, credentials_universe):
+        """Returns True if the universe domain used by the client
+            is the same as the universe domain in credentials.
+
+        Args:
+            client_universe (str): The universe domain configured via the client options.
+            credentials_universe (str): The universe domain in the credentials.
+
+        Returns:
+            bool: True if valid universe, else raise a "ValueError" exception.
```
parthea, line 378:
```suggestion
        Returns:
            bool: Returns True if valid universe domain is valid, otherwise False.
        Raises:
            ValueError: If universe domain is an empty string.
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -316,36 +322,94 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
                 client_cert_source = mtls.default_client_cert_source()
         return client_cert_source
 
-    def _get_api_endpoint(api_override, client_cert_source, use_mtls_endpoint):
+    def _get_api_endpoint(api_override, client_cert_source, universe_domain, use_mtls_endpoint):
         """Return the API endpoint used by the client.
 
         Args:
-            api_override (str): The API endpoint override. If specified, this is always the return value of this function.
+            api_override (str): The API endpoint override. If specified, this is always the return value of this function
```
vchudnov-g, line 329: nit: line wrap (cf https://google.github.io/styleguide/pyguide.html#32-line-length)

parthea (approved):

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -316,36 +322,94 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
 
         Returns:
             str: The API endpoint to be used by the client.
         """
-
         if api_override is not None:
             api_endpoint = api_override
         elif use_mtls_endpoint == "always" or (use_mtls_endpoint == "auto" and client_cert_source):
+            if universe_domain != {{ service.client_name }}._DEFAULT_UNIVERSE:
+                raise MutualTLSChannelError("MTLS is not supported in any universe other than googleapis.com.")
```
vchudnov-g, line 343:
```suggestion
                raise MutualTLSChannelError(f"MTLS is not supported in any universe other than { {{ service.client_name }}._DEFAULT_UNIVERSE} ")
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
vchudnov-g:
```suggestion
            bool: True if the universes match
            
          Raises: 
            ValueError: when the universes do not match.
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -316,36 +322,94 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+        return True
+    
+    def _validate_universe_domain(self):
+        """Validates the universe domain used by the client instance against
+            the universe domain in the credentials.
+        
+        Returns:
+            bool: True if valid universe, else raise a "ValueError" exception.
+        """
+        self._is_universe_domain_valid = {{ service.client_name }}._compare_universes(self.universe_domain, self.transport._credentials.universe_domain)
```
vchudnov-g:
```suggestion
        self._is_universe_domain_valid = (self._is_universe_domain_valid or
                                         {{ service.client_name }}._compare_universes(self.universe_domain, self.transport._credentials.universe_domain))
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -402,9 +466,19 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
             self._client_options = client_options_lib.ClientOptions()
         self._client_options = cast(client_options_lib.ClientOptions, self._client_options)
 
-        self._use_client_cert, self._use_mtls_endpoint = {{ service.client_name }}._read_environment_variables()
+        try:
+            universe_domain_opt = getattr(self._client_options, 'universe_domain')
+        except AttributeError:
+            warnings.warn("Attribute universe_domain does not exist in self._client_options. Setting universe_domain_opt to None.")
```
vchudnov-g: nit: I wonder whether we need this warning. If the user is passing in a dict that doesn't have this key, do we want to warn? Seems a bit annoying....and this behavior of "if it's not present, behave as though it were set to None" seems like what people would expect.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -402,9 +466,19 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
             self._client_options = client_options_lib.ClientOptions()
         self._client_options = cast(client_options_lib.ClientOptions, self._client_options)
 
-        self._use_client_cert, self._use_mtls_endpoint = {{ service.client_name }}._read_environment_variables()
+        try:
+            universe_domain_opt = getattr(self._client_options, 'universe_domain')
```
vchudnov-g, line 476: In the `__init__` docstring above, the section for `client_options` says " It won't take effect if a ``transport`` instance is provided." That should be updated, and the description should have a third clause explaining the new ``universe_domain`` property.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -581,6 +655,9 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
                 (("resource", request.resource),)),
         )
 
+        # Validate the universe domain.
+        self.transport._credentials and self._validate_universe_domain()
```
vchudnov-g: Let's move the `self.transport._credentials` check into `_validate_universe_domain`, with something in that method like

```
self._is_universe_domain_valid = ( self.transport._credentials is None or
                                   ConfigServiceV2Client._compare_universes(self.universe_domain, self.transport._credentials.universe_domain)
```


and then this line here becomes just the function call

vchudnov-g (commented):
Replying with some comments after looking at the generated `client` files. I'll keep looking at the other files, but feel free to start acting on these comments, (ping me again when the updated code is stable for review)

gapic/templates/%namespace/%name_%version/%sub/services/%service/_async_mixins.py.j2
```diff
@@ -47,6 +47,9 @@
                 (("name", request.name),)),
         )
 
+        # Validate the universe domain.
+        self.transport._credentials and self._client._validate_universe_domain()
```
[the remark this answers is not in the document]
ohmayr: addressed.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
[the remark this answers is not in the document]
ohmayr: addressed.

tests/integration/goldens/redis/google/cloud/redis_v1/services/cloud_redis/client.py
```diff
@@ -359,37 +364,95 @@ def _get_client_cert_source(provided_cert_source, use_cert_flag):
...
         """
         return self._api_endpoint
 
+    @property
+    def universe_domain(self) -> str:
+        """Return the universe domain used by the client instance.
+
+        Returns:
+            str: The universe domain used
+                by the client instance.
```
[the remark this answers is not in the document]
ohmayr: addressed.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
ohmayr: addressed.

vchudnov-g:
BTW, for future reference: we decided in a meeting earlier today that we would not yet port parts of this implementation to `api_core`, as I had suggested. We will do that as a non-breaking follow-up set of changes, since other issues are prevent quick changes to api_core at the moment and we want to release this functionality soon.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
parthea: Perhaps we could just let the error bubble up in the case that the user passes a dictionary with unsupported attributes

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -131,12 +131,17 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+    # Note: DEFAULT_ENDPOINT is deprecated. Use DEFAULT_ENDPOINT_TEMPLATE instead.
     DEFAULT_ENDPOINT = {% if service.host %}"{{ service.host }}"{% else %}None{% endif %}
 
     DEFAULT_MTLS_ENDPOINT = _get_default_mtls_endpoint.__func__(  # type: ignore
         DEFAULT_ENDPOINT
     )
 
+    DEFAULT_ENDPOINT_TEMPLATE = {% if service.host %}"{{ service.host.replace("googleapis.com", "{UNIVERSE_DOMAIN}") }}"{% else %}None{% endif %}
+    
+    _DEFAULT_UNIVERSE = "googleapis.com"
```
vchudnov-g, line 143: Suggest not hard-coding the string replacement:

```suggestion
    _DEFAULT_UNIVERSE = "googleapis.com"
    DEFAULT_ENDPOINT_TEMPLATE = {% if service.host %}"{{ service.host.replace(_DEFAULT_UNIVERSE, "{UNIVERSE_DOMAIN}") }}"{% else %}None{% endif %}
    

```

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -316,36 +322,103 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+            credentials_universe (str): The universe domain in the credentials.
+
+        Returns:
+            bool: Returns True if universes match.
+        Raises:
+            ValueError: when the universes do not match.
+        """
+        default_universe = {{ service.client_name }}._DEFAULT_UNIVERSE
+        if client_universe != credentials_universe:
+            raise ValueError(f"The configured universe domain ({client_universe}) does not match the universe domain found in the credentials ({credentials_universe}). If you haven't configured the universe domain explicitly, `{default_universe}` is the default.")
```
vchudnov-g, line 387: Consistent quoting of universes:
```suggestion
            raise ValueError(f"The configured universe domain `({client_universe})` does not match the universe domain found in the credentials `({credentials_universe})`. If you haven't configured the universe domain explicitly, `{default_universe}` is the default.")
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
ohmayr: addressed.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
ohmayr: addressed.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
ohmayr: addressed.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
ohmayr: addressed.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -316,36 +322,103 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+        """Returns True if the universe domain used by the client
+            is the same as the universe domain in credentials.
+
+        Args:
+            client_universe (str): The universe domain configured via the client options.
+            credentials_universe (str): The universe domain in the credentials.
+
+        Returns:
+            bool: Returns True if universes match.
+        Raises:
```
vchudnov-g, line 382:
```suggestion

        Raises:
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
ohmayr: addressed.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
ohmayr: addressed.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -316,36 +322,103 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+            raise ValueError(f"The configured universe domain ({client_universe}) does not match the universe domain found in the credentials ({credentials_universe}). If you haven't configured the universe domain explicitly, `{default_universe}` is the default.")
+        return True
+    
+    def _validate_universe_domain(self):
+        """Validates the universe domain used by the client instance against
+            the universe domain in the credentials.
+        
+        Returns:
+            bool: Returns True if universe domain is valid, otherwise False.
+        Raises:
```
vchudnov-g, line 396:
```suggestion

        Raises:
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -316,36 +322,103 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+    def _compare_universes(client_universe, credentials_universe):
+        """Returns True if the universe domain used by the client
+            is the same as the universe domain in credentials.
+
+        Args:
+            client_universe (str): The universe domain configured via the client options.
+            credentials_universe (str): The universe domain in the credentials.
+
+        Returns:
+            bool: Returns True if universes match.
```
vchudnov-g, line 381:
```suggestion
            bool:  True if universes match.
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -316,36 +322,103 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+        if client_universe != credentials_universe:
+            raise ValueError(f"The configured universe domain ({client_universe}) does not match the universe domain found in the credentials ({credentials_universe}). If you haven't configured the universe domain explicitly, `{default_universe}` is the default.")
+        return True
+    
+    def _validate_universe_domain(self):
+        """Validates the universe domain used by the client instance against
+            the universe domain in the credentials.
+        
+        Returns:
+            bool: Returns True if universe domain is valid, otherwise False.
```
vchudnov-g, line 395:
```suggestion
            bool: True if universe domain is valid, otherwise False.
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
```diff
@@ -54,8 +54,12 @@ class {{ service.async_client_name }}:
 
     _client: {{ service.client_name }}
 
+    # Copy defaults from the synchronous client for use here.
+    # Note: DEFAULT_ENDPOINT is deprecated. Use DEFAULT_ENDPOINT_TEMPLATE instead.
     DEFAULT_ENDPOINT = {{ service.client_name }}.DEFAULT_ENDPOINT
     DEFAULT_MTLS_ENDPOINT = {{ service.client_name }}.DEFAULT_MTLS_ENDPOINT
+    DEFAULT_ENDPOINT_TEMPLATE = {{ service.client_name }}.DEFAULT_ENDPOINT_TEMPLATE
```
vchudnov-g, line 61: Does this symbol really need to be public? If not, let's underscore it  (here and in the sync_client)

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
```diff
@@ -133,7 +133,7 @@ class {{ service.name }}GrpcTransport({{ service.name }}Transport):
             warnings.warn("client_cert_source is deprecated", DeprecationWarning)
 
         if channel:
-            # Ignore credentials if a channel was passed.
+            # Ignore credentials if a channel was passed
```
vchudnov-g, line 136:
```suggestion
            # Ignore credentials if a channel was passed.
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc_asyncio.py.j2
```diff
@@ -178,7 +178,7 @@ class {{ service.grpc_asyncio_transport_name }}({{ service.name }}Transport):
             warnings.warn("client_cert_source is deprecated", DeprecationWarning)
 
         if channel:
-            # Ignore credentials if a channel was passed.
+            # Ignore credentials if a channel was passed
```
vchudnov-g, line 181:
```suggestion
            # Ignore credentials if a channel was passed.
```

vchudnov-g (commented):
Some more incremental comments.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -556,58 +614,63 @@ def test_{{ service.client_name|snake_case }}_get_mtls_endpoint_and_cert_source(
...
-        assert client.api_endpoint == mock_api_endpoint
-
-    # If ClientOptions.api_endpoint is set and GOOGLE_API_USE_CLIENT_CERTIFICATE="false",
-    # use ClientOptions.api_endpoint as the api endpoint.
-    with mock.patch.dict(os.environ, {"GOOGLE_API_USE_CLIENT_CERTIFICATE": "false"}):
-        mock_api_endpoint = "foo"
-        options = client_options.ClientOptions(client_cert_source=mock_client_cert_source, api_endpoint=mock_api_endpoint)
-        client = client_class(client_options=options, credentials=ga_credentials.AnonymousCredentials())
-        assert client.api_endpoint == mock_api_endpoint
+        options = client_options.ClientOptions(client_cert_source=mock_client_cert_source, api_endpoint=api_override)
```
vchudnov-g, line 633: It seems like we're testing combinations of 3 paramters, `GOOGLE_API_USE_CLIENT_CERTIFICATE` `(true` or `false`), `GOOGLE_API_USE_MTLS_ENDPOINT` (`never`, `always`, `auto`) and `ClientOptions.api_endpoint` ( set or unset) for 2×3×2=12 combinations. We should list and test them systematically. For those combinations that have the same behavior, we could just choose one representative test case and add a comment about the other cases expected to behave the same way

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -556,58 +614,63 @@ def test_{{ service.client_name|snake_case }}_get_mtls_endpoint_and_cert_source(
...
+    if universe_exists:
+        options = client_options.ClientOptions(universe_domain=mock_universe)
+        client = client_class(client_options=options, credentials=_AnonymousCredentialsWithUniverseDomain())
+    else:
+        with pytest.warns(UserWarning, match="Attribute universe_domain does not exist in self._client_options. Setting universe_domain to None."):
+            client = client_class(client_options=options, credentials=_AnonymousCredentialsWithUniverseDomain())
+    assert client.api_endpoint == (mock_endpoint if universe_exists else default_endpoint)
+    assert client.universe_domain == (mock_universe if universe_exists else default_universe)
+
+    # If ClientOptions does not have a universe domain attribute and GOOGLE_API_USE_MTLS_ENDPOINT="never",
```
vchudnov-g, line 664: We could split this into a separate test function, and here too it seems like we have several options to test. Maybe the new test function should be like the current one, except that it explicitly deletes `"universe_domain"` in all test cases. Or we choose one or a few representative test cases? WDYT? I don't want to go overboard, so check me on this.

vchudnov-g (commented):
Some more comments. Looking good!

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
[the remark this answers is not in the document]
vchudnov-g: As per our off-line discussion, we'll defer refactoring some functionality into `google-api-core` until after this PR is in.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
vchudnov-g: This is passing a dictionary with an attribute missing, and it seems to me this attribute should be semantically optional (if not present, we default to it being set to `None` and use the GDU. So catching the exception seems fine; it's just the warning that I would remove.

vchudnov-g: Going with silently defaulting to `None`. Please speak up if there are objections.

parthea: You're right @vchudnov-g . No warning is needed here. I was thinking of a different code path https://github.com/googleapis/gapic-generator-python/blob/414e3cc5e4246b7bf86df6500f68bab443e50c32/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/client.py.j2#L400

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
vchudnov-g: Replicated docstring for async client (and updated other parts of the docstring that were out of date).

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
vchudnov-g: Done

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
vchudnov-g: NM. They're in parentheses, so that's fine.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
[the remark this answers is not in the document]
vchudnov-g: ACK.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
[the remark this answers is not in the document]
vchudnov-g: And I withdraw my original suggestion here. The string substitution is happening at the Jinjha template level, so what I suggested does not make sense.

That said, I would strongly encourage taking this business logic out of the templates and into the generator, but that should probably be part of a larger refactoring.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
```diff
@@ -133,8 +134,6 @@ class {{ service.name }}GrpcTransport({{ service.name }}Transport):
             warnings.warn("client_cert_source is deprecated", DeprecationWarning)
 
         if channel:
-            # Ignore credentials if a channel was passed.
```
vchudnov-g, line 136: @ohmayr I wound up removing this again, as you had originally. I convinced myself that this will not break the API surface, and we do need it for testing unless we make major changes to the architecture.

vchudnov-g: Update: now I'm putting it back, because it looks like removing it screwed u pother tests. I think we could make a case for correct behavior either way, but keeping it in is fine. And I think I was able to finesse the changes in the new test from leaving it in.

vchudnov-g:
Now that you're back, @ohmayr , LMK what you think of my recent commits. IMO, we're ready to merge this, but I certainly want an extra mind verifying my work.

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
```diff
@@ -172,22 +184,42 @@ class {{ service.async_client_name }}:
...
                 transport to use. If set to None, a transport is chosen
                 automatically.
-            client_options (ClientOptions): Custom options for the client. It
-                won't take effect if a ``transport`` instance is provided.
-                (1) The ``api_endpoint`` property can be used to override the
-                default endpoint provided by the client. GOOGLE_API_USE_MTLS_ENDPOINT
-                environment variable can also be used to override the endpoint:
+                {% if 'rest' in opts.transport and not opts.rest_numeric_enums %}
+                NOTE: "rest" transport functionality is currently in a
+                beta state (preview). We welcome your feedback via an
```
parthea, line 189: TODO: file a bug to remove the beta (preview) comment

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -316,36 +322,110 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+            if universe_domain != _default_universe:
+                raise MutualTLSChannelError(f"mTLS is not supported in any universe other than {_default_universe}.")
             api_endpoint = {{ service.client_name }}.DEFAULT_MTLS_ENDPOINT
         else:
-            api_endpoint = {{ service.client_name }}.DEFAULT_ENDPOINT
+            api_endpoint = {{ service.client_name }}._DEFAULT_ENDPOINT_TEMPLATE.format(UNIVERSE_DOMAIN=universe_domain)
         return api_endpoint
 
+    @staticmethod
+    def _get_universe_domain(client_universe_domain, universe_domain_env):
```
parthea, line 351: Add type hints for `client_universe_domain` and `client_universe_domain` as well as type hints for the return value.

tests/system/test_universe_domain.py
```diff
@@ -0,0 +1,94 @@
...
+        'content': 'Universe validation succeeded!'
+        })
+    assert response.content == "Universe validation succeeded!"
+
+
+# What we'd like to do for the following test is:
+#    @pytest.mark.parametrize("channel_creator", [grpc.insecure_channel])
+#    @pytest.mark.parametrize("transport_name",["grpc","rest"])
+#    @pytest.mark.parametrize("transport_endpoint", ["showcase.googleapis.com", "localhost:7469" ])
+#    @pytest.mark.parametrize("credential_universe",["showcase.googleapis.com", "localhost:7469"])
```
parthea, line 55: Remove commented code. Add a comment that reads like a comment instead of code. Applies throughout.

tests/system/conftest.py
```diff
@@ -107,33 +116,62 @@ def construct_client(
...
-                host="localhost:7469",
+                credentials=credentials,
+                host=transport_endpoint,
                 url_scheme="http",
             )
         else:
             raise RuntimeError(f"Unexpected transport type: {transport_name}")
 
-        return client_class(transport=transport)
+        # return client_class(transport=transport)
```
parthea, line 135: Remove commented code

tests/system/conftest.py
```diff
@@ -107,33 +116,62 @@ def construct_client(
     else:
         transport_cls = client_class.get_transport_class(transport_name)
         if transport_name in ["grpc", "grpc_asyncio"]:
+            # TODO: Need to test grpc transports without a channel_creator
```
parthea, line 119: Add link to bug for TODO comment . Applies throughout.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -316,36 +322,110 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+        if client_universe_domain is not None:
+            universe_domain = client_universe_domain
+        elif universe_domain_env is not None:
+            universe_domain = universe_domain_env
+        if len(universe_domain.strip()) == 0:
+            raise ValueError("Universe Domain cannot be an empty string.") 
+        return universe_domain
+
+    @staticmethod
+    def _compare_universes(client_universe: str,
```
parthea, line 374: Add type hint for return value

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
parthea: Please add a todo comment for the verbose test case.

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
```diff
@@ -172,22 +184,42 @@ class {{ service.async_client_name }}:
...
-                precedence if provided.
-                (2) If GOOGLE_API_USE_CLIENT_CERTIFICATE environment variable
+                use the default regular endpoint) and "auto" (auto-switch to the
+                default mTLS endpoint if client certificate is present; this is
+                the default value).
+
+                2. If the GOOGLE_API_USE_CLIENT_CERTIFICATE environment variable
                 is "true", then the ``client_cert_source`` property can be used
-                to provide client certificate for mutual TLS transport. If
+                to provide client certificate for mTLS transport. If
```
parthea, line 208:
```suggestion
                to provide a client certificate for mTLS transport. If
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -369,22 +449,32 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
-                precedence if provided.
-                (2) If GOOGLE_API_USE_CLIENT_CERTIFICATE environment variable
+                use the default regular endpoint) and "auto" (auto-switch to the
+                default mTLS endpoint if client certificate is present; this is
+                the default value).
+
+                2. If the GOOGLE_API_USE_CLIENT_CERTIFICATE environment variable
                 is "true", then the ``client_cert_source`` property can be used
-                to provide client certificate for mutual TLS transport. If
+                to provide client certificate for mTLS transport. If
```
parthea, line 468:
```suggestion
                to provide a client certificate for mTLS transport. If
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -369,22 +449,32 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
                 not provided, the default SSL client certificate will be used if
                 present. If GOOGLE_API_USE_CLIENT_CERTIFICATE is "false" or not
                 set, no client certificate will be used.
+                
+                3. The ``universe_domain`` property can be used to override the
+                default "googleapis.com" universe. Note that ``api_endpoint``
```
parthea, line 474:
```suggestion
                default "googleapis.com" universe. Note that the ``api_endpoint``
```

parthea (approved):
LGTM. Added minor observations

tests/system/test_universe_domain.py
vchudnov-g: Drive-through comment: the comment would be about not getting this code to work, so including the code seems the most concise way. Maybe we can expand the preamble, saying the this attempted parametrization stacking was not working.

parthea: @ohmayr , Please could you take a look to see if we can achieve the same result using `@pytest.mark.parametrize` similar to how we do it here? Let me know if you'd like to pair program on it

https://github.com/googleapis/gapic-generator-python/blob/2e86a654896054f1ac4b6833127bb095d2bf5b90/tests/integration/goldens/asset/tests/unit/gapic/asset_v1/test_asset_service.py#L223-L227

parthea (approved):

PR #1931, opened 2024-02-06
fix: fix `ValueError` in `test__validate_universe_domain`

### LINKED ISSUE #1791: Add support for pre-release testing of dependencies
There is a `prerelease_deps` nox session for packages in `google-cloud-python`, [for example](https://github.com/googleapis/google-cloud-python/blob/9f0608987b8b72f9e613b170b91b6ab40493287f/packages/google-cloud-access-approval/noxfile.py#L340), that tests against pre-release versions of direct dependencies. We should create a similar presubmit check in this repository that tests pre-release versions of direct dependencies in https://github.com/googleapis/gapic-generator-python/blob/main/requirements.in. For example, when a new pre-release version of `google-api-core` is published, there should be a prerelease presubmit check in this repo that run tests against it.

### LINKED ISSUE #1930: Autogenerated tests failing in `python-bigquery-storage`
See PR https://github.com/googleapis/python-bigquery-storage/pull/742 where the autogenerated tests are failing in the `Kokoro Prerelease Dependencies ` presubmit. See the stack trace below:

```
            # Test the case when there is a universe mismatch from the client.
            #
            # TODO: Make this test unconditional once the minimum supported version of
            # google-api-core becomes 2.15.0 or higher.
>           api_core_major, api_core_minor, _ = [
                int(part) for part in api_core_version.__version__.split(".")
            ]

tests/unit/gapic/bigquery_storage_v1beta2/test_big_query_read.py:342: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

.0 = 

    api_core_major, api_core_minor, _ = [
>       int(part) for part in api_core_version.__version__.split(".")
    ]
E   ValueError: invalid literal for int() with base 10: '0rc0'

tests/unit/gapic/bigquery_storage_v1beta2/test_big_query_read.p
```
The `Kokoro Prerelease Dependencies ` tests autogenerated code against pre-release versions of dependencies. The issue started when `google-api-core` `2.17.0rc0` was released. We need to update this code to work with rc versions.

https://pypi.org/project/google-api-core/2.17.0rc0/

See related issue https://github.com/googleapis/gapic-generator-python/issues/1791

parthea:
Fixes #1791 🦕
Fixes #1930 🦕

See commit https://github.com/googleapis/gapic-generator-python/pull/1931/commits/0bb4dcd52f91892d9fac358feb29c7db7c7c33ea where stacktrace from #1930 appears in `goldens-prerelease` presubmit: https://github.com/googleapis/gapic-generator-python/actions/runs/7800448991/job/21273305989?pr=1931

### DIFF
--- a/gapic/templates/noxfile.py.j2
+++ b/gapic/templates/noxfile.py.j2
@@ -4,6 +4,7 @@
 
 import os
 import pathlib
+import re
 import shutil
 import subprocess
 import sys
@@ -38,6 +39,7 @@ nox.sessions = [
     "docs",
     "blacken",
     "lint",
+    "prerelease_deps",
 ]
 
 @nox.session(python=ALL_PYTHON)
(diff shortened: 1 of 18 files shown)


ohmayr (approved):

tests/integration/goldens/asset/tests/unit/gapic/asset_v1/test_asset_service.py
```diff
@@ -206,7 +206,7 @@ def test__validate_universe_domain(client_class, transport_class, transport_name
     # TODO: This is needed to cater for older versions of google-auth
     # Make this test unconditional once the minimum supported version of
     # google-auth becomes 2.23.0 or higher.
-    google_auth_major, google_auth_minor, _ = [int(part) for part in google.auth.__version__.split(".")]
+    google_auth_major, google_auth_minor = [int(part) for part in google.auth.__version__.split(".")[0:2]]
```
ohmayr, line 209: @parthea What's this change for?

gapic/templates/noxfile.py.j2
```diff
@@ -58,6 +60,80 @@ def unit(session):
         os.path.join('tests', 'unit', ''.join(session.posargs))
     )
 
+@nox.session(python=ALL_PYTHON[-1])
```
ohmayr, line 63: @parthea Can you explain a scenario in which this session would fail?

tests/integration/goldens/asset/tests/unit/gapic/asset_v1/test_asset_service.py
parthea: This is the fix for issue https://github.com/googleapis/gapic-generator-python/issues/1930. We don't use the minor version in the check so we don't need to parse it.

gapic/templates/noxfile.py.j2
parthea: See https://github.com/googleapis/gapic-generator-python/issues/1930 which has details on the issue where a pre-release test failed downstream but we do not see the issue in this repository because we're not testing pre-release versions.

PR #1942, opened 2024-02-13
fix: require google-api-core 1.34.1

### LINKED ISSUE #1940: rpc in REST Streaming Fragment test is missing the http annotation
There is a fragment test for testing REST streaming however the REST code would not be exercised because the http annotation is missing in the rpc.

https://github.com/googleapis/gapic-generator-python/blob/90ba3f7738aa18476ba896e3097c1fde43a815d9/tests/fragments/test_rest_streaming.proto#L24-L27

parthea:
test: fix rest streaming fragment test
chore: update dependencies in requirements.txt

Fixes https://github.com/googleapis/gapic-generator-python/issues/1940 🦕


parthea:
```
self = <google.api_core.rest_streaming.ResponseIterator object at 0x7ff861cdca50>

    def _grab(self):
        # Add extra quotes to make json.loads happy.
>       return self._response_message_cls.from_json(self._ready_objs.popleft())
E       AttributeError: from_json
```
This issue was addressed in https://github.com/googleapis/python-api-core/pull/604

ohmayr (approved):

tests/fragments/google/api/httpbody.proto
```diff
@@ -0,0 +1,81 @@
+// Copyright 2023 Google LLC
```
vchudnov-g, line 1: Date!

parthea: This file was copied from https://github.com/googleapis/googleapis/blob/master/google/api/httpbody.proto which is why I kept the date. It's not clear if we should update the date when the file is copied from another repository.

PR #1959, opened 2024-02-27
chore: update copyright year

vchudnov-g:
BEGIN_COMMIT_OVERRIDE
fix: update copyright year
END_COMMIT_OVERRIDE


vchudnov-g:
This won't have issues when updating previously released files, will it? I don't think we've had issues with previous years....

parthea (approved):

PR #1975, opened 2024-03-13
feat: Add support for reading MethodSettings from service configuration YAML

parthea:
Adds support for reading Method settings in the service configuration YAML

https://github.com/googleapis/googleapis/blob/7dab3de7ec79098bb367b6b2ac3815512a49dd56/google/api/client.proto#L325

Towards b/322910372


gapic/schema/api.py
```diff
@@ -560,6 +561,119 @@ def mixin_http_options(self):
...
+
+            # The field name must be listed in the google.api.MethodSettings.auto_populated_fields entry in google.api.Publishing.method_settings for the target method.
+            auto_populated_fields = [
+                field for field in method_settings.auto_populated_fields
+            ]
+            for field in auto_populated_fields:
+                if field not in _get_auto_populated_fields_candidates(method_in_scope):
+                    raise ValueError(
+                        f"Field {field} is not valid as an auto populated field for the top level request message of selector {method_settings.selector}"
+                    )
```
ohmayr, line 662: How about something like this for optimization?

```python3
candidates = _get_auto_populated_fields_candidates(method_in_scope)
for field in auto_populated_fields:
    if field not in candidates:
        raise ValueError  
```

gapic/schema/api.py
```diff
@@ -560,6 +561,119 @@ def mixin_http_options(self):
...
+                raise ValueError(
+                    f"Selector {method_settings.selector} is not a valid method in this API."
+                )
+
+            if (
+                method_in_scope.client_streaming or method_in_scope.server_streaming
+            ) and method_settings.auto_populated_fields:
+                raise ValueError(
+                    f"Selector {method_settings.selector} is a streaming rpc. `auto_populated_fields` are only supported in unary rpcs."
+                )
```
ohmayr, line 652: My suggestion would be to split up this function and have a separate function for each check.

```python3
def is_streaming_rpc(method_in_scope)
    if ( 
    method_in_scope.client_streaming or
    method_in_scope.server_streaming
    ) and
    method_settings.auto_populated_fields:
      raise ValueError(...)
    return False
```

Overall I do understand how the code flow works, but I find it hard to follow through the tests due to the nested functions. Not necessary, but my suggestion would be to move these functions out of `get_all_methods` and test each piece individually which in my opinion will be much easier to follow through!
Similarly, for the code below, we could have:
```python3
def is_top_level_method():
   # add logic to check top level field
```

parthea: I refactored the code in https://github.com/googleapis/gapic-generator-python/pull/1975/commits/41c582c6a1f87481b94b237bc76a65eea0273b0c. Please take another look

gapic/schema/api.py
parthea: Fixed in https://github.com/googleapis/gapic-generator-python/pull/1975/commits/9f5c5d56fd7fc6860f369deb2429a7236991c521

gapic/schema/api.py
```diff
@@ -560,6 +561,133 @@ def mixin_http_options(self):
...
+                populated, taking into account the requirements
+                from AIP-4235.
+                https://google.aip.dev/client-libraries/4235
+
+        Raises:
+            ValueError: if `auto_populated_fields` are incorrectly set in the service config
+        """
+        auto_populated_fields = []
+
+        def _check_service_yaml_validity(method_in_scope: MethodDescriptorProto, method_settings: client_pb2.MethodSettings):
```
vchudnov-g, line 615: What's the advantage of having this as a nested function? It's not terribly short, so breaking it out and clarifying the name might its purpose clearer.

gapic/schema/api.py
```diff
@@ -560,6 +561,133 @@ def mixin_http_options(self):
...
+            https://google.aip.dev/client-libraries/4235
+            """
+
+            # The method must not be a streaming rpc
+            if method_in_scope.client_streaming or method_in_scope.server_streaming:
+                raise ValueError(
+                    f"Selector {method_settings.selector} is a streaming rpc. `auto_populated_fields` are only supported in unary rpcs."
+                )
+            # The field name must be listed in the google.api.MethodSettings.auto_populated_fields entry in google.api.Publishing.method_settings for the target method.
+            candidates = self.get_auto_populated_fields_candidates(
```
vchudnov-g, line 629: I'm curious why you are compiling a list of candidates first, and then checking each field against the list. Why not do below something like?

```
for field in auto_populated_fields: 
  if not can_make_field_optional(method_in_scope, field):
    raise ValueError(...)
```

I'm asking mostly out of curiosity. IMO that's a bit clearer, but it's probably OK either way. WDYT?

gapic/schema/api.py
```diff
@@ -560,6 +561,133 @@ def mixin_http_options(self):
...
+        message: MethodDescriptorProto,
+    ) -> Sequence[str]:
+        """Return fields which meet the criteria to be automatically populated for a given message.
+        For a field to be automatically populated, all the below configurations must
+        be true according to https://google.aip.dev/client-libraries/4235:
+
+        - The field must be of type string
+        - The field must be at the top-level of the request message
+        - The field must not be annotated with google.api.field_behavior = REQUIRED.
+        - The field must be annotated with google.api.field_info.format = UUID4.
```
vchudnov-g, line 575: Add
```
        The field presence requirements in that AIP are checked at run time.
```

Right?

vchudnov-g (commented):
Some initial comments, all minor (nothing blocking).

I haven't yet looked at the test cases,  Will do that next.

gapic/schema/api.py
```diff
@@ -560,6 +561,133 @@ def mixin_http_options(self):
             res[s] = [rule for rule in opt_gen if rule]
         return res
 
+    def get_auto_populated_fields_candidates(
+        self,
+        message: MethodDescriptorProto,
```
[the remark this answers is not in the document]
parthea: Fixed in 12f4333d8881e0fd284199ae375b599185ce6ddd

gapic/schema/api.py
```diff
@@ -560,6 +561,133 @@ def mixin_http_options(self):
...
+    ) -> Sequence[str]:
+        """Return fields which meet the criteria to be automatically populated for a given message.
+        For a field to be automatically populated, all the below configurations must
```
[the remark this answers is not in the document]
parthea: Fixed in 12f4333d8881e0fd284199ae375b599185ce6ddd

gapic/schema/api.py
```diff
@@ -560,6 +561,133 @@ def mixin_http_options(self):
             res[s] = [rule for rule in opt_gen if rule]
         return res
 
+    def get_auto_populated_fields_candidates(
+        self,
+        message: MethodDescriptorProto,
+    ) -> Sequence[str]:
+        """Return fields which meet the criteria to be automatically populated for a given message.
```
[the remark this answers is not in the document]
parthea: Fixed in https://github.com/googleapis/gapic-generator-python/commit/12f4333d8881e0fd284199ae375b599185ce6ddd

gapic/schema/api.py
```diff
@@ -560,6 +561,133 @@ def mixin_http_options(self):
...
+            and not field.required
+            and field.uuid4
+        ]
+
+    def get_auto_populated_fields(
+        self,
+        method_settings: client_pb2.MethodSettings,
+    ) -> Sequence[str]:
+        """
+        Return a list of auto-populated fields based on the service YAML.
```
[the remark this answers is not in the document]
parthea: Fixed in b12753d4afd12aaef4a98d7f4633ee5a188a5a27

gapic/schema/api.py
```diff
@@ -560,6 +561,133 @@ def mixin_http_options(self):
...
+
+        Raises:
+            ValueError: if `auto_populated_fields` are incorrectly set in the service config
+        """
+        auto_populated_fields = []
+
+        def _check_service_yaml_validity(method_in_scope: MethodDescriptorProto, method_settings: client_pb2.MethodSettings):
+            """
+            Ensure that the method and fields meet the criteria of AIP-4235.
+            Use the helper function get_auto_populated_fields_candidates() to return on fields
```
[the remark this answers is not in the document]
parthea: Fixed in https://github.com/googleapis/gapic-generator-python/pull/1975/commits/8d64606a88f54f6d641fb958327adab48a62798b

gapic/schema/api.py
```diff
@@ -560,6 +561,133 @@ def mixin_http_options(self):
...
+
+        if method_settings.auto_populated_fields:
+            _check_service_yaml_validity(method_in_scope, method_settings)
+            auto_populated_fields = method_settings.auto_populated_fields
+
+        return auto_populated_fields
+
+    @cached_property
+    def all_methods(self) -> Mapping[str, MethodDescriptorProto]:
+        """Return a map of all API-wide methods.
```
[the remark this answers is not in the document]
parthea: Fixed in https://github.com/googleapis/gapic-generator-python/pull/1975/commits/e7f8a02f0863a8849cba93b872c142b97e816ddb

I used the same terminology from here but `Return a map of all methods for the API.` reads better. 
https://github.com/googleapis/gapic-generator-python/blob/d5f90a2455cb671e67e2cd30f3eb470092e4f889/gapic/schema/api.py#L408-L410

gapic/schema/api.py
parthea: Fixed in https://github.com/googleapis/gapic-generator-python/commit/12f4333d8881e0fd284199ae375b599185ce6ddd

gapic/schema/api.py
parthea: c3563c18660e4bef34f62381f91ba22f0e42f5a8 includes a major refactor where all of the checks are added under `check_method_settings_validity` and this code no longer exists

gapic/schema/api.py
parthea: Fixed in c3563c18660e4bef34f62381f91ba22f0e42f5a8

gapic/schema/api.py
```diff
@@ -560,6 +561,116 @@ def mixin_http_options(self):
...
+        is set, verify each field against the criteria of AIP-4235
+        (https://google.aip.dev/client-libraries/4235). All the configurations 
+        below must be true:
+
+        - The field must be of type string
+        - The field must be at the top-level of the request message
+        - The RPC must be a unary RPC (i.e. streaming RPCs are not supported)
+        - The field must not be annotated with google.api.field_behavior = REQUIRED.
+        - The field must be annotated with google.api.field_info.format = UUID4.
+        - The field presence requirements in AIP-4235 are checked at run time.
```
vchudnov-g, line 593: Nit: this item doesn't read right under a list introduced as 
'All conditions below must be true". I say take take this out of the list (outdent it)

```suggestion
        Note that the field presence requirements in AIP-4235 are checked at run time.
```

gapic/schema/api.py
```diff
@@ -560,6 +561,116 @@ def mixin_http_options(self):
...
+        }
+
+    def check_method_settings_validity(
+        self, service_method_settings: Sequence[client_pb2.MethodSettings]
+    ) -> None:
+        """
+        Checks each `google.api.client.MethodSettings` for validity. If
+        `google.api.client.MethodSettings.auto_populated_fields`
+        is set, verify each field against the criteria of AIP-4235
+        (https://google.aip.dev/client-libraries/4235). All the configurations
```
vchudnov-g, line 585:
```suggestion
        (https://google.aip.dev/client-libraries/4235). All the conditions 
```

gapic/schema/api.py
```diff
@@ -560,6 +561,116 @@ def mixin_http_options(self):
...
+                        if not field.uuid4:
+                            selector_error_message += f"Field `{field_str}` is not a UUID4 field. "
+
+                if selector_error_message:
+                    all_errors += selector_error_message
+                    invalid_selectors.append(method_settings.selector)
+
+        if all_errors:
+            raise ValueError(
+                f"Fields cannot be automatically populated in the top level "
```
vchudnov-g, line 642: This error header text should be more general. For example, maybe the only error condition is the `not method_descriptor` above, and nothing in the config even intended to have auto-populated fields.

gapic/schema/api.py
```diff
@@ -560,6 +561,116 @@ def mixin_http_options(self):
...
+                            selector_error_message += f"Field `{field_str}` is a required field. "
+                        if not field.uuid4:
+                            selector_error_message += f"Field `{field_str}` is not a UUID4 field. "
+
+                if selector_error_message:
+                    all_errors += selector_error_message
+                    invalid_selectors.append(method_settings.selector)
+
+        if all_errors:
+            raise ValueError(
```
vchudnov-g, line 641: Consider defining a sub-type of `ValueError` and raising that. Maybe `MethodSettingsError` or `ConfigurationError`?

Not strictly necessary; up to you. It would make it easier for us and users to test for the error condition without brittle examination of the error text.

And the custom type gives us the option to internally keep the various individual errors in a list, so in our tests  we could check each element of the list contains key parts (eg method selector and `"UUID4"`). In fact, you could have `invalid_selectors` be a list of pairs, where the second element of the pair is the list of error strings for that selector.  I.e. by doing `invalid_selectors.append((method_settings.selector, errors_for_this_method))` the structure could be (in loose pseudo-yaml form)

```
invalid_selectors:
  - google.example.v1beta1.SomeExample.DoFoo"
     - "Field `auto_fill_me` is a required field"
     - "Field `auto_fill_me` is not a UUID4 field"
  - google.example.v1beta1.SomeExample.DoBar
     - "Field `request_id` is not of type string"
     - "Field `request_id` is a required field"
```


(To be clear, I do NOT think we should define custom error types for each of these conditions. That seems like too much pedantic overhead at this point. Nor should we use error chaining. But the single custom error type with some internal structure seems reasonable to me.)

tests/unit/schema/test_api.py
```diff
@@ -2602,3 +2604,215 @@ def test_has_iam_mixin():
...
+                            "mollusc",
+                        ],
+                    }
+                ]
+            },
+        }
+    )
+    api_schema = api.API.build(fd, "google.example.v1beta1", opts=opts)
+
+    exception_message = """Fields cannot be automatically populated in the top level \
```
vchudnov-g, line 2805: We shouldn't compare exact text of error messages. We don't want users to rely on that and we shouldn't either. I suggest checking the custom error type (if you go with my other suggestion) and then checking that the error message contains the relevant strings (eg selector strings defined in this test).

gapic/schema/api.py
```diff
@@ -560,6 +561,116 @@ def mixin_http_options(self):
...
+        for method_settings in service_method_settings:
+            method_descriptor = self.all_methods.get(method_settings.selector)
+            if not method_descriptor:
+                all_errors += f"Selector {method_settings.selector} is not a valid method in this API. "
+                invalid_selectors.append(method_settings.selector)
+                continue
+            if method_settings.auto_populated_fields:
+                # The method must not be a streaming method.
+                if method_descriptor.client_streaming or method_descriptor.server_streaming:
+                    all_errors += f"Selector {method_settings.selector} is a streaming method. `auto_populated_fields` are only supported for unary methods. "
```
vchudnov-g, line 616:
```suggestion
                    all_errors += f"Selector {method_settings.selector} is a streaming method. Only unary methods support the `auto_populated_fields` setting. "
```

(Just so we can capitalize the first word of the second sentence ;-) )

gapic/schema/api.py
```diff
@@ -560,6 +561,116 @@ def mixin_http_options(self):
...
+            ValueError: if fields in `method_settings.auto_populated_fields` cannot be
+                automatically populated.
+        """
+
+        all_errors: str = ""
+        invalid_selectors = []
+        for method_settings in service_method_settings:
+            method_descriptor = self.all_methods.get(method_settings.selector)
+            if not method_descriptor:
+                all_errors += f"Selector {method_settings.selector} is not a valid method in this API. "
```
vchudnov-g, line 610: These `+=` result in run-on text. At the very least, insert `"\n"`.

I suggest having `all_errors` be a list you `.append()` to, and in the end the exception message can `"\n".join(all_errors)`. This will work nicely with my suggestion for a custom exception and testing that exception.

tests/unit/schema/test_api.py
```diff
@@ -2602,3 +2604,215 @@ def test_has_iam_mixin():
...
+        name="mollusc", type="TYPE_STRING", options=field_options, number=2
+    )
+    # Also include a required field
+    field_options.Extensions[field_behavior_pb2.field_behavior].append(
+        field_behavior_pb2.FieldBehavior.Value("REQUIRED")
+    )
+    clam = make_field_pb2(
+        name="clam", type="TYPE_STRING", options=field_options, number=3
+    )
+    # Also include a field which is not of type string
```
vchudnov-g, line 2635: This tests both non-string and a non-UUID4. I suggest separate test cases.

tests/unit/schema/test_api.py
```diff
@@ -2602,3 +2604,215 @@ def test_has_iam_mixin():
...
+    api_schema = api.API.build(fd, "google.example.v1beta1", opts=opts)
+
+    exception_message = """Fields cannot be automatically populated in the top level \
+request message for selectors: \['google.example.v1beta1.DoesNotExist.Example1', \
+'google.example.v1beta1.SomeExample.Example1', \
+'google.example.v1beta1.SomeExample.Example2', \
+'google.example.v1beta1.SomeExample.Example3'\]. Selector \
+google.example.v1beta1.DoesNotExist.Example1 is not a valid method in this API. \
+Field `doesnotexist` is not in the top level request message. Field `octopus` is not \
+of type string. Field `octopus` is a required field. Field `octopus` is not a UUID4 \
```
vchudnov-g, line 2812: Field `octopus` is testing several failing conditions at once. I think we we should test fields with one failing condition at a time. I think a better, accepted practice is to test simple cases and assume if those work then combinations ought to work too, rather than test a combination case and assume that if it works, its constituent parts will then work by themselves.

(The full coverage solution is to test all combinations, but I think that's overkill here.)

gapic/schema/api.py
```diff
@@ -560,6 +561,116 @@ def mixin_http_options(self):
...
+            Mapping[str, MethodDescriptorProto]: A mapping of MethodDescriptorProto
+                values for the API.
+        """
+        return {
+            f"{service_key}.{method_key}": method_value
+            for service_key, service_value in self.services.items()
+            for method_key, method_value in service_value.methods.items()
+        }
+
+    def check_method_settings_validity(
```
vchudnov-g, line 578: nit nit suggestion: `def enforce _valid_method_settings`

ohmayr (approved):

gapic/schema/api.py
parthea: Fixed in d5d0aee518514cf10a1474ca5ad41aadf098500f

gapic/schema/api.py
parthea: Fixed in 5a2d33b8b160a246945e3752e145878581a351b1

gapic/schema/api.py
parthea: Fixed in 5a2d33b8b160a246945e3752e145878581a351b1

gapic/schema/api.py
parthea: Fixed in 5a2d33b8b160a246945e3752e145878581a351b1

tests/unit/schema/test_api.py
parthea: Fixed in 5a2d33b8b160a246945e3752e145878581a351b1

gapic/schema/api.py
parthea: Fixed in 5a2d33b8b160a246945e3752e145878581a351b1

gapic/schema/api.py
parthea: Fixed in 5a2d33b8b160a246945e3752e145878581a351b1

tests/unit/schema/test_api.py
parthea: Fixed in 5a2d33b8b160a246945e3752e145878581a351b1

tests/unit/schema/test_api.py
parthea: Fixed in 5a2d33b8b160a246945e3752e145878581a351b1

gapic/schema/api.py
parthea: Fixed in 5a2d33b8b160a246945e3752e145878581a351b1

tests/unit/schema/test_api.py
```diff
@@ -2602,3 +2607,374 @@ def test_has_iam_mixin():
     })
     api_schema = api.API.build(fd, 'google.example.v1', opts=opts)
     assert api_schema.has_iam_mixin
+
+
+def get_file_descriptor_proto_for_method_settings_tests(
```
vchudnov-g, line 2612: Thanks for factoring this out.  It's much easier to follow!


Minor, non-blocking: I thought I left the following comment when I reviewed last week, but I don't see it on GitHub so it must not have gotten saved. I'll put it here just for your consideration. It's less pressing now that you've done this refactoring, but it would make it nicer when defining the various fields, etc.

If it's trivial to read from a proto "file" (string) in tests, it might be simpler to simply encode the protofile inline rather than calling `make_file_pb2` and it's ilk... something like
```
proto_descriptor = make_protodescriptor("""
package xx;

message Foo { 
  string foo = 1;
}
""")
```

(This is similar in spirit to what you do with the YAML file below!)

This would make it easier to keep the proto concepts being tested next to the code without a lot of boilerplate. If it's not trivial to do now, let's defer it. (But I personally would love to eventually have most of our tests use this declarative style rather than the harder-to-follow procedural set-up of they test cases.)

parthea:
@vchudnov-g Please could you take another look?

tests/unit/schema/test_api.py
```diff
@@ -2602,3 +2607,374 @@ def test_has_iam_mixin():
...
+        make_file_pb2(
+            name="someexample.proto",
+            package="google.example.v1beta1",
+            messages=(
+                make_message_pb2(name="ExampleRequest", fields=fields),
+                make_message_pb2(name="ExampleResponse", fields=()),
+            ),
+            services=(
+                descriptor_pb2.ServiceDescriptorProto(
+                    name="SomeExample",
```
vchudnov-g, line 2638: nit, non-blocking: consider more descriptive names, like `ServiceOne` and `ServiceTwo` for theservices, and `DoFoo` for the methods. Just to make it easier for humans to track which names are used where. (`ExampleRequest` and `ExampleResponse` do that, which is great.)

tests/unit/schema/test_api.py
```diff
@@ -2602,3 +2607,374 @@ def test_has_iam_mixin():
...
+def test_method_settings_invalid_selector_raises_error():
+    """
+    Test that `MethodSettingsError` when `client_pb2.MethodSettings.selector`
+    cannot be mapped to a method in the API.
+    """
+    fd = get_file_descriptor_proto_for_method_settings_tests()
+    api_schema = api.API.build(fd, "google.example.v1beta1")
+    methodsettings = [
+        client_pb2.MethodSettings(
+            selector="google.example.v1beta1.DoesNotExist.Example1",
```
vchudnov-g, line 2757: nit: Do you want to test a non-existent method in a valid service as well?

tests/unit/schema/test_api.py
```diff
@@ -2602,3 +2607,374 @@ def test_has_iam_mixin():
...
+        api_schema.enforce_valid_method_settings(methodsettings)
+
+
+def test_method_settings_auto_populated_field_not_found_raises_error():
+    """
+    Test that `MethodSettingsError` is raised when a field in
+    `client_pb2.MethodSettings.auto_populated_fields` is not found in the top-level
+    request message of the selector.
+    """
+    fd = get_file_descriptor_proto_for_method_settings_tests()
```
vchudnov-g, line 2790: Consider populating some other field of the message.

Also, is it simple to test when the field in `auto_populated_fields` is nested one level in ? (I.e., it still exists but not at the top-level, so we still get the exception)

vchudnov-g (approved):
Looks good!

I have one suggestion about using YAML rather than JSON in error text, and some more minor, non-blocking suggestions about possible enhancements.

gapic/schema/api.py
```diff
@@ -575,74 +584,91 @@ def all_methods(self) -> Mapping[str, MethodDescriptorProto]:
...
+                            )
+                if selector_errors:
+                    all_errors[method_settings.selector] = selector_errors
         if all_errors:
-            raise ValueError(
-                f"Fields cannot be automatically populated in the top level "
-                f"request message for selectors: {invalid_selectors}. "
-                f"{all_errors.strip()}"
-            )
+            raise MethodSettingsError(json.dumps(all_errors))
```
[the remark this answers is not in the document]
parthea: Fixed in https://github.com/googleapis/gapic-generator-python/pull/1975/commits/4161702a982d0f4c3bd57fe7a393f797b1fe9c4f and https://github.com/googleapis/gapic-generator-python/pull/1975/commits/e1c87858bd12ea813fce637ff23872d6ebb49816

gapic/schema/api.py
```diff
@@ -575,74 +584,91 @@ def all_methods(self) -> Mapping[str, MethodDescriptorProto]:
...
         """
 
-        all_errors: str = ""
-        invalid_selectors = []
+        all_errors: dict = {}
+        selectors_seen = []
         for method_settings in service_method_settings:
+            # Check if this selector is defind more than once
+            if method_settings.selector in selectors_seen:
+                all_errors[method_settings.selector] = ["Duplicate selector"]
```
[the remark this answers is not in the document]
parthea: If something wrong with the selector (it is a duplicate, doesn't map to a method or maps to a streaming method), I  decided not to continue checking the validity of the auto-populated fields. The issue with the selector should be fixed and providing additional detail in the error message on the auto-populated fields would take away from the more important error with the selector.

tests/unit/schema/test_api.py
parthea: I created https://github.com/googleapis/gapic-generator-python/issues/1986 as a follow up issue to explore whether this is feasible

tests/unit/schema/test_api.py
parthea: Fixed in https://github.com/googleapis/gapic-generator-python/pull/1975/commits/04fa73e4ab7801ddfdd0bdca34e47867b375a7fb

tests/unit/schema/test_api.py
parthea: Fixed in https://github.com/googleapis/gapic-generator-python/pull/1975/commits/eac51b69af4a2f3b03102a3e609f62e442a52e3e

tests/unit/schema/test_api.py
parthea: Done in https://github.com/googleapis/gapic-generator-python/pull/1975/commits/1ab06eae2427367f3bdf5f7ce5dc659345265e2b

PR #1989, opened 2024-03-21
fix: set `default` argument of `jinja-filters.map()` when looking up attributes

parthea:
When[ jinja-filters.map](https://jinja.palletsprojects.com/en/3.0.x/templates/#jinja-filters.map) is used to look up an attribute, the `default` argument should be set in case an object in the list does not have the given attribute.


vchudnov-g (approved):
Thanks for doing this.

It might be good to add to the PR description why we're doing this. Though I haven't found a statement to this effect in the official Jinja documentation, [this issue](https://stackoverflow.com/q/65674651/23828460) suggests that if the key isn't set and we don't set a `default`, then `map` returns a `LookupError`.

ohmayr (approved):

Issue #791, opened 2021-02-26
'empty_call' unit tests failing on some libraries


busunkim96:
https://source.cloud.google.com/results/invocations/a8592aaa-9284-46c8-b2a1-0abbecbd3e08/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-secret-manager%2Fpresubmit%2Fpresubmit/log

```
=========================== short test summary info ============================
FAILED tests/unit/gapic/secretmanager_v1/test_secret_manager_service.py::test_set_iam_policy_empty_call
FAILED tests/unit/gapic/secretmanager_v1/test_secret_manager_service.py::test_get_iam_policy_empty_call
FAILED tests/unit/gapic/secretmanager_v1/test_secret_manager_service.py::test_test_iam_permissions_empty_call
FAILED tests/unit/gapic/secretmanager_v1beta1/test_secret_manager_service.py::test_set_iam_policy_empty_call
FAILED tests/unit/gapic/secretmanager_v1beta1/test_secret_manager_service.py::test_get_iam_policy_empty_call
FAILED tests/unit/gapic/secretmanager_v1beta1/test_secret_manager_service.py::test_test_iam_permissions_empty_call
6 failed, 432 passed in 6.87s
```

```
________________________ test_set_iam_policy_empty_call ________________________

    def test_set_iam_policy_empty_call():
        # This test is a coverage failsafe to make sure that totally empty calls,
        # i.e. request == None and no flattened fields passed, work.
        client = SecretManagerServiceClient(
            credentials=credentials.AnonymousCredentials(), transport="grpc",
        )

        # Mock the actual call within the gRPC stub, and fake the request.
        with mock.patch.object(type(client.transport.set_iam_policy), "__call__") as call:
>           client.set_iam_policy()

tests/unit/gapic/secretmanager_v1/test_secret_manager_service.py:3539:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <google.cloud.secretmanager_v1.services.secret_manager_service.client.SecretManagerServiceClient object at 0x7fe0d352c850>
request = None

    def set_iam_policy(
        self,
        request: iam_policy.SetIamPolicyRequest = None,
        *,
        retry: retries.Retry = gapic_v1.method.DEFAULT,
        timeout: float = None,
[...]

PR #1997, opened 2024-03-22
fix: cater for empty async call

parthea:
This PR fixes the issue in the stack trace below seen in cl/618127799 which is similar to #791 but for the async client

```
            # Certain fields should be provided within the metadata header;
            # add these here.
            metadata = tuple(metadata) + (
                gapic_v1.routing_header.to_grpc_metadata((
>                   ("resource", request.resource),
                )),
            )
E           AttributeError: 'NoneType' object has no attribute 'resource'
```

### DIFF
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
@@ -326,7 +326,7 @@ class {{ service.async_client_name }}:
         if isinstance(request, dict):
             request = {{ method.input.ident }}(**request)
         elif not request:
-            request = {{ method.input.ident }}({% if method.flattened_fields %}{% for f in method.flattened_fields.values() %}{{ f.name }}={{ f.name }}, {% endfor %}{% endif %})
+            request = {{ method.input.ident }}({% if method.flattened_fields %}{% for f in method.flattened_fields.values() %}{{ f.name }}={{ f.name }}{% if not loop.last %}, {% endif %}{% endfor %}{% endif %})
         {% else %}
         request = {{ method.input.ident }}(request)
         {% endif %} {# different request package #}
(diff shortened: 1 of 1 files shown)


gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
```diff
@@ -325,10 +325,8 @@ class {{ service.async_client_name }}:
         # so it must be constructed via keyword expansion.
         if isinstance(request, dict):
             request = {{ method.input.ident }}(**request)
-        {% if method.flattened_fields %}{# Cross-package req and flattened fields #}
         elif not request:
-            request = {{ method.input.ident }}({% if method.input.ident.package != method.ident.package %}{% for f in method.flattened_fields.values() %}{{ f.name }}={{ f.name }}, {% endfor %}{% endif %})
-        {% endif %}{# Cross-package req and flattened fields #}
+            request = {{ method.input.ident }}({% if method.flattened_fields %}{% for f in method.flattened_fields.values() %}{{ f.name }}={{ f.name }}, {% endfor %}{% endif %})
         {% else %}
```
ohmayr, line 330: Do we still need this j2 else case here?

ohmayr (approved):

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
parthea: Yes, the else case is needed for when`{% if method.input.ident.package != method.ident.package %}` is false and also exists in the sync version here

https://github.com/googleapis/gapic-generator-python/blob/23073450ca2b27bf08a8f005682007efb98fbefa/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/_client_macros.j2#L106-L113

PR #2051, opened 2024-06-24
fix: missing DEFAULT_HOST should still result in compiling code

achew22:
Previously the code would emit

```python
  DEFAULT_HOST: str =
```

Note that there is nothing at the end of the line. Now it emits

```python
  DEFAULT_HOST: str = ''
```

Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/gapic-generator-python/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Fixes #<issue_number_goes_here> 🦕


vchudnov-g (approved):
Thanks for contributing this fix!

PR #2079, opened 2024-07-27
chore: allow custom routing header metadata

### LINKED ISSUE #2078: support customizing routing metadata
The gapic generator [automatically attaches routing metadata](https://github.com/googleapis/python-bigtable/blob/0bddf11cdf9f9e6574b45fc142b7c37f0b57efc2/google/cloud/bigtable_v2/services/bigtable/async_client.py#L1382) based on the protos, by appending the `x-goog-request-params` to whatever metadata the user passes in. But sometimes users may want to customize the `x-goog-request-params` field itself. In bigtable, we need to customize the field based on some values set in the veneer layer

To solve this, I propose we add a check to ensure we are only setting `x-goog-request-params` when it doesn't exist in the passed in metadata, to avoid adding a duplicate field. But let me know if you have other thoughts

Related bigtable bug: https://github.com/googleapis/python-bigtable/issues/1004

daniel-sanche:
Avoid setting the routing metadata field (`x-goog-request-params`) if it already exists in the passed in metadata, to prevent duplicate routing headers

Fixes https://github.com/googleapis/gapic-generator-python/issues/2078


gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
```diff
@@ -368,15 +368,18 @@ class {{ service.async_client_name }}:
...
-            gapic_v1.routing_header.to_grpc_metadata((
-                {% for field_header in method.field_headers %}
-                {% if not method.client_streaming %}
-                ("{{ field_header.raw }}", request.{{ field_header.disambiguated }}),
-                {% endif %}
-                {% endfor %}
-            )),
-        )
+        metadata = tuple(metadata)
+        if all(m[0] != gapic_v1.routing_header.ROUTING_METADATA_KEY for m in metadata):
```
parthea, line 372: Please can you add comments to clarify the reason for the additional logic?

daniel-sanche: I updated the comment above this block to make it more clear, let me know what you think

daniel-sanche:
@parthea I'm a little unsure about the failing showcase tests. Looking at the method names, it seems to line up with the methods in [_test_mixins](https://github.com/googleapis/gapic-generator-python/blob/96870eff62aab87d52934d07d36449bfb3f2df76/gapic/templates/tests/unit/gapic/%25name_%25version/%25sub/_test_mixins.py.j2). But I don't fully understand that file. 

Do I need to duplicate my new tests for each method in _test_mixins by hand? Why is that needed on top of the templates I already added to [test_macros](https://github.com/googleapis/gapic-generator-python/blob/96870eff62aab87d52934d07d36449bfb3f2df76/gapic/templates/tests/unit/gapic/%25name_%25version/%25sub/test_macros.j2)?

daniel-sanche:
Unrelated to this change, a customer reported a separate issue with routing for the async client. We may want to address this at the same time? https://github.com/googleapis/gapic-generator-python/issues/2091

daniel-sanche:
Blocked on https://github.com/googleapis/gapic-generator-python/pull/2133. I will update this after those changes are merged

daniel-sanche:
From https://github.com/googleapis/gapic-generator-python/pull/2133#discussion_r1765185149:

For https://github.com/googleapis/gapic-generator-python/issues/2078 we need to update AIPs first so there is consistent behavior across all language-specific generators: https://google.aip.dev/client-libraries/4222

daniel-sanche:
Closing this because it is out of date after https://github.com/googleapis/gapic-generator-python/pull/2133 (and may no longer be needed)

PR #2098, opened 2024-08-13
feat: Add support for reading ClientLibrarySettings from service configuration YAML

parthea:
Fixes b/356854861

### DIFF
--- a/gapic/schema/api.py
+++ b/gapic/schema/api.py
@@ -623,13 +623,13 @@ class API:
         """
 
         all_errors: dict = {}
-        selectors_seen = []
+        selectors_seen: set = set()
         for method_settings in service_method_settings:
             # Check if this selector is defind more than once
             if method_settings.selector in selectors_seen:
                 all_errors[method_settings.selector] = ["Duplicate selector"]
                 continue
-            selectors_seen.append(method_settings.selector)
+            selectors_seen.add(method_settings.selector)
 
             method_descriptor = self.all_methods.get(method_settings.selector)
             # Check if this selector can be mapped to a method in the API.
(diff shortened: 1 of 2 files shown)


gapic/schema/api.py
```diff
@@ -670,6 +678,63 @@ def enforce_valid_method_settings(
...
+                library settings to be used when generating API methods.
+        Return:
+            None
+        Raises:
+            ClientLibrarySettingsError: if fields in `client_library_settings.experimental_features`
+                are not supported.
+        """
+
+        all_errors: dict = {}
+        versions_seen = []
```
ohmayr, line 727: Can we make this a set?

```
versions_seen = {}
```

gapic/schema/api.py
```diff
@@ -670,6 +678,63 @@ def enforce_valid_method_settings(
...
 
+    @cached_property
+    def all_library_settings(
+        self,
+    ) -> Mapping[str, Sequence[client_pb2.ClientLibrarySettings]]:
+        """Return a map of all `google.api.client.ClientLibrarySettings` to be used
+        when generating client libraries.
+        https://github.com/googleapis/googleapis/blob/master/google/api/client.proto#L130
+
+        Return:
```
ohmayr, line 689: nit: `Returns`.
Same for the other function.

ohmayr (approved):

gapic/schema/api.py
parthea: Fixed in https://github.com/googleapis/gapic-generator-python/pull/2098/commits/e0d252e8cc474efd199bd6d8d1d44ec2806d997c

gapic/schema/api.py
parthea: Fixed in https://github.com/googleapis/gapic-generator-python/pull/2098/commits/d3c4b25845ce97a8d3d8422caac8596018b33dd0

PR #2101, opened 2024-08-21
process: move assertion outside of `pytest.raises` context in showcase tests

### LINKED ISSUE #2100: Update showcase tests to assert conditions outside of the context manager.
The context manager exits immediately as soon as an exception is raised i.e. any following assert statements  or code within the `with` context are not executed. For example, the following test case will pass (even with `assert 2 == 3`):

```python3
def test_dummy():
  with pytest.raises(ValueError) as exc:
      raise ValueError("test")
      assert 2 == 3
      exc.value == "test"  
```

Therefore, showcase tests need to be updated and the assert statements need to be moved out of the `with` context to be executed properly. This also exposes a gap in our tests where some of them are incorrectly passing (since the assertion does not occur).

ohmayr:
Fixes #2100 🦕

### DIFF
--- a/tests/system/test_error_details.py
+++ b/tests/system/test_error_details.py
@@ -57,7 +57,8 @@ def test_bad_request_details(echo):
         )
 
     # Note: gRPC errors expose e.value.details.
-    exc_details = e.details if "rest" in str(echo.transport) else e.value.details
+    exc_details = e.details if "rest" in str(
+        echo.transport) else e.value.details
     assert exc_details == [bad_request_details]
 
 
@@ -88,7 +89,8 @@ def test_precondition_failure_details(echo):
         )
 
     # Note: gRPC errors expose e.value.details.
-    exc_details = e.details if "rest" in str(echo.transport) else e.value.details
+    exc_details = e.details if "rest" in str(
+        echo.transport) else e.value.details
     assert exc_details == [pf_details]
 
 
--- a/tests/system/test_error_details.py
+++ b/tests/system/test_error_details.py
@@ -54,7 +54,10 @@ def test_bad_request_details(echo):
         _ = echo.echo(showcase.EchoRequest(
             error=status,
         ))
-        assert e.details == [bad_request_details]
+
+    # Note: gRPC errors expose e.value.details.
+    exc_details = e.details if "rest" in str(echo.transport) else e.value.details
+    assert exc_details == [bad_request_details]
 
 
 def test_precondition_failure_details(echo):
@@ -80,8 +83,10 @@ def test_precondition_failure_details(echo):
         _ = echo.echo(showcase.EchoRequest(
             error=status,
         ))
-        assert e.details == [pf_details]
 
+    # Note: gRPC errors expose e.value.details.
+    exc_details = e.details if "rest" in str(echo.transport) else e.value.details
+    assert exc_details == [pf_details]
 
 def test_unknown_details(echo):
     status = create_status()
@@ -89,4 +94,3 @@ def test_unknown_details(echo):
         _ = echo.echo(showcase.EchoRequest(
             error=status,
         ))
-        assert e.details == status.details
--- a/tests/system/test_error_details.py
+++ b/tests/system/test_error_details.py
@@ -56,7 +56,7 @@ def test_bad_request_details(echo):
             )
         )
 
-    # Note: error details are exposes e.value.details.
+    # Note: error details are exposed as e.value.details.
     assert e.value.details == [bad_request_details]
 
 
@@ -86,7 +86,7 @@ def test_precondition_failure_details(echo):
             )
         )
 
-    # Note: error details are exposes e.value.details.
+    # Note: error details are exposed as e.value.details.
     assert e.value.details == [pf_details]
 
 
@@ -106,5 +106,5 @@ def test_unknown_details(echo):
             )
         )
 
-    # Note: error details are exposes e.value.details.
+    # Note: error details are exposed as e.value.details.
     assert e.value.details == list(status.details)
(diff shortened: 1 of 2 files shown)


tests/system/test_error_details.py
```diff
@@ -77,16 +82,23 @@ def create_precondition_failure_details():
...
     with pytest.raises(exceptions.GoogleAPICallError) as e:
-        _ = echo.echo(showcase.EchoRequest(
-            error=status,
-        ))
-        assert e.details == status.details
+        _ = echo.echo(
+            showcase.EchoRequest(
+                error=status,
+            )
+        )
```
vchudnov-g, line 104: Is it possible to keep the assertion? 

If it would be possible with non-trivial work, I suggest keeping it, but commenting it out and adding a TODO summarizing what it would take to re-enable it.

tests/system/test_unary.py
```diff
@@ -76,59 +84,79 @@ def test_unary_with_dict(echo):
 
 
 def test_unary_error(echo):
-    message = 'Bad things! Bad things!'
+    message = "Bad things! Bad things!"
+    http_message = (
+        "POST http://localhost:7469/v1beta1/echo:echo: Bad things! Bad things!"
```
vchudnov-g, line 89:
```suggestion
        f"POST http://localhost:7469/v1beta1/echo:echo: {message}"
```

tests/system/test_unary.py
```diff
@@ -76,59 +84,79 @@ def test_unary_with_dict(echo):
...
+            {
+                "error": {
+                    "code": code_pb2.Code.Value("INVALID_ARGUMENT"),
+                    "message": message,
+                },
+            }
+        )
+    err_message = message if "grpc" in str(echo.transport) else http_message
+    assert exc.value.code == 400
+    assert exc.value.message == err_message
```
vchudnov-g, line 110: The TODO I suggest refers to separate work about cleaning error mesaging.

```suggestion
    # TODO: Fix rest errors so that we can use `message` rather than `err_message`
    assert exc.value.message == err_message
```

tests/system/test_error_details.py
```diff
@@ -47,14 +45,21 @@ def create_bad_request_details():
...
+
+    # Note: gRPC errors expose e.value.details.
+    exc_details = e.details if "rest" in str(
+        echo.transport) else e.value.details
+    assert exc_details == [bad_request_details]
```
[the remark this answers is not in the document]
ohmayr: I've removed the variable and used `e.value.details` directly since it's the same for both rest and gRPC. We'll keep the tests disable for rest until the issue mentioned in the TODO is fixed.

tests/system/test_error_details.py
```diff
@@ -47,14 +45,21 @@ def create_bad_request_details():
         field_violation.field = "test field"
         field_violation.description = "test description"
         return bad_request_details
+
```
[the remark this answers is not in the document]
ohmayr: The tests are currently disabled until the issue mentioned in the TODO is resolved.

tests/system/test_error_details.py
```diff
@@ -77,16 +82,23 @@ def create_precondition_failure_details():
...
+        _ = echo.echo(
+            showcase.EchoRequest(
+                error=status,
+            )
+        )
+
+    # Note: gRPC errors expose e.value.details.
+    exc_details = e.details if "rest" in str(
+        echo.transport) else e.value.details
+    assert exc_details == [pf_details]
```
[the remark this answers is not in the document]
ohmayr: addressed.

tests/system/test_error_details.py
```diff
@@ -77,16 +82,23 @@ def create_precondition_failure_details():
...
-        _ = echo.echo(showcase.EchoRequest(
-            error=status,
-        ))
-        assert e.details == [pf_details]
+        _ = echo.echo(
+            showcase.EchoRequest(
+                error=status,
+            )
+        )
+
```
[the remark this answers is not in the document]
ohmayr: See other comment.

tests/system/test_error_details.py
ohmayr: Removed it because we were just comparing empty details but added it back now.

tests/system/test_unary.py
ohmayr: addressed.

tests/system/test_unary.py
ohmayr: Hmm, I don't think this is an issue with the rest error messages. Rather, we've been formatting the rest errors intentionally [here](https://github.com/googleapis/python-api-core/blob/b5ddd6236560f70405c0ea8614d67b27230341c5/google/api_core/exceptions.py#L508).

parthea (approved):

tests/system/test_unary.py
vchudnov-g: Right, that's what I mean by separate work that we'll need to tackle as part of a separate project. But we can defer the TODO for now so we don't confuse the current PR.

tests/system/test_error_details.py
```diff
@@ -47,14 +45,19 @@ def create_bad_request_details():
...
-            error=status,
-        ))
-        assert e.details == [bad_request_details]
+        _ = echo.echo(
+            showcase.EchoRequest(
+                error=status,
+            )
+        )
+
+    # Note: error details are exposes e.value.details.
```
vchudnov-g, line 59:
```suggestion
    # Note: error details are exposed as e.value.details.
```

Also in multiple places in this file

tests/system/test_error_details.py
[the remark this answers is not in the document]
vchudnov-g: ok

vchudnov-g (approved):
Some minor comments, but looks good!

PR #2105, opened 2024-08-26
feat: leverage async anonymous credentials in tests

ohmayr:
asynchronous tests should use `google.auth.aio.credentials.AnonymousCredentials`, if available. For now, we can conditionally make our tests use these credentials until the minimum supported version of `google-auth` is upgraded to one which exposes `google.auth.aio`.

This change will also make it convenient to use the same async tests for async rest transport (once supported) which expects the provided credentials to be of type `google.auth.aio.credentials.Credentials`.

### DIFF
--- a/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
+++ b/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
@@ -93,6 +93,7 @@ def client_cert_source_callback():
     return b"cert bytes", b"key bytes"
 
 # TODO: use async auth anon credentials by default once the minimum version of google-auth is upgraded.
+# See related issue: https://github.com/googleapis/gapic-generator-python/issues/2107.
 def async_anonymous_credentials():
     if HAS_GOOGLE_AUTH_AIO:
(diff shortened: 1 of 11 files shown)


gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -84,6 +92,12 @@ from google.iam.v1 import policy_pb2  # type: ignore
 def client_cert_source_callback():
     return b"cert bytes", b"key bytes"
 
+# TODO: use async auth anon credentials by default once the minimum version of google-auth is upgraded.
```
parthea, line 95: Please can you file a bug to follow up on the todo comment and include the link here?

parthea (approved):

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
ohmayr: Filed https://github.com/googleapis/gapic-generator-python/issues/2107 and added link.

PR #2113, opened 2024-08-28
chore: Update response method macro to support async calls

ohmayr:
This PR updates the response method macro so that it can be re-used in an asynchronous call (similar to what we do for a synchronous call).


parthea:
Marking as draft until presubmits are passing

parthea (approved):

vchudnov-g (approved):

PR #2115, opened 2024-08-29
feat: add cred info to auth related errors

arithmetic1728:
Adds cred info to error details for 401/403/404 errors. See go/python-auth-error-message-improvement.

Manually tested with KMS client: https://github.com/arithmetic1728/google-cloud-python/pull/1


tests/integration/goldens/eventarc/google/cloud/eventarc_v1/services/eventarc/client.py
```diff
@@ -545,6 +547,26 @@ def _validate_universe_domain(self):
             EventarcClient._compare_universes(self.universe_domain, self.transport._credentials))
         return self._is_universe_domain_valid
 
+    def _add_cred_info_for_auth_errors(
```
sai-sunder-s, line 550: why do tests have the actual logic for appending error details?

arithmetic1728: These are golden files. The `tests/integration/goldens/` folder contains the client code generated by the generator for 5 selected apis (redis, eventarc, etc)

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -427,6 +429,26 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+            error (google.api_core.exceptions.GoogleAPICallError): The error to add the cred info.
+        """
+        if error.code not in [HTTPStatus.UNAUTHORIZED, HTTPStatus.FORBIDDEN, HTTPStatus.NOT_FOUND]:
+            return
+
+        cred = self._transport._credentials
+        if not hasattr(cred, "get_cred_info"):
+            return
+
+        cred_info = cred.get_cred_info()  # type: ignore
```
parthea, line 448: Please could you add a comment to clarify the reason that we have `# type: ignore` here, and include a link to a bug to follow up on it?

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -427,6 +429,26 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
             {{ service.client_name }}._compare_universes(self.universe_domain, self.transport._credentials))
         return self._is_universe_domain_valid
 
+    def _add_cred_info_for_auth_errors(
+        self,
+        error: core_exceptions.GoogleAPICallError
+    ) -> None:
+        """Adds credential info string to error details for 401/403/404 errors.
```
parthea, line 436: Is this part of a client library AIP? https://google.aip.dev/client-libraries. If not, should we consider creating one? The AIPs are meant to help define requirements for client libraries which will also be used when we create client libraries for a new language like rust.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -427,6 +429,26 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+
+        Args:
+            error (google.api_core.exceptions.GoogleAPICallError): The error to add the cred info.
+        """
+        if error.code not in [HTTPStatus.UNAUTHORIZED, HTTPStatus.FORBIDDEN, HTTPStatus.NOT_FOUND]:
+            return
+
+        cred = self._transport._credentials
+        if not hasattr(cred, "get_cred_info"):
```
parthea, line 445: Should we emit a warning/ update the error details to let users know that they can get more helpful error messages if they upgrade to a certain version of google-auth?

gapic/templates/%namespace/%name_%version/%sub/services/%service/_mixins.py.j2
```diff
@@ -48,12 +48,16 @@
...
+            response = rpc(
+                request, retry=retry, timeout=timeout, metadata=metadata,)
 
-        # Done; return the response.
-        return response
+            # Done; return the response.
+            return response
+        except core_exceptions.GoogleAPICallError as e:
+            self._add_cred_info_for_auth_errors(e)
+            raise e
```
parthea, line 60: Non-blocking comment. This PR touches a lot of duplicate code. Can we refactor this into a macro to avoid making the change in so many places? Feel free to file a bug to follow up on it later. If we address it now, then subsequent updates to the same code will be less toilsome.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
```diff
@@ -252,6 +259,44 @@ def test__get_universe_domain():
...
+    cred.get_cred_info = mock.Mock(return_value=cred_info_json)
+    client = {{ service.client_name }}(credentials=cred)
+    client._transport._credentials = cred
+
+    error = core_exceptions.GoogleAPICallError("message", details=[])
+    error.code = error_code
+
+    client._add_cred_info_for_auth_errors(error)
+    if show_cred_info:
+        assert error.details == [CRED_INFO_STRING]
```
parthea, line 283: Can we also add a test to ensure that error details that we receive from the API are not clobbered? 

As an example, we could update the `test_error_details` showcase test which already has tests with error details 
https://github.com/googleapis/gapic-generator-python/blob/main/tests/system/test_error_details.py

parthea (commented):
Added minor observations

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
sai-sunder-s: Yes that makes sense

sai-sunder-s: Actually on second thought, we cannot do that since its not necessary that all credentials in all scenarios have cred info.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
arithmetic1728: this is because old version of google-auth doesn't have this field, which causes pypy test to fail. I added the comment.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
arithmetic1728: This is not part of AIP. I think this is flexible, each language has their own way of providing the info, some lang chooses to use subclass instead of this error info string, so I guess we don't create an AIP for this.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
arithmetic1728: that's right, it's not necessary that all credentials have cred info in all scenarios, so we don't need a warning here.

gapic/templates/%namespace/%name_%version/%sub/services/%service/_mixins.py.j2
arithmetic1728: I will create a bug linking the code after the PR is merged.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_%service.py.j2
arithmetic1728: I modified the unit test by adding error details to the exception: `error = core_exceptions.GoogleAPICallError("message", details=["foo"])` and tested that `error.details==["foo",CRED_INFO_STRING]`.

parthea (approved):

PR #2130, opened 2024-09-06
feat: leverage auth session and implement close method

ohmayr:
This PR leverages `google.auth.aio` in async rest transport and implements a close method for the transport.

### DIFF
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
@@ -573,9 +573,7 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
             )
 
             if "rest_asyncio" in str(transport_init):
-                # Note: The following parameters are not support in async rest:
-                # credentials_file, scopes, client_cert_source_for_mtls, quota_project_id
-                # always_use_jwt_access, api_audience.
+                {# TODO(https://github.com/googleapis/gapic-generator-python/issues/2136): Support the following parameters in async rest: #}
                 unsupported_params = {
                     "google.api_core.client_options.ClientOptions.credentials_file": self._client_options.credentials_file,
                     "google.api_core.client_options.ClientOptions.scopes": self._client_options.scopes,
@@ -586,7 +584,7 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
                 }
                 provided_unsupported_params = [name for name, value in unsupported_params.items() if value is not None]
                 if provided_unsupported_params:
-                    raise NotImplementedError(
+                    raise core_exceptions.AsyncRestUnsupportedParameterError(
                         f"The following provided parameters are not supported for `transport=rest_asyncio`: {', '.join(provided_unsupported_params)}"
                     )
                 self._transport = transport_init(
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
@@ -584,7 +584,7 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
                 }
                 provided_unsupported_params = [name for name, value in unsupported_params.items() if value is not None]
                 if provided_unsupported_params:
-                    raise core_exceptions.AsyncRestUnsupportedParameterError(
+                    raise core_exceptions.AsyncRestUnsupportedParameterError(  # type: ignore
                         f"The following provided parameters are not supported for `transport=rest_asyncio`: {', '.join(provided_unsupported_params)}"
                     )
                 self._transport = transport_init(
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
@@ -584,6 +584,7 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
                 }
                 provided_unsupported_params = [name for name, value in unsupported_params.items() if value is not None]
                 if provided_unsupported_params:
+                    {# TODO(https://github.com/googleapis/gapic-generator-python/issues/2137): Remove `type: ignore` once we add a version check for google-api-core. #}
                     raise core_exceptions.AsyncRestUnsupportedParameterError(  # type: ignore
                         f"The following provided parameters are not supported for `transport=rest_asyncio`: {', '.join(provided_unsupported_params)}"
                     )
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
@@ -66,7 +66,13 @@ from .transports.grpc_asyncio import {{ service.grpc_asyncio_transport_name }}
 from .transports.rest import {{ service.name }}RestTransport
 {# TODO(https://github.com/googleapis/gapic-generator-python/issues/2121): Remove this condition when async rest is GA. #}
 {% if rest_async_io_enabled %}
-from .transports.rest_asyncio import Async{{ service.name }}RestTransport
+try:
+    from .transports.rest_asyncio import Async{{ service.name }}RestTransport
+    HAS_GOOGLE_AUTH_AIO = True
+except ImportError as e: # pragma: NO COVER
+    HAS_GOOGLE_AUTH_AIO = False
+    GOOGLE_AUTH_AIO_EXCEPTION = e
+
 {% endif %}{# if rest_async_io_enabled #}
 {% endif %}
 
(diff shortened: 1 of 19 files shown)


gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -552,16 +564,52 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+            transport_init: Union[Type[{{ service.name  }}Transport], Callable[..., {{ service.name }}Transport]] = (
+                {{ service.client_name  }}.get_transport_class(transport)
+                if isinstance(transport, str) or transport is None
+                else cast(Callable[..., {{ service.name }}Transport], transport)
+            )
+
+            if "rest_asyncio" in str(transport_init):
+                # Note: The following parameters are not support in async rest:
+                # credentials_file, scopes, client_cert_source_for_mtls, quota_project_id
+                # always_use_jwt_access, api_audience.
```
parthea, line 578:
```suggestion
                # TODO(<bug>): The following parameters are not supported in async rest:
                # `credentials_file, scopes`, `client_cert_source_for_mtls`, `quota_project_id`
                # `always_use_jwt_access`, `api_audience`.
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_base.py.j2
```diff
@@ -56,7 +55,7 @@ class _Base{{ service.name }}RestTransport({{service.name}}Transport):
     {# TODO: handle mtls stuff if that is relevant for rest transport #}
     def __init__(self, *,
             host: str{% if service.host %} = '{{ service.host }}'{% endif %},
-            credentials: Optional[ga_credentials.Credentials] = None,
+            credentials: Optional[Any] = None,
```
parthea, line 58: Please a comment in the code to clarify the reason that we have `Any` instead of `ga_credentials.Credentials`

parthea (approved):

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
vchudnov-g: There's no need to list the parameters in the comment when they're listed very clearly in the code just below.

```suggestion
                # TODO(<bug>): Support the following parameters in async rest:
```

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
```diff
@@ -1892,9 +1892,61 @@ def test_{{ method_name }}_empty_call():
...
+    options = client_options.ClientOptions(quota_project_id="octopus")
+    with pytest.raises(NotImplementedError) as exc:
+        client = {{ get_client(service, True) }}(
+            credentials={{get_credentials(True)}},
+            transport="rest_asyncio",
+            client_options=options
+        )
+    exc.match(
+        "The following provided parameters are not supported for `transport=rest_asyncio`: google.api_core.client_options.ClientOptions.quota_project_id"
+    )
```
vchudnov-g, line 1950: You could define and raise this in the appropriate module:
```
class AsyncRestUnsupportedParameterError(NotImplementedError):
    pass
```

and then test here a bit more robustly/succinctly:

```suggestion
    with pytest.raises(AsyncRestUnsupportedParameterError, match="google.api_core.client_options.ClientOptions.quota_project_id") :
        client = {{ get_client(service, True) }}(
            credentials={{get_credentials(True)}},
            transport="rest_asyncio",
            client_options=options
        )
    
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -552,16 +564,52 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+                return
+
+            {% endif %}
             import google.auth._default  # type: ignore
 
             if api_key_value and hasattr(google.auth._default, "get_api_key_credentials"):
                 credentials = google.auth._default.get_api_key_credentials(api_key_value)
 
+            {# TODO(https://github.com/googleapis/gapic-generator-python/issues/2121): Remove this condition when async rest is GA. #}
+            {% if not rest_async_io_enabled %}
```
vchudnov-g, line 606: Why the `not` here? The init was working for non-sync-rest before, right?You'll have a different constructor for AsyncRest? If so, I would suggest tweaking to make this explicit. For example, something like:

```suggestion
           {% if rest_async_io_enabled %}
              raise NotImplementedError("transport_init for async REST not implemented yet")
           {% else %}
            transport_init: ...
            {% endif %}
```

(Or clarify if I interpreted this wrong.)

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_base.py.j2
vchudnov-g: +1. If we could have a stronger type here, that would be great. Same for similar places below.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
```diff
@@ -1892,9 +1892,61 @@ def test_{{ method_name }}_empty_call():
...
+{% set async_prefix = "async " if is_async else "" %}
+{% set async_decorator = "@pytest.mark.asyncio " if is_async else "" %}
+{% set transport_name = get_transport_name(transport, is_async) %}
+{% set close_session = {
+    'rest': "_session",
+    'grpc': "_grpc_channel"}
+-%}
+{{async_decorator}}
+{{async_prefix}}def test_transport_close_{{transport_name}}():
+
```
vchudnov-g, line 1918:
```suggestion
```

vchudnov-g (approved):
One non-trivial comment on [client.py.j2](https://github.com/googleapis/gapic-generator-python/pull/2130/files/a2d897a53d09facf9bbf8e2ae49d683ef5963cc4#diff-c71dde17175c66932820c2ad6ba1347a2df51eccb8d39b9060abb007bb0e5f63), but otherwise LGTM. TAL at my other minor suggestions.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
ohmayr: `transport_init` is already defined  here`https://github.com/googleapis/gapic-generator-python/pull/2130/files/a2d897a53d09facf9bbf8e2ae49d683ef5963cc4#diff-c71dde17175c66932820c2ad6ba1347a2df51eccb8d39b9060abb007bb0e5f63R569` when async rest is enabled so we don't need to re-initialize it here.

I could just generate `transport_init` unconditionally like it was before but I needed it above to do some checks. And not doing it the way i'm doing would regenerate every client. I was gonna do it as a follow-up so that I could keep the diffs for the generated clients smaller.

Let me know if this makes sense.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
ohmayr: For [this comment](https://github.com/googleapis/gapic-generator-python/pull/2130#discussion_r1750757100), I don't want to have a dependency on google.auth in the base client.

1- Given that we support versions of `google-auth` which don't have all the features like `google.auth.aio` as an example, we'll have to do some hacks to make it work and that doesn't look clean.
2- The base rest transport class isn't meant to be used directly so skipping type hints for it shouldn't affect users.

If you have some other suggestion then I can incorporate it. Let me know.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
vchudnov-g: Ah, OK. Makes sense.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
vchudnov-g: OK. We shouldn't get stuck on this, so I'm fine proceeding. But the concrete classes can have these tighter type hints?

It seems to me the most elegant thing to do is seeing how the we use the parameter `Credentials`; define an ABC with that interface; have the concrete credential classes derive from that ABC, and use that ABC as the type hint here. That would make the clear and self-documenting, especially for people seeking to write new derivations of that ABC. I would support following up on this, but this is not a high priority so let's not block this PR on that. Do add a comment on the code explaining the `Any` type hint, as @parthea suggested.

ohmayr: addressed by adding a note.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
ohmayr: addressed here: https://github.com/googleapis/python-api-core/pull/694

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_base.py.j2
ohmayr: addressed.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
ohmayr: addressed.

Issue #2091, opened 2024-08-06
async client does not support explicit routing


daniel-sanche:
A python-datastore user [recently reported](https://github.com/googleapis/python-datastore/issues/544#event-13103173857) that the async client is missing some routing headers added to the sync client, making it unusable

Sync ([link](https://github.com/googleapis/python-datastore/blob/5e773cb8c766303fef53965dd100b3c4c93b98be/google/cloud/datastore_v1/services/datastore/client.py#L765C1-L776C14)):
```
header_params = {}

if request.project_id:
    header_params["project_id"] = request.project_id

if request.database_id:
    header_params["database_id"] = request.database_id

if header_params:
    metadata = tuple(metadata) + (
        gapic_v1.routing_header.to_grpc_metadata(header_params),
    )
```

Async ([link](https://github.com/googleapis/python-datastore/blob/5e773cb8c766303fef53965dd100b3c4c93b98be/google/cloud/datastore_v1/services/datastore/async_client.py#L368)):
```
metadata = tuple(metadata) + (
    gapic_v1.routing_header.to_grpc_metadata(
        (("project_id", request.project_id),)
    ),
)
```

daniel-sanche:
It seems like the async client doesn't support [explicit_routing](https://github.com/googleapis/gapic-generator-python/blob/b64e7a3351804613aaf56d7a699390d8dbc70712/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/_client_macros.j2#L147)?

daniel-sanche:
Related issue: https://github.com/googleapis/gapic-generator-python/issues/1757 discusses the performance cost of the regex verification. But also addresses that the generated output is different

adrianwbsf:
Hi @daniel-sanche, do you have any updates on the issue? Will this issue make selecting the database possible?

daniel-sanche:
Hey, no concrete updates yet, but it's being looked at!

PR #2131, opened 2024-09-09
test: improve routing parameter assert

parthea:
This PR fixes the assert in `test_*_routing_parameters` to verify the routing headers in `kw['metadata']`. 

The `resolve` function in `tests/unit/schema/wrappers/test_routing.py` was moved to `gapic/schema/wrappers.py` so that it can be used in `gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2`

```
    # This test doesn't assert anything useful.
    assert kw['metadata']
```

Towards https://github.com/googleapis/gapic-generator-python/issues/2091

### DIFF
--- a/gapic/schema/wrappers.py
+++ b/gapic/schema/wrappers.py
@@ -55,12 +55,6 @@ from gapic.schema import metadata
 from gapic.utils import uri_sample
 
 
-# This is the value for the `maxsize` argument of @functools.lru_cache
-# https://docs.python.org/3/library/functools.html#functools.lru_cache
-# This represents the number of recent function calls to store.
-ROUTING_PARAM_REGEX_CACHE_SIZE = 32
-
-
 @dataclasses.dataclass(frozen=True)
 class Field:
     """Description of a field."""
@@ -1043,7 +1037,8 @@ class RoutingParameter:
         return re.compile(f"^{self._convert_to_regex(path_template)}$")
 
     # Use caching to avoid repeated computation
-    @functools.lru_cache(maxsize=ROUTING_PARAM_REGEX_CACHE_SIZE)
+    # https://docs.python.org/3/library/functools.html#functools.cache
+    @functools.cache
     def to_regex(self) -> Pattern:
         return self._to_regex(self.path_template)
 
--- a/gapic/schema/wrappers.py
+++ b/gapic/schema/wrappers.py
@@ -1037,15 +1037,19 @@ class RoutingParameter:
         return re.compile(f"^{self._convert_to_regex(path_template)}$")
 
     # Use caching to avoid repeated computation
+    # TODO(https://github.com/googleapis/gapic-generator-python/issues/2161):
+    # Use `@functools.cache` instead of `@functools.lru_cache` once python 3.8 is dropped.
     # https://docs.python.org/3/library/functools.html#functools.cache
-    @functools.cache
+    @functools.lru_cache(max_size=None)
     def to_regex(self) -> Pattern:
         return self._to_regex(self.path_template)
 
     @property
     # Use caching to avoid repeated computation
+    # TODO(https://github.com/googleapis/gapic-generator-python/issues/2161):
+    # Use `@functools.cache` instead of `@functools.lru_cache` once python 3.8 is dropped.
     # https://docs.python.org/3/library/functools.html#functools.cache
-    @functools.cache
+    @functools.lru_cache(max_size=None)
     def key(self) -> Union[str, None]:
         if self.path_template == "":
             return self.field
--- a/gapic/schema/wrappers.py
+++ b/gapic/schema/wrappers.py
@@ -30,6 +30,7 @@ Documentation is consistently at ``{thing}.meta.doc``.
 import collections
 import copy
 import dataclasses
+import functools
 import json
 import keyword
 import re
@@ -52,6 +53,12 @@ from google.protobuf.json_format import MessageToDict  # type: ignore
 from gapic import utils
 from gapic.schema import metadata
 from gapic.utils import uri_sample
+
+
+# This is the value for the `maxsize` argument of @functools.lru_cache
+# https://docs.python.org/3/library/functools.html#functools.lru_cache
+# This represents the number of recent function calls to store.
+ROUTING_PARAM_REGEX_CACHE_SIZE = 32
 
 
 @dataclasses.dataclass(frozen=True)
@@ -1035,6 +1042,8 @@ class RoutingParameter:
         """
         return re.compile(f"^{self._convert_to_regex(path_template)}$")
 
+    # Use caching to avoid repeated computation
+    @functools.lru_cache(maxsize=ROUTING_PARAM_REGEX_CACHE_SIZE)
     def to_regex(self) -> Pattern:
         return self._to_regex(self.path_template)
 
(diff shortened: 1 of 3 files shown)


gapic/schema/wrappers.py
```diff
@@ -1067,6 +1067,66 @@ def try_parse_routing_rule(cls, routing_rule: routing_pb2.RoutingRule) -> Option
...
+        header_params = {}
+        for routing_param in routing_rule.routing_parameters:
+            request_field_value = _get_field(request, routing_param.field)
+            # Only resolve the header for routing parameter fields which are populated in the request
+            if request_field_value is not None:
+                # If there is a path_template for a given routing parameter field, the value of the field must match
+                # If multiple Routing Parameters describe the same key
+                # (via the `path_template` field or via the `field` field when
+                # `path_template` is not provided), "last one wins" rule
+                # determines which Parameter gets used.
```
vchudnov-g, line 1116:
```suggestion
                # If there is a path_template for a given routing parameter field, the value of the field must match
                # If multiple `routing_param`s describe the same key
                # (via the `path_template` field or via the `field` field when
                # `path_template` is not provided), the "last one wins" rule
                # determines which parameter gets used. See https://google.aip.dev/client-libraries/4222.
```

gapic/schema/wrappers.py
```diff
@@ -1067,6 +1067,66 @@ def try_parse_routing_rule(cls, routing_rule: routing_pb2.RoutingRule) -> Option
...
+                if routing_param.path_template:
+                    routing_param_regex = routing_param.to_regex()
+                    regex_match = routing_param_regex.match(
+                        request_field_value
+                    )
+                    if regex_match:
+                        header_params[routing_param.key] = regex_match.group(
```
vchudnov-g, line 1123: Every call to `routing_param.key` with a non-trivial `path_template` calls `to_regex()`. We should memoize `to_regex()`, and then either also memoize `key` or set it in this method outside of the `if routing_param.path_template`

gapic/schema/wrappers.py
```diff
@@ -1067,6 +1067,66 @@ def try_parse_routing_rule(cls, routing_rule: routing_pb2.RoutingRule) -> Option
         params = [RoutingParameter(x.field, x.path_template) for x in params]
         return cls(params)
 
+    @classmethod
+    def resolve(cls, routing_rule: routing_pb2.RoutingRule, request: Union[dict, str]) -> dict:
+        """Resolves the routing header which should be sent along with the request.
+        This function performs dynamic header resolution, identical to what's in `client.py.j2`.
```
vchudnov-g, line 1073: Where, exactly, is it identical? It seems like if it is identical it should move to a shared library, like api_core.

Also, I'm not clear about how `wrappers.py` gets propagated to the templates. Does this comment mean it isn't?

vchudnov-g (commented):
Looks good, but I have some questions to make sure I understand the (pre-existing) structure of the code touched here.

gapic/schema/wrappers.py
parthea: Fixed in https://github.com/googleapis/gapic-generator-python/pull/2131/commits/f0ffb4c8f7e0a0adbaeb24c2f94c8807c7b82698

gapic/schema/wrappers.py
parthea: > Where, exactly, is it identical? 

It is actually in `_client_macros.j2` not `client.py.j2`. The code was copied from the test below

https://github.com/googleapis/gapic-generator-python/blob/4c5de8791795f8101f6ec66f80b8a8e5e9a21822/tests/unit/schema/wrappers/test_routing.py#L27

I've updated the comment and included a link to the specific code in `_client_macros.j2` which is here

https://github.com/googleapis/gapic-generator-python/blob/4c5de8791795f8101f6ec66f80b8a8e5e9a21822/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/_client_macros.j2#L150-L164


> It seems like if it is identical it should move to a shared library, like api_core.
Good point. I've filed https://github.com/googleapis/gapic-generator-python/issues/2160. I didn't make the change in this PR as this is a refactor of `_client_macros.j2`  which is not the purpose of this PR. This PR is improving the assert and moving logic which already exists in `tests/unit/schema/wrappers/test_routing.py` to `gapic/schema/wrappers.py` so that the same code that we use in gapic-generator tests can also be used in generated libraries.

> Also, I'm not clear about how wrappers.py gets propagated to the templates. Does this comment mean it isn't?

We can call the helper functions in wrappers.py from templates but we cannot use it to generate client code. For example, see the code below where we call the helper function `method.routing_rule.resolve`.

https://github.com/googleapis/gapic-generator-python/blob/e50bbd6921ba3dc112c5c625c5f16fc7d7f6f57c/gapic/templates/tests/unit/gapic/%25name_%25version/%25sub/test_macros.j2#L428

https://github.com/googleapis/gapic-generator-python/issues/2160 is for refactoring  `_client_macros.j2` to use helper functions to build the explicit routing headers in `google-api-core` rather than having the explicit routing header logic in templates.

gapic/schema/wrappers.py
```diff
@@ -1067,6 +1078,68 @@ def try_parse_routing_rule(cls, routing_rule: routing_pb2.RoutingRule) -> Option
         params = [RoutingParameter(x.field, x.path_template) for x in params]
         return cls(params)
 
+    @classmethod
+    def resolve(cls, routing_rule: routing_pb2.RoutingRule, request: Union[dict, str]) -> dict:
+        """Resolves the routing header which should be sent along with the request.
+        This function performs dynamic header resolution, identical to what's in `_client_macros.j2`.
+        https://github.com/googleapis/gapic-generator-python/blob/4c5de8791795f8101f6ec66f80b8a8e5e9a21822/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/_client_macros.j2#L150-L164
```
vchudnov-g, line 1085: nit: Consider using a TinyURL or similar, and enclosing the URL in `<>` so that it's rendered clickable by some tools.

gapic/schema/wrappers.py
```diff
@@ -1067,6 +1078,68 @@ def try_parse_routing_rule(cls, routing_rule: routing_pb2.RoutingRule) -> Option
         params = [RoutingParameter(x.field, x.path_template) for x in params]
         return cls(params)
 
+    @classmethod
+    def resolve(cls, routing_rule: routing_pb2.RoutingRule, request: Union[dict, str]) -> dict:
+        """Resolves the routing header which should be sent along with the request.
+        This function performs dynamic header resolution, identical to what's in `_client_macros.j2`.
```
vchudnov-g, line 1084: nit: I would add a TODO referencing #2160 about refactoring the common code and then removing this sentence.

gapic/schema/wrappers.py
vchudnov-g: Ah, thanks!

vchudnov-g (approved):

gapic/schema/wrappers.py
parthea: Superseded by review comment https://github.com/googleapis/gapic-generator-python/pull/2131#discussion_r1761945891

gapic/schema/wrappers.py
parthea: Fixed in https://github.com/googleapis/gapic-generator-python/pull/2131/commits/4bbfd9f5b3918b9d2da1c43f4c53f1fccadf7995

PR #2139, opened 2024-09-10
feat: wrap async rest methods with retry and error logic

ohmayr:
This PR adds error handling and retry logic to async rest methods. Note that the `callables` will be implemented in a follow up PR.

This PR should be reviewed and merged after: https://github.com/googleapis/gapic-generator-python/pull/2138.

PR #2140, opened 2024-09-11
feat: add async rest transport call methods

ohmayr:
This PR adds asynchronous rest transport call methods which when called raise a `NotImplementedError`. Follow up PRs will implement this method class.

As of this PR, asynchronous rest transport does not support calling any type of methods.

This PR should be reviewed and merged after: https://github.com/googleapis/gapic-generator-python/pull/2139.

### DIFF
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_asyncio.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_asyncio.py.j2
@@ -6,7 +6,7 @@
 {% block content %}
 
 try:
-    from google.auth.aio.transport.sessions import AsyncAuthorizedSession  # type: ignore
+    from google.auth.aio.transport.sessions import AsyncAuthorizedSession # type: ignore
 except ImportError as e:  # pragma: NO COVER
     {# TODO(https://github.com/googleapis/google-auth-library-python/pull/1577): Update the version of google-auth once the linked PR is merged. #}
     raise ImportError("async rest transport requires google.auth >= 2.x.x") from e
@@ -17,6 +17,7 @@ from google.api_core import exceptions as core_exceptions
 from google.api_core import gapic_v1
 from google.api_core import retry_async as retries
 
+import dataclasses
 from typing import Any, Callable, Tuple, Optional, Sequence, Union
 
 {{ shared_macros.operations_mixin_imports(api, service, opts) }}
@@ -25,12 +26,23 @@ from .rest_base import _Base{{ service.name }}RestTransport
 
 from .base import DEFAULT_CLIENT_INFO as BASE_DEFAULT_CLIENT_INFO
 
+try:
+    OptionalRetry = Union[retries.AsyncRetry, gapic_v1.method._MethodDefault, None]
+except AttributeError:  # pragma: NO COVER
+    OptionalRetry = Union[retries.AsyncRetry, object, None]  # type: ignore
+
 {# TODO (https://github.com/googleapis/gapic-generator-python/issues/2128): Update `rest_version` to include the transport dependency version. #}
 DEFAULT_CLIENT_INFO = gapic_v1.client_info.ClientInfo(
     gapic_version=BASE_DEFAULT_CLIENT_INFO.gapic_version,
     grpc_version=None,
     rest_version=None,
 )
+
+{# TODO: Add an `_interceptor` property once implemented #}
+@dataclasses.dataclass
+class Async{{service.name}}RestStub:
+    _session: AsyncAuthorizedSession
+    _host: str
 
 class Async{{service.name}}RestTransport(_Base{{ service.name }}RestTransport):
     """Asynchronous REST backend transport for {{ service.name }}.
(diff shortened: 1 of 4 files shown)


gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_asyncio.py.j2
```diff
@@ -26,13 +26,24 @@ from .rest_base import _Base{{ service.name }}RestTransport
...
+    OptionalRetry = Union[retries.AsyncRetry, object, None]  # type: ignore
+
 {# TODO (https://github.com/googleapis/gapic-generator-python/issues/2128): Update `rest_version` to include the transport dependency version. #}
 DEFAULT_CLIENT_INFO = gapic_v1.client_info.ClientInfo(
     gapic_version=BASE_DEFAULT_CLIENT_INFO.gapic_version,
     grpc_version=None,
     rest_version=None,
 )
 
+{# TODO: Add an `_interceptor` property once implemented #}
```
parthea, line 41: Please include the TODO bug to follow up

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_asyncio.py.j2
```diff
@@ -93,14 +104,29 @@ class Async{{service.name}}RestTransport(_Base{{ service.name }}RestTransport):
...
-    {# TODO(b/362949446) Return a callable once implemented and remove `type:  ignore`. #}
+    {% endfor %}
+    {% for method in service.methods.values()|sort(attribute="name") %}
     @property
     def {{method.transport_safe_name|snake_case}}(self) -> Callable[
             [{{method.input.ident}}],
             {{method.output.ident}}]:
         
-        return  # type: ignore
+        return self._{{method.name}}(self._session, self._host)  # type: ignore
```
parthea, line 129: Do we still need the `# type: ignore` ?

parthea (approved):
Added non-blocking comments. Feel free to follow up in a later PR

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_asyncio.py.j2
ohmayr: I think it's fine to ignore this one since this comment is addressed and removed in a follow up PR: https://github.com/googleapis/gapic-generator-python/pull/2151.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_asyncio.py.j2
ohmayr: Yes. My initial thought was that it's probably a mismatch since I wasn't returning a Callable. However, I later realised that we're doing the same in `rest.py`.

I haven't really looked into why we've been ignoring the type in `rest.py` but I'm following whatever we're doing there. Probably because the Callables for client streaming which raise `NotImplmentedError` do not return anything so mypy isn't happy about it? Regardless, I've filed an issue to investigate and remove this as a follow up.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_asyncio.py.j2
```diff
@@ -93,14 +104,29 @@ class Async{{service.name}}RestTransport(_Base{{ service.name }}RestTransport):
     {{ shared_macros.wrap_async_method_macro()|indent(4) }}
 
     {% for method in service.methods.values()|sort(attribute="name") %}
+    class _{{method.name}}(_Base{{ service.name }}RestTransport._Base{{method.name}}, Async{{service.name}}RestStub):
+        def __hash__(self):
+            return hash("Async{{service.name}}RestTransport.{{method.name}}")
+
+        async def __call__(self,
+                    request: {{method.input.ident}}, *,
```
vchudnov-g, line 112: What does the asterisk mean here?

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_asyncio.py.j2
vchudnov-g, line 129: So each of these properties is a callable that invokes the appropriate inner class. When do the `self._wrapped_methods` get called? (And which of your PR will ensure that?)

vchudnov-g (commented):
Looking good, but I'll have another look after this is rebased on top of the previous PRs.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_asyncio.py.j2
ohmayr: arguments following `*` must be passed in as keyword arguments.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_asyncio.py.j2
ohmayr: wrapping happens during transport initialization / construction via `self._prep_wrapped_messages(client_info)`. If you look at the implementation for this helper, it creates a dictionary called `_wrapped_methods` which is something like `{method : wrapped_method, ... }`.

At the client level instead of calling `method()`, we call `self._wrapped_methods[method]()`. However, we're not doing this for the following methods:

- `set_iam_policy`
- `get_iam_policy`
- `test_iam_permissions`

FYI these are mixin methods. We don't pre-wrap mixin methods which is why this logic isn't added. We probably should do that or at least pass down the kind property to avoid gRPC code path in a rest call at the very least.

vchudnov-g (approved):
LGTM, but PTAL @ my previous comments before merging.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_asyncio.py.j2
ohmayr: I think this is important. I've filed https://github.com/googleapis/gapic-generator-python/issues/2156 to ensure that we do this as a follow up since this PR doesn't touch the client layer.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_asyncio.py.j2
vchudnov-g: Oh, it's a Python thing. Because it was on the same line, I assumed it was part of the type hint and glossed over the comma.

I suggest having the `*` in its own line, like the other parameters. It gets lost here.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_asyncio.py.j2
vchudnov-g: > At the client level instead of calling `method()`, we call `self._wrapped_methods[method]()`

OK, this is what I was looking for.

Once we have `self._wrapped_methods`, do we ever call a `method()` directly? If we don't, then the most elegant thing would arguably be to overwrite the properties with their wrapped versions, and just called each (now-wrapped) `method()` directly. Does that make sense? If so, it would be a good code simplification to do later.

PR #2141, opened 2024-09-11
chore: refactor rest transport bad request test macro

ohmayr:
This PR refactors the macro for testing a rest method call which raises a `google.api_core.exceptions.BadRequest` error.

For now, the macro is guarded and only generates tests for sync REST methods. Once we implement the relevant logic in async rest, the guard can be removed. (This will be addressed in a follow up PR.)

This PR should be reviewed and merged after: https://github.com/googleapis/gapic-generator-python/pull/2140.

### DIFF
--- a/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
+++ b/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
@@ -2013,6 +2013,7 @@ def test_initialize_client_w_{{transport_name}}():
 
 {# rest_method_bad_request_test generates tests for rest methods
  # which raise a google.api.core.exceptions.BadRequest error.
+ # TODO(https://github.com/googleapis/gapic-generator-python/issues/2157): Refactor this macro to include `gRPC` coverage.
 #}
 {% macro rest_method_bad_request_test(service, method, transport, is_async) %}
 {% set await_prefix = get_await_prefix(is_async) %}
--- a/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
+++ b/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
@@ -1948,15 +1948,13 @@ def test_unsupported_parameter_rest_asyncio():
 # into `run_transport_tests_for_config` and make any of the rest specific specific macros which are called within more generic. 
 #}
 {% macro run_transport_tests_for_config(service, transport, is_async) %}
-{% if 'rest' in transport %}
 {% for method in service.methods.values() %}
 {% if is_rest_unsupported_method(method, is_async) == 'True' or not method.http_options %}
 {{ rest_method_not_implemented_error(service, method, transport, is_async) }}
 {% else %}
-{{ rest_method_bad_request_test(service, method, transport, is_async) }}
+{{ bad_request_test(service, method, transport, is_async) }}
 {% endif %}{# is_rest_unsupported_method(method, is_async) == 'False' and method.http_options #}
 {% endfor %}
-{% endif %}{# if 'rest' in transport #}
 {{ initialize_client_with_transport_test(service, transport, is_async) }}
 {% endmacro %}
 
@@ -1964,6 +1962,7 @@ def test_unsupported_parameter_rest_asyncio():
  # which are not supported for rest transport.
 #}
 {% macro rest_method_not_implemented_error(service, method, transport, is_async) %}
+{% if 'rest' in transport %}
 {% set await_prefix = get_await_prefix(is_async) %}
 {% set async_prefix = get_async_prefix(is_async) %}
 {% set async_decorator = get_async_decorator(is_async) %}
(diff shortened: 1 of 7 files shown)


gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
```diff
@@ -2015,3 +1990,41 @@ def test_{{transport_name}}_initialize_client():
     )
 
 {% endmacro %}
+
+{# rest_method_bad_request_test generates tests for rest methods
+ # which raise a google.api.core.exceptions.BadRequest error.
+#}
+{% macro rest_method_bad_request_test(service, method, method_name, transport_name, async_prefix, await_prefix, async_decorator, is_async) %}
```
parthea, line 1997: Should we add a TODO bug to follow up on making this test more generic to cover gRPC as well? For example, we could simulate a 400 response using either gRPC or REST

parthea (approved):
Added non-blocking comment. Feel free to follow up in a later PR

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
ohmayr: addressed.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
```diff
@@ -2035,3 +2010,47 @@ def test_initialize_client_w_{{transport_name}}():
     assert client is not None
 
 {% endmacro %}
+
+{# rest_method_bad_request_test generates tests for rest methods
+ # which raise a google.api.core.exceptions.BadRequest error.
+ # TODO(https://github.com/googleapis/gapic-generator-python/issues/2157): Refactor this macro to include `gRPC` coverage.
+#}
+{% macro rest_method_bad_request_test(service, method, transport, is_async) %}
```
vchudnov-g, line 2018: Since you're passing the transport in, and since you'll want to have gRPC in here:
```suggestion
{% macro bad_request_test(service, method, transport, is_async) %}
```

vchudnov-g, line 2018: non-blocking, but desirable: Same comment as in the other PR: I'd have  a macro-level fail-safe check that `"rest" in transport` that will cause a generation-time exception if we call the macro for non-REST. We can adapt/remove that as we factor gRPC functionality in here.

tests/integration/goldens/redis/tests/unit/gapic/redis_v1/test_cloud_redis.py
```diff
@@ -8371,6 +8151,237 @@ def test_transport_kind_rest():
     assert transport.kind == "rest"
 
 
+def test_list_instances_rest_bad_request(request_type=cloud_redis.ListInstancesRequest):
```
vchudnov-g, line 8154: Do any of the goldens test the `await_prefix`, `async_prefix`, etc?

vchudnov-g (approved):

tests/integration/goldens/redis/tests/unit/gapic/redis_v1/test_cloud_redis.py
ohmayr: Not here but in one of the follow up PRs where we turn on this test for async.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
```diff
@@ -1989,6 +1988,7 @@ def test_unsupported_parameter_rest_asyncio():
         in str(not_implemented_error.value)
     )
 
+{% endif %}{# if 'rest' in transport #}
```
vchudnov-g, line 1991: Have an `else` to raise a generation time error.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
```diff
@@ -2053,4 +2054,5 @@ def test_initialize_client_w_{{transport_name}}():
         req.return_value = response_value
         {{ await_prefix }}client.{{ method_name }}(request)
 
+{% endif %}{# if 'rest' in transport #}
```
vchudnov-g, line 2057: Have an `else` to raise a generation time error. It's even more important here because we don't have `rest` in the name (as we shouldn't, since we pass `transport`). 

Otherwise, if somebody else comes along and see the call to this macro, unless they look inside they could assume it will work for gRPC, and delete the existing gRPC-testing functionality where it is now in favor of calling this macro....and things will "work" because all that would do is remove the test code, not move them around.

Having the failsafe is a second line of defense. I strongly encourage it. I know I harp on this a lot, but I've found it useful to assume that future developers, including myself, will overlook something that is obvious now.

ohmayr: As discussed online, we'll be raising a runtime error within the macro just to safeguard if it's called incorrectly without updating it.

This will be addressed in a separate PR.

PR #2150, opened 2024-09-12
chore: refactor interceptor tests into a jinja macro

ohmayr:
This PR refactors the existing rest interceptor tests into a jinja macro which will allow us to re-use the same macro to generate tests for an async rest interceptor class in a follow up PR.

This PR should be reviewed and merged after: #2149.


gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
```diff
@@ -2325,3 +2261,89 @@ def test_initialize_client_w_{{transport_name}}():
...
+{% set async_prefix = get_async_prefix(is_async) %}
+{% set async_decorator = get_async_decorator(is_async) %}
+{% set transport_name = get_transport_name(transport, is_async) %}
+{% set method_name = method.name|snake_case %}
+{% set async_method_prefix = "Async" if is_async else "" %}
+{{async_decorator}}
+@pytest.mark.parametrize("null_interceptor", [True, False])
+{{async_prefix}}def test_{{ method_name }}_{{transport_name}}_interceptors(null_interceptor):
+{% if 'grpc' in transport %}
+    raise NotImplementedError("gRPC is currently not supported for this test case.")
```
parthea, line 2278: Please add a TODO comment for the outstanding task of adding support for gRPC

parthea (approved):

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
ohmayr: We don't support interceptors for `gRPC`. The guard is added to ensure that the macro isn't incorrectly called against the `gRPC` transport.

PR #2133, opened 2024-09-09
fix: resolve issue where explicit routing metadata was not sent in async clients

parthea:
This PR depends on https://github.com/googleapis/gapic-generator-python/pull/2126 and https://github.com/googleapis/gapic-generator-python/pull/2131. Please review those PRs first.

See https://google.aip.dev/client-libraries/4222

- Refactor `test_*_routing_parameters` so that it can be used for both sync and async test cases

Fixes https://github.com/googleapis/gapic-generator-python/issues/2091

PR #2165, opened 2024-09-16
test: add routing parameter and empty call test for REST

### LINKED ISSUE #2159: Update the `empty_call_test` and `routing_parameter_test` macro so that it can be used for both `gRPC` and `REST`
The [empty_call_test](https://github.com/googleapis/gapic-generator-python/blob/28097536e1a47063a5d3211e9c1c498a1f06c724/gapic/templates/tests/unit/gapic/%25name_%25version/%25sub/test_macros.j2#L1811
) macro only supports `gRPC`. We should update this macro so that it can be used for `REST` transport as well.

https://github.com/googleapis/gapic-generator-python/blob/4c5de8791795f8101f6ec66f80b8a8e5e9a21822/gapic/templates/tests/unit/gapic/%25name_%25version/%25sub/test_macros.j2#L1904-L1909

The same applies to `routing_parameter_test` which was introduced in PR https://github.com/googleapis/gapic-generator-python/pull/2133

parthea:
This PR adds routing parameter tests for REST transport

This PR depends on https://github.com/googleapis/gapic-generator-python/pull/2133 and https://github.com/googleapis/gapic-generator-python/pull/2131. Please review those PRs first.

Fixes https://github.com/googleapis/gapic-generator-python/issues/2159
Closes b/364925979

### DIFF
--- a/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
+++ b/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
@@ -1478,27 +1478,7 @@ def test_{{ method_name }}_rest_no_http_options():
     # Mock the actual call, and fake the request.
     {% if 'rest' in transport %}
     with mock.patch.object(
-            type(client.transport.{{ method.transport_safe_name|snake_case }}),
-            '__call__') as call, mock.patch.object(type(client.transport._session), 'request') as req:
-        req.return_value = mock.Mock()
-        req.return_value.status_code = 200
-        {% if method.void %}
-        return_value = None
-        json_return_value = ''
-        {% elif method.lro %}
-        return_value = operations_pb2.Operation(name='operations/spam')
-        json_return_value = json_format.MessageToJson(return_value)
-        {% else %}
-        return_value = {{ method.output.ident }}()
-        {% if method.output.ident.is_proto_plus_type %}
-        # Convert return value to protobuf type
-        return_value = {{ method.output.ident }}.pb(return_value)
-        {% endif %}
-        json_return_value = json_format.MessageToJson(return_value)
-        {% endif %}
-
-        req.return_value.read = mock.AsyncMock(return_value=json_return_value)
-        call.return_value = req.return_value
+            type(client.transport.{{ method.transport_safe_name|snake_case }}), '__call__') as call:
     {% else %}{# gRPC #}
     with mock.patch.object(
             type(client.transport.{{ method.transport_safe_name|snake_case }}),
--- a/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
+++ b/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
@@ -1466,6 +1466,10 @@ def test_{{ method_name }}_rest_no_http_options():
 @pytest.mark.asyncio
 {% endif %}{# is_async #}
 {{ async_method_prefix }}def test_{{ method_name }}_{{ test_name }}_{{transport_name}}():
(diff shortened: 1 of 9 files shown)


gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
```diff
@@ -1468,9 +1468,34 @@ def test_{{ method_name }}_rest_no_http_options():
...
+        {% else %}
+        return_value = {{ method.output.ident }}()
+        {% if method.output.ident.is_proto_plus_type %}
+        # Convert return value to protobuf type
+        return_value = {{ method.output.ident }}.pb(return_value)
+        {% endif %}
+        json_return_value = json_format.MessageToJson(return_value)
+        {% endif %}
+
+        req.return_value.read = mock.AsyncMock(return_value=json_return_value)
```
ohmayr, line 1500: this doesn't seem right for the case of sync and should be guarded with a conditional statement.

We may instead want something like:
```
{% if is_async %}
req.return_value.read = mock.AsyncMock(return_value=json_return_value)
{% else %}
response_value.content = json_return_value.encode
{% endif %}

```

ohmayr: If the test case isn't failing because of this then there may be something going wrong here.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
```diff
@@ -1466,11 +1466,40 @@ def test_{{ method_name }}_rest_no_http_options():
...
+    {% endif %}
     client = {{ get_client(service=service, is_async=is_async) }}(
         credentials={{ get_credentials(is_async=is_async) }},
+        transport="{{ transport_name }}",
     )
 
-    # Mock the actual call within the gRPC stub, and fake the request.
+    # Mock the actual call, and fake the request.
+    {% if 'rest' in transport %}
+    with mock.patch.object(
```
ohmayr, line 1480: if we're mocking the the `__call__` method, then why do we need to mock the underlying response that it returns below?

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
```diff
@@ -1466,11 +1466,20 @@ def test_{{ method_name }}_rest_no_http_options():
...
+            type(client.transport.{{ method.transport_safe_name|snake_case }}), '__call__') as call:
```
ohmayr, line 1481: the `if` and `else` statements for rest and gRPC look the same to me and can be combined? WDYT?

ohmayr (approved):

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
parthea: Fixed in b6f76934ae12d207c26dce86f94b499f8eb6d258

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
parthea: Fixed in b6f76934ae12d207c26dce86f94b499f8eb6d258

ohmayr:
We may want to clean this up: https://github.com/googleapis/gapic-generator-python/pull/2165/files#diff-0f64a5dcd8e2f3ac9b10417f899604726f8111968a1b5e46b6339c39df071af0R2813-R2815.

PR #2172, opened 2024-09-17
chore: update auth import error statement

ohmayr:


tests/system/test_universe_domain.py
```diff
@@ -73,9 +73,9 @@ def test_universe_domain_validation_fail(parametrized_echo, channel_creator, tra
     # Make this test unconditional once the minimum supported version of
     # google-auth becomes 2.23.0 or higher.
     google_auth_major, google_auth_minor, _ = [
-        int(part) for part in google.auth.__version__.split(".")
+        part for part in google.auth.__version__.split(".")
     ]
-    if google_auth_major > 2 or (google_auth_major == 2 and google_auth_minor >= 23):
+    if int(google_auth_major) > 2 or (int(google_auth_major) == 2 and int(google_auth_minor) >= 23):
```
parthea, line 78: Is this change necessary?

parthea (approved):

tests/system/test_universe_domain.py
ohmayr: I'm using the pre-release version in the noxfile so the patch version becomes something like `rcX` when split. And parsing it to int within the for loop fails.

PR #2175, opened 2024-09-20
test: add showcase tests for mixins

parthea:
Towards https://github.com/googleapis/google-cloud-python/issues/16275

Depends on https://github.com/googleapis/gapic-generator-python/pull/2186

PR #2181, opened 2024-09-20
feat: implement async rest mixin methods

ohmayr:
Fixes: https://github.com/googleapis/google-cloud-python/issues/16279, https://github.com/googleapis/google-cloud-python/issues/16275

### DIFF
--- a/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/_shared_macros.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/_shared_macros.j2
@@ -235,7 +235,7 @@ def _get_http_options():
 {% endmacro %}
 
 
-{% macro prep_wrapped_messages_async_method(service) %}
+{% macro prep_wrapped_messages_async_method(api, service) %}
 def _prep_wrapped_messages(self, client_info):
     """ Precompute the wrapped methods, overriding the base class method to use async wrappers."""
     self._wrapped_methods = {
@@ -265,6 +265,13 @@ def _prep_wrapped_messages(self, client_info):
             client_info=client_info,
         ),
         {% endfor %}{# service.methods.values() #}
+        {% for method_name in api.mixin_api_methods.keys() %}
+        self.{{ method_name|snake_case }}: self._wrap_method(
+            self.{{ method_name|snake_case }},
+            default_timeout=None,
+            client_info=client_info,
+        ),
+        {% endfor %} {# method_name in api.mixin_api_methods.keys() #}
     }
 {% endmacro %}
 
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/_async_mixins.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/_async_mixins.py.j2
@@ -370,7 +370,7 @@
             request = iam_policy_pb2.SetIamPolicyRequest(**request)
 
         # Wrap the RPC method; this adds retry and timeout information,
-        # and friendly error handling
+        # and friendly error handling.
         rpc = self.transport._wrapped_methods[self._client._transport.set_iam_policy]
 
         # Certain fields should be provided within the metadata header;
@@ -604,7 +604,6 @@
         # Wrap the RPC method; this adds retry and timeout information,
         # and friendly error handling.
         rpc = self.transport._wrapped_methods[self._client._transport.get_location]
-        
 
         # Certain fields should be provided within the metadata header;
         # add these here.
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/_async_mixins.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/_async_mixins.py.j2
@@ -34,11 +34,7 @@
 
         # Wrap the RPC method; this adds retry and timeout information,
         # and friendly error handling.
-        rpc = gapic_v1.method_async.wrap_method(
-            self._client._transport.list_operations,
-            default_timeout=None,
-            client_info=DEFAULT_CLIENT_INFO,
-        )
+        rpc = self.transport._wrapped_methods[self._client._transport.list_operations]
 
         # Certain fields should be provided within the metadata header;
         # add these here.
@@ -91,11 +87,7 @@
 
         # Wrap the RPC method; this adds retry and timeout information,
         # and friendly error handling.
-        rpc = gapic_v1.method_async.wrap_method(
-            self._client._transport.get_operation,
-            default_timeout=None,
-            client_info=DEFAULT_CLIENT_INFO,
-        )
+        rpc = self.transport._wrapped_methods[self._client._transport.get_operation]
 
         # Certain fields should be provided within the metadata header;
         # add these here.
@@ -151,11 +143,7 @@
 
         # Wrap the RPC method; this adds retry and timeout information,
         # and friendly error handling.
-        rpc = gapic_v1.method_async.wrap_method(
-            self._client._transport.delete_operation,
-            default_timeout=None,
-            client_info=DEFAULT_CLIENT_INFO,
-        )
+        rpc = self.transport._wrapped_methods[self._client._transport.delete_operation]
 
         # Certain fields should be provided within the metadata header;
         # add these here.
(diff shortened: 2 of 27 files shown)


gapic/templates/%namespace/%name_%version/%sub/services/%service/_async_mixins.py.j2
```diff
@@ -91,10 +91,15 @@
...
+            gapic_version=DEFAULT_CLIENT_INFO.gapic_version,
+            grpc_version=None,
+            rest_version=google.auth.__version__
+        )
+        rpc = self._client._transport._wrap_method(  # type: ignore
             self._client._transport.get_operation,
             default_timeout=None,
-            client_info=DEFAULT_CLIENT_INFO,
+            client_info=TEMP_CLIENT_INFO,
         )
```
parthea, line 103: We could move this to `_wrapped_methods` similar to PR https://github.com/googleapis/gapic-generator-python/pull/2175.
In that case, the code here would be
```
rpc = self._transport._wrapped_methods[self._transport.get_operation]
```
also see the changes in `gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/base.py.j2` which are needed to update `_prep_wrapped_messages`

WDYT?

ohmayr: What I've done here is a hack (I was trying to figure out why the showcase tests were failing) but isn't something we should be doing.

I think what you're suggesting makes sense. We should precompute the wrapped logic for mixins similar to what we do for other methods.

ohmayr: I took a look at the changes in #2175. I'd like to point out that we override `_prep_wrapped_messages` for async transports (both rest and grpc). So the relevant logic should go in there as well.

We also want to make sure that the `kind` property is set (to avoid a gRPC code path for a rest call), if available.

We can address this issue separately first and then rebase our work on top of it.

parthea: Rebasing on top of #2175 sounds good

ohmayr: As per https://github.com/googleapis/gapic-generator-python/pull/2175#issuecomment-2369809636, #2175 will be rebased on top of this PR.

ohmayr: wrapping logic for mixins has been addressed.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc_asyncio.py.j2
```diff
@@ -388,7 +388,7 @@ class {{ service.grpc_asyncio_transport_name }}({{ service.name }}Transport):
         return self._stubs["test_iam_permissions"]
     {% endif %}
 
-    {{ shared_macros.prep_wrapped_messages_async_method(service)|indent(4) }}
```
vchudnov-g, line 391: I'm concerned in that higher up in this functions, we add `set_iam_policy` if `opts.add_iam_methods`. But the TODO above `{% if opts.add_iam_methods %}` makes it seem like that may not be used any more. WDYT?

Incidentally, I just filed #2196 to remove these TODOs.

tests/integration/goldens/credentials/google/iam/credentials_v1/services/iam_credentials/async_client.py
```diff
@@ -23,6 +23,7 @@
 from google.api_core import exceptions as core_exceptions
 from google.api_core import gapic_v1
 from google.api_core import retry_async as retries
+import google.auth
```
vchudnov-g, line 26: Why do we need `auth` here and in other files? I don't see references to `auth.*` in the file.

tests/integration/goldens/redis/google/cloud/redis_v1/services/cloud_redis/transports/rest_asyncio.py
```diff
@@ -205,6 +205,127 @@ async def post_get_instance_auth_string(self, response: cloud_redis.InstanceAuth
...
+        self, response: locations_pb2.Location
+    ) -> locations_pb2.Location:
+        """Post-rpc interceptor for get_location
+
+        Override in a subclass to manipulate the response
+        after it is returned by the CloudRedis server but before
+        it is returned to user code.
+        """
+        return response
+    async def pre_list_locations(
```
vchudnov-g, line 228: Spaces between methods:

```suggestion
        return response
        
    async def pre_list_locations(
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_asyncio.py.j2
```diff
@@ -225,6 +225,9 @@ class Async{{service.name}}RestTransport(_Base{{ service.name }}RestTransport):
         return self._{{method.name}}(self._session, self._host, self._interceptor)  # type: ignore
 
     {% endfor %}
+    {% for name, sig in api.mixin_api_signatures.items() %} 
+    {{ shared_macros.generate_mixin_call_method(service, api, name, sig, is_async=True) | indent(4) }}
```
vchudnov-g, line 229: This doesn't need to block, but would be a good follow-up: this macro is defining the each `_Mixin` class  next to the corresponding `@property`. The other `_Method` classes are defined all together, and then their `@property` all together. I think it would be good to define the `_mixin` classes right after the `_Method` classes, and the mixin `@property` right after the method `@property`. WDYT? Could you file an issue so we don't forget?

vchudnov-g (approved):
Looks good.

A non-trivial comment about the auth import. Some layout comments an nits. A question.  And a suggestion for a follow-up.

gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/_shared_macros.j2
```diff
@@ -265,6 +265,13 @@ def _prep_wrapped_messages(self, client_info):
             client_info=client_info,
         ),
         {% endfor %}{# service.methods.values() #}
+        {% for method_name in api.mixin_api_methods.keys() %}
```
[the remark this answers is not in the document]
ohmayr: I'm not sure if `transport_safe_name` is available in `api.mixin_api_methods.keys() `. I've filed and linked an issue to follow up on this.

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
```diff
@@ -632,11 +629,7 @@ class {{ service.async_client_name }}:
 
         # Wrap the RPC method; this adds retry and timeout information,
         # and friendly error handling.
-        rpc = gapic_v1.method_async.wrap_method(
-            self._client._transport.get_iam_policy,
-            default_timeout=None,
-            client_info=DEFAULT_CLIENT_INFO,
-        )
+        rpc = self.transport._wrapped_methods[self._client._transport.get_iam_policy]
```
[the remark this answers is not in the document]
ohmayr: I may have updated all the instances just because it makes sense to use `wraped_methods` everywhere. However, the methods defined in async client aren't generated / used. There's a linked PR which mentions that those methods should be cleaned up. We can follow up on it. Thanks for filing the issue!

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc_asyncio.py.j2
ohmayr: You're right, the methods defined within async client aren't used. The ones defined in async mixins file are.

tests/integration/goldens/credentials/google/iam/credentials_v1/services/iam_credentials/async_client.py
ohmayr: addressed.

tests/integration/goldens/redis/google/cloud/redis_v1/services/cloud_redis/transports/rest_asyncio.py
ohmayr: addressed.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_asyncio.py.j2
ohmayr: I've filed and linked an issue to follow up on this.

PR #2186, opened 2024-09-20
chore: pre-compute wrappers for mixins

parthea:
Towards https://github.com/googleapis/google-cloud-python/issues/16275

### DIFF
--- a/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/_rest_mixins_base.py.j2
+++ b/gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/_rest_mixins_base.py.j2
@@ -26,7 +26,9 @@
 
     {% for name, sig in api.mixin_api_signatures.items() %}
     class _Base{{ name }}:
-        
+        def __hash__(self):
+            return hash("{{ service.name }}RestTransport.{{ name }}")
+
         {{ shared_macros.http_options_method(api.mixin_http_options["{}".format(name)])|indent(8)}}
 
         @staticmethod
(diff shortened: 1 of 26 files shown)


gapic/ads-templates/%namespace/%name/%version/%sub/services/%service/transports/_rest_mixins_base.py.j2
```diff
@@ -26,7 +26,9 @@
 
     {% for name, sig in api.mixin_api_signatures.items() %}
     class _Base{{ name }}:
-        
+        def __hash__(self):
```
ohmayr, line 29: @parthea I think we can define an unimplemented method in the base class similar to what we're doing in `rest_base.py.j2`

```python
def __hash__(self):  # pragma: NO COVER
    return NotImplementedError("__hash__ must be implemented.")
```

And we can define the hash method in the respective sync and async classes. Since the child classes are generated via `shared_macros.generate_mixin_call_method(service, api, name, sig, is_async=False)`, we can update the macro to include:

```python3
 def __hash__(self):
     return hash("{{ async_method_name_prefix }}{{ service.name }}RestTransport.{{ name }}")
```

parthea: Fixed in https://github.com/googleapis/gapic-generator-python/pull/2186/commits/b983346060aed2c5cbbc24a3e51157036808a038 and https://github.com/googleapis/gapic-generator-python/pull/2186/commits/110ac1422e1303244f15c12bd9b0d806787e8b13

ohmayr (approved):

ohmayr:
@parthea I think it makes sense to update all the methods of a sync client to use `self._transport._wrapped_methods` for wrapping logic within this PR. I will be addressing that shortly.

ohmayr:
> Note: pre-compute wrappers for async mixins will be addressed in a follow up PR.

I got confused with some duplicated methods for mixins / iam methods which aren't generated anymore and are supposed to be cleaned up. See: https://github.com/googleapis/gapic-generator-python/pull/1240. We can clean that up separately.

Issue #2171, opened 2024-09-17
Add support for asynchronous rest pagers


ohmayr:
Implement async rest pager api methods in the async rest transport layer to add support for async rest pagination.

Also add unit test coverage and add an e2e showcase test for it.

The related code change locations can be looked up by searching for this issue.

PR #2194, opened 2024-09-24
feat: implement async rest pager methods

ohmayr:
Fixes https://github.com/googleapis/gapic-generator-python/issues/2171 🦕

### DIFF
--- a/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
+++ b/gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
@@ -1005,12 +1005,11 @@ def test_{{ method_name }}_raw_page_lro():
 {# TODO(kbandes): remove this if condition when lro and client streaming are supported. #}
 {% if not method.client_streaming %}
 {# TODO(https://github.com/googleapis/gapic-generator-python/issues/2170): Remove unit test for long running operation method. #}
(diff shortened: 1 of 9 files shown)


gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
```diff
@@ -2242,6 +2240,11 @@ def test_initialize_client_w_{{transport_name}}():
         {% endif %}{# method.server_streaming #}
         req.return_value = response_value
         response = {{ await_prefix }}client.{{ method_name }}(request)
+    {% if "next_page_token" in method_output.fields.values()|map(attribute='name', default="") and not method.paged_result_field %}
+    {# Cheeser assertion to force code coverage for bad paginated methods #}
```
parthea, line 2244: Please could you add more detail to help better understand the bad paginated methods?

parthea (approved):

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
ohmayr: This is copied from another test within the same file, which I've used as reference for this macro. This specific assertion is rarely generated. I'm unsure what these bad paginated methods are and don't think we should spend time investigating on it given the approaching deadline. I've filed https://github.com/googleapis/gapic-generator-python/issues/2199 and linked the bug here to follow up on it.

PR #2195, opened 2024-09-24
chore: add `async_rest` extra for async rest dependencies

### LINKED ISSUE #2187: Add an `async_rest` extra for GAPICs
In order to support async rest in GAPICs, we should add an `async_rest` extra in `setup.py` which installs the necessary dependencies required for the feature such as the correct version of `google-api-core` and `google-auth` with `aiohttp` extra.

ohmayr:
Fixes googleapis/gapic-generator-python#2187 🦕


gapic/templates/testing/_default_constraints.j2
```diff
@@ -1,7 +1,12 @@
+{# TODO(https://github.com/googleapis/gapic-generator-python/issues/2121): Remove the following variable (and the condition later in this file) for async rest transport once support for it is GA. #}
+{% set rest_async_io_enabled = api.all_library_settings[api.naming.proto_package].python_settings.experimental_features.rest_async_io_enabled %}
 {% from '_pypi_packages.j2' import pypi_packages %}
 # This constraints file is required for unit tests.
 # List all library dependencies and extras in this file.
 google-api-core
+{% if rest_async_io_enabled %}
+google-auth[aiohttp]==2.35.0
```
parthea, line 8: We should get the latest version of `google-auth` if we don't specify a version. If we're not getting the right version, we can try adding `--force-reinstall` to the install command

```suggestion
google-auth
# from google-auth[aiohttp]
aiohttp
requests
```

gapic/templates/testing/constraints-3.7.txt.j2
```diff
@@ -1,12 +1,19 @@
...
 {% from '_pypi_packages.j2' import pypi_packages %}
 # This constraints file is used to check that lower bounds
 # are correct in setup.py
 # List all library dependencies and extras in this file.
 # Pin the version to the lower bound.
 # e.g., if setup.py has "google-cloud-foo >= 1.14.0, < 2.0.0dev",
 # Then this file should have google-cloud-foo==1.14.0
+{% if rest_async_io_enabled %}
+google-api-core==2.20.0
+google-auth[aiohttp]==2.35.0
```
parthea, line 12:
```suggestion
google-auth==2.35.0
# from google-auth[aiohttp]
aiohttp==3.6.2
requests==2.20.0
```
```

Comment (ohmayr) [reply gapic/templates/testing/constraints-3.7.txt.j2]: addressed.

File: gapic/templates/testing/_default_constraints.j2 (hunk 1 above)
Comment (ohmayr) [reply gapic/templates/testing/_default_constraints.j2]: addressed.

File: gapic/templates/testing/_default_constraints.j2 [hunk 3]
```diff
@@ -1,7 +1,15 @@
...
+{% set rest_async_io_enabled = api.all_library_settings[api.naming.proto_package].python_settings.experimental_features.rest_async_io_enabled %}
 {% from '_pypi_packages.j2' import pypi_packages %}
 # This constraints file is required for unit tests.
 # List all library dependencies and extras in this file.
 google-api-core
+{% if rest_async_io_enabled %}
+google-auth
+# from google-auth[aiohttp]
+aiohttp
+requests
```
parthea, line 11: We shouldn't include the `async_rest` extra in all constraints because we won't have any tests without the `async_rest` extra installed.

gapic/templates/testing/constraints-3.7.txt.j2
```diff
@@ -1,12 +1,22 @@
...
 # List all library dependencies and extras in this file.
 # Pin the version to the lower bound.
 # e.g., if setup.py has "google-cloud-foo >= 1.14.0, < 2.0.0dev",
 # Then this file should have google-cloud-foo==1.14.0
+{% if rest_async_io_enabled %}
+google-api-core==2.20.0
+google-auth==2.35.0
+# from google-auth[aiohttp]
+aiohttp==3.6.2
+requests==2.20.0
```
parthea, line 15: Consider creating a separate constraints file so that we know that the client library without the `async_rest` extra  works. See https://github.com/googleapis/python-api-core/pull/701 where we have a separate constraints file `testing/constraints-async-rest-3.7.txt` and separate nox session for testing `async_rest`

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_asyncio.py.j2
```diff
@@ -7,30 +7,22 @@
...
-{# TODO(https://github.com/googleapis/gapic-generator-python/issues/2187): Add an `async_rest` extra for GAPICs and update the error message to include that the extra needs to be installed. #}
 try:
     import aiohttp # type: ignore
     from google.auth.aio.transport.sessions import AsyncAuthorizedSession # type: ignore
+    from google.api_core import rest_streaming_async # type: ignore
 {# NOTE: `pragma: NO COVER` is needed since the coverage for presubmits isn't combined. #}
 except ImportError as e:  # pragma: NO COVER
-    {# TODO(https://github.com/googleapis/google-auth-library-python/pull/1577): Update the version of google-auth once the linked PR is merged. #}
-    raise ImportError("async rest transport requires google-auth >= 2.35.0 with aiohttp extra. Install google-auth with the aiohttp extra using `pip install google-auth[aiohttp]==2.35.0`.") from e
+    raise ImportError("`rest_asyncio` transport requires the library to be installed with the `async_rest` extra. Install the library with the `async_rest` extra using `pip install <insert_library_name>[async_rest]`") from e
```
parthea, line 16: Can we use `{{ api.naming.warehouse_package_name }}` instead of `<insert_library_name>`?

tests/integration/goldens/redis/tests/unit/gapic/redis_v1/test_cloud_redis.py
```diff
@@ -8781,12 +8776,8 @@ def test_initialize_client_w_rest():
 
 
 def test_transport_kind_rest_asyncio():
-    if not HAS_GOOGLE_AUTH_AIO:
-        pytest.skip("google-auth >= 2.35.0 is required for async rest transport.")
-    elif not HAS_AIOHTTP_INSTALLED:
-        pytest.skip("aiohttp is required for async rest transport.")
-    elif not HAS_ASYNC_REST_SUPPORT_IN_CORE:
-        pytest.skip("google-api-core >= 2.20.0 is required for async rest transport.")
+    if not HAS_ASYNC_REST_EXTRA:
```
vchudnov-g, line 8779: Before, you were also checking for `ga_credentials_async` being installed, which was and is guarded by `HAS_GOOGLE_AUTH_AIO` but not by `HAS_ASYNC_REST_EXTRA`. Do you not need that import, then?

tests/integration/goldens/redis/setup.py
```diff
@@ -46,6 +46,12 @@
     "proto-plus >= 1.22.3, <2.0.0dev",
     "protobuf>=3.20.2,<6.0.0dev,!=4.21.0,!=4.21.1,!=4.21.2,!=4.21.3,!=4.21.4,!=4.21.5",
 ]
+extras = {
```
vchudnov-g, line 49: Where do you use `extras`? From what I gather (eg https://xebia.com/blog/a-practical-guide-to-using-setup-py/, https://sukhbinder.wordpress.com/2023/04/09/a-short-primer-on-extra_requires-in-setup-py/) you need to pass this to `setup(..., extras_require = extras)`

tests/integration/goldens/redis/google/cloud/redis_v1/services/cloud_redis/transports/rest_asyncio.py
```diff
@@ -18,8 +18,9 @@
 try:
     import aiohttp # type: ignore
     from google.auth.aio.transport.sessions import AsyncAuthorizedSession # type: ignore
+    from google.api_core import rest_streaming_async # type: ignore
 except ImportError as e:  # pragma: NO COVER
-    raise ImportError("async rest transport requires google-auth >= 2.35.0 with aiohttp extra. Install google-auth with the aiohttp extra using `pip install google-auth[aiohttp]==2.35.0`.") from e
+    raise ImportError("`rest_asyncio` transport requires the library to be installed with the `async_rest` extra. Install the library with the `async_rest` extra using `pip install <insert_library_name>[async_rest]`") from e
```
vchudnov-g, line 23:
```suggestion
    raise ImportError("`rest_asyncio` transport requires the library to be installed with the `async_rest` extra. Install the library with the `async_rest` extra using `pip install LIBRARY_NAME[async_rest]`") from e
```

We could fill in the `LIBRARY_NAME`, no? (We seem to have the package name in `setup.py:name`). I think that would aid usability.

tests/integration/goldens/redis/tests/unit/gapic/redis_v1/test_cloud_redis.py
ohmayr: The newer version of auth will have `ga_credentials_async` so we don't need to check that.

gapic/templates/testing/_default_constraints.j2
ohmayr: addressed.

gapic/templates/testing/constraints-3.7.txt.j2
ohmayr: addressed.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/rest_asyncio.py.j2
ohmayr: addressed.

tests/integration/goldens/redis/setup.py
ohmayr: Good catch Victor! I missed setting extras_require  in setup. Fixed now.

tests/integration/goldens/redis/google/cloud/redis_v1/services/cloud_redis/transports/rest_asyncio.py
ohmayr: addressed.

tests/integration/goldens/redis/tests/unit/gapic/redis_v1/test_cloud_redis.py
vchudnov-g: "will have": Is this still work to be done? If so, we should make sure we enable the generator to work with the current version of auth.

ohmayr: To clarify even further:

`ga_credentials_async` was introduced in `v2.33.0`. It's given that the features introduced earlier exist in `v2.35.0`.

So checking for `HAS_ASYNC_REST_EXTRA` and `HAS_GOOGLE_AUTH_AIO` is redundant in the case here.

noxfile.py
```diff
@@ -300,8 +300,9 @@ def showcase_library(
             # supported version of a dependency.
             # This is needed to recreate the issue reported in
             # https://github.com/googleapis/google-cloud-python/issues/12254
+            constraints_type = "async-rest-" if rest_async_io_enabled else ""
             constraints_path = str(
-            f"{tmp_dir}/testing/constraints-{session.python}.txt"
+            f"{tmp_dir}/testing/constraints-{constraints_type}{session.python}.txt"
```
vchudnov-g, line 305: A note: unlike what I remarked on another comment thread in this PR, where we're using the constraints files as `-c` arguments to PIP (which hence do not control whether those dependencies are installed), here we are passing the file via `-r` (three lines below) thus making it into a _requirements_ file where the listed dependencies _are_ installed. AFAICT, we do  _not_ use `-r` for the (same??!) constraints files under `templates/`.

We should be clear about the semantics and how we are using the files. I'd suggest for the purposes of _this_ file, creating proper requirements files that do list what will be installed.

gapic/templates/setup.py.j2
```diff
@@ -94,6 +98,9 @@ setuptools.setup(
     packages=packages,
     python_requires=">=3.7",
     install_requires=dependencies,
+{% if rest_async_io_enabled %}
+    extras_require=extras,
```
vchudnov-g, line 102: So if this wasn't there before, why weren't some tests failing? This would suggest a failure in our tests. At the GAPIC level, the test should silently pass if the GAPIC was generated without REST transport or installed without the extra. But at the generator level, we want to ensure that the tests with the extras pass, so how can we be notified if we run a test suite where we expect the extra and the associated tests, but it turns out the extra wasn't installed?

ohmayr: Tests weren't failing because we're not using the `async_rest` extra to install the dependencies. We've explicitly defined each required dependency within the constraints file.

gapic/templates/setup.py.j2
```diff
@@ -46,12 +48,14 @@ dependencies = [
...
     {% endfor %}
 ]
+{% if rest_async_io_enabled %}
 extras = {
     "async_rest": [
         "google-api-core[grpc] >= 2.20.0, < 3.0.0dev",
         "google-auth[aiohttp] >= 2.35.0, <3.0.0dev"
     ],
 }
+{% endif %}
```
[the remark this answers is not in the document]
ohmayr: I've addressed this but I personally preferred the previous solution only because that didn't update the setup file in every GAPIC with an empty extra.

noxfile.py
ohmayr: Removed the constraints file.

We're infact installing the `constraints` file with a `-r` flag which is why this wasn't an issue here (at least at the generator layer). Anyways, this has been addressed by installing the deps with the extra.

gapic/templates/testing/_default_constraints.j2
ohmayr: addressed. See my other comment and discussing this over our offline chat.

gapic/templates/setup.py.j2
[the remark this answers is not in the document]
vchudnov-g: Either way is fine. This makes it explicit there are no extras, on purpose.

noxfile.py
```diff
@@ -304,7 +304,16 @@ def showcase_library(
...
             )
             # Install the library with a constraints file.
-            session.install("-e", tmp_dir, "-r", constraints_path)
+            if session.python == "3.7":
+                session.install("-e", tmp_dir, "-r", constraints_path)
+                if rest_async_io_enabled:
+                    # NOTE: We re-install `google-api-core` and `google-auth` to override the respective
+                    # versions for each specified in constraints-3.7.txt. This is needed because async REST
+                    # is not supported with the minimum version of `google-api-core` and `google-auth`.
+                    session.install('--no-cache-dir', '--force-reinstall', "google-api-core")
```
vchudnov-g, line 313: Since the 3.7 constraints file is meant to address the minimum version of our dependencies, could we have these installs hard code the minimum version of `google-api-core` and `google-auth` that work with async REST?

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
```diff
@@ -1761,12 +1761,8 @@ def test_{{ method_name }}_rest_no_http_options():
...
 def test_transport_kind_{{ transport_name }}():
     {% if transport_name == 'rest_asyncio' %}
-    if not HAS_GOOGLE_AUTH_AIO:
-        pytest.skip("google-auth >= 2.35.0 is required for async rest transport.")
-    elif not HAS_AIOHTTP_INSTALLED:
-        pytest.skip("aiohttp is required for async rest transport.")
-    elif not HAS_ASYNC_REST_SUPPORT_IN_CORE:
-        pytest.skip("google-api-core >= 2.20.0 is required for async rest transport.")
+    if not HAS_ASYNC_REST_EXTRA:
+        pytest.skip("the library must be installed with `async_rest` extra to use this feature.")
```
vchudnov-g, line 1765: Could you file an issue to follow up on making sure or documenting how to see these skip messages? Also, mention that in some cases we should use `mark.skipif`: https://docs.pytest.org/en/stable/reference/reference.html#pytest-mark-skipif-ref

noxfile.py
```diff
@@ -304,7 +304,16 @@ def showcase_library(
...
+                    session.install('--no-cache-dir', '--force-reinstall', "google-auth[aiohttp]")
+            else:
+                session.install("-e", tmp_dir + ("[async_rest]" if rest_async_io_enabled else ""), "-r", constraints_path)
```
vchudnov-g, line 316: `tmp_dir` is just a `DIRECTORY_NAME`. So the syntax `-e DIRECTORY_NAME[async_rest]` works? OK...

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
vchudnov-g, line 1765:
```suggestion
        pytest.skip("the library must be installed with the `async_rest` extra to test this feature.")
```

Here and below

vchudnov-g (approved):

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
ohmayr: addressed.

noxfile.py
ohmayr: Yes it works.

gapic/templates/tests/unit/gapic/%name_%version/%sub/test_macros.j2
ohmayr: Filed https://github.com/googleapis/gapic-generator-python/issues/2203

noxfile.py
ohmayr: addressed.

PR #2210, opened 2024-10-09
chore: update api-core version in async extra

### LINKED ISSUE #2208: [async rest] Update the minimum version of `core` in async rest option
We're currently using the pre-release version of `google-api-core` i.e. `2.21.0rc0` in the `async_rest` option in `setup.py.j2`. Update it to `2.21.0` when released.

Also update the noxfile for the generator where we're using the pre-release version.

ohmayr:
Fixes #2208 🦕

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -310,8 +310,7 @@ def showcase_library(
                     # NOTE: We re-install `google-api-core` and `google-auth` to override the respective
                     # versions for each specified in constraints-3.7.txt. This is needed because async REST
                     # is not supported with the minimum version of `google-api-core` and `google-auth`.
-                    # TODO(https://github.com/googleapis/gapic-generator-python/issues/2208): Update the minimum supported version of api-core to `2.21.0` when released.
-                    session.install('--no-cache-dir', '--force-reinstall', "google-api-core[grpc, async_rest]==2.21.0rc0")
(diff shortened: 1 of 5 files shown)


noxfile.py
```diff
@@ -310,8 +310,7 @@ def showcase_library(
                     # NOTE: We re-install `google-api-core` and `google-auth` to override the respective
                     # versions for each specified in constraints-3.7.txt. This is needed because async REST
                     # is not supported with the minimum version of `google-api-core` and `google-auth`.
-                    # TODO(https://github.com/googleapis/gapic-generator-python/issues/2208): Update the minimum supported version of api-core to `2.21.0` when released.
-                    session.install('--no-cache-dir', '--force-reinstall', "google-api-core[grpc, async_rest]==2.21.0rc0")
+                    session.install('--no-cache-dir', '--force-reinstall', "google-api-core[grpc, async_rest]==2.21.0")
```
parthea, line 313: Please file a TODO bug to remove the hardcoded minimum version here. It should be in a constraints file. We can also create a bug to improve documentation around constraints files and how/when they should be used.

parthea (approved):

noxfile.py
ohmayr: addressed here https://github.com/googleapis/gapic-generator-python/pull/2210/commits/15a45e98cefd695539c9e5459777bd77f082c3ee.

parthea (approved):

PR #2212, opened 2024-10-09
fix: add default library settings for incorrect lib version

ohmayr:
consider the case where `self.naming.proto_package` differs from `library_settings.version`. Ideally, this shouldn't happen but if the version configured against `library_settings.version` is incorrect, it breaks our generator.

As an example, see: https://github.com/googleapis/googleapis/blob/master/google/cloud/confidentialcomputing/v1alpha1/confidentialcomputing_v1alpha1.yaml#L63 where the service yaml file for confidentialcomputing specifies an incorrect version for library settings.
In this case:
- `self.naming.proto_package = google.cloud.confidentialcomputing.v1alpha1`
- `library_settings.version = google.cloud.confidentialcomputing.v1`

### DIFF
--- a/gapic/schema/api.py
+++ b/gapic/schema/api.py
@@ -711,11 +711,9 @@ class API:
         # - if library settings are not specified in the service config.
         # - if library_settings.version != self.naming.proto_package (proto package name)
         if self.naming.proto_package not in result:
-            result = {
-                self.naming.proto_package: client_pb2.ClientLibrarySettings(
+            result[self.naming.proto_package] = client_pb2.ClientLibrarySettings(
                     version=self.naming.proto_package
                 )
-            }
 
         return result
 
--- a/gapic/schema/api.py
+++ b/gapic/schema/api.py
@@ -712,7 +712,7 @@ class API:
         # - if library_settings.version != self.naming.proto_package (proto package name)
         if self.naming.proto_package not in result:
             result[self.naming.proto_package] = client_pb2.ClientLibrarySettings(
-                    version=self.naming.proto_package
+                version=self.naming.proto_package
                 )
 
         return result
--- a/gapic/schema/api.py
+++ b/gapic/schema/api.py
@@ -706,8 +706,11 @@ class API:
             for library_setting in self.service_yaml_config.publishing.library_settings
         }
 
-        # Add default settings for the current proto package
-        if not result:
+        # NOTE: Add default settings for the current proto package
+        # for the following cases:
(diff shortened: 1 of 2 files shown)


gapic/schema/api.py
```diff
@@ -706,13 +706,14 @@ def all_library_settings(
...
-                self.naming.proto_package: client_pb2.ClientLibrarySettings(
-                    version=self.naming.proto_package
+        # NOTE: Add default settings for the current proto package
+        # for the following cases:
+        # - if library settings are not specified in the service config.
+        # - if library_settings.version != self.naming.proto_package (proto package name)
+        if self.naming.proto_package not in result:
+            result[self.naming.proto_package] = client_pb2.ClientLibrarySettings(
+                version=self.naming.proto_package
                 )
```
parthea, line 716:
```suggestion
            result[self.naming.proto_package] = client_pb2.ClientLibrarySettings(
                version=self.naming.proto_package
            )
```

parthea (commented):
LGTM once a unit test is added that fails without the fix

gapic/schema/api.py
ohmayr: this results in the style check to fail.

parthea (approved):

gapic/schema/api.py
```diff
@@ -706,13 +706,14 @@ def all_library_settings(
...
-            result = {
-                self.naming.proto_package: client_pb2.ClientLibrarySettings(
-                    version=self.naming.proto_package
+        # NOTE: Add default settings for the current proto package
+        # for the following cases:
+        # - if library settings are not specified in the service config.
+        # - if library_settings.version != self.naming.proto_package (proto package name)
+        if self.naming.proto_package not in result:
+            result[self.naming.proto_package] = client_pb2.ClientLibrarySettings(
+                version=self.naming.proto_package
```
vchudnov-g, line 715: To confirm, you're not passing `python_settings` here like you are in line 704 because we _intend_ to pass empty settings since they weren't explicitly specified for `self.naming.proto_package` version?

ohmayr: We want to set empty settings against `self.naming.proto_package` if the key is not already set in line 704 which happens only when someone has an incorrect `library_settings.version` configured in their service yaml. We're doing this so our generator doesn't if it fails to find the settings that it's looking for.

vchudnov-g: Sg. You're confirming the intent I phrased, so I understood correctly. Thanks!

PR #2224, opened 2024-10-16
build: add presubmits for unit tests in golden files

### LINKED ISSUE #2073: Run all unit tests for golden files
Currently for golden files we only have `lint` and `prerelease_deps` presubmit checks. See below:

`goldens-lint`  - runs mypy and lint checks
https://github.com/googleapis/gapic-generator-python/blob/8d6a035f2c91cab0bc9fad25b7c0af5c5bfa63af/.github/workflows/tests.yaml#L310-L328

`goldens-prerelease` - runs unit tests against pre-release dependencies
https://github.com/googleapis/gapic-generator-python/blob/8d6a035f2c91cab0bc9fad25b7c0af5c5bfa63af/.github/workflows/tests.yaml#L329-L349

We could consider running unit tests as well, especially the unit tests which tests against the minimum version of dependencies: 
https://github.com/googleapis/gapic-generator-python/blob/main/gapic/templates/testing/constraints-3.7.txt.j2

parthea:
Towards googleapis/google-cloud-python#16292 🦕

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -341,8 +341,10 @@ jobs:
         run: |
             python -m pip install nox
       - name: Run the `unit` nox session
+        # Exclude testing for asset which requires dependency google-cloud-org-policy
+        # in order to run unit tests
+        # See https://github.com/googleapis/gapic-generator-python/issues/1806
         run: |
-          nox -f tests/integration/goldens/asset/noxfile.py -s unit-3.12
           nox -f tests/integration/goldens/credentials/noxfile.py -s unit-3.12
           nox -f tests/integration/goldens/eventarc/noxfile.py -s unit-3.12
           nox -f tests/integration/goldens/logging/noxfile.py -s unit-3.12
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yaml
```diff
@@ -328,6 +328,27 @@ jobs:
           nox -f tests/integration/goldens/eventarc/noxfile.py -s mypy-3.12 blacken lint
           nox -f tests/integration/goldens/logging/noxfile.py -s mypy-3.12 blacken lint
           nox -f tests/integration/goldens/redis/noxfile.py -s mypy-3.12 blacken lint
+  goldens-unit:
```
ohmayr, line 331: I think we were using `3.12` as the latest supported Python version. If that's the case, should we be using `3.13` here?

ohmayr (approved):

.github/workflows/tests.yaml
parthea: Sounds good. I'll update this to Python 3.13 in PR https://github.com/googleapis/gapic-generator-python/pull/2215

PR #2233, opened 2024-10-22
chore: update required checks

parthea:
The required checks are `showcase (3.12, _w_rest_async)` but they should be `showcase (3.12, showcase_w_rest_async)`

https://github.com/googleapis/gapic-generator-python/blob/d78910f7217d2d1a45662ae01365611b2f33a142/.github/sync-repo-settings.yaml#L19-L20


parthea:
Moved to https://github.com/googleapis/gapic-generator-python/pull/2215

PR #2237, opened 2024-10-25
fix: allow google-cloud-documentai 3.x

parthea:
`google-cloud-documentai` 3.0.0 will be released next week. We need to bump the upper bound in `_pypi_packages.j2` 

https://github.com/googleapis/gapic-generator-python/blob/aaef013a40b78b575edd09d27013b3cc6560a080/gapic/templates/_pypi_packages.j2#L14

https://pypi.org/project/google-cloud-documentai


holtskinner (approved):

ohmayr:
Why do we even have an upper bound if we're allowing the latest every time?

ohmayr (approved):

parthea:
> Why do we even have an upper bound if we're allowing the latest every time?

We can't guarantee that the next major version of a dependency will be compatible. By restricting the major version of dependencies, we know that what we ship won't break later when there is a breaking change in a dependency

PR #2243, opened 2024-11-07
feat: add support for metadata in unary calls - prototype 1

parthea:
Prototype 1: `ResponseMetadata` via a callback , using a conditional `with_call`

Fixes b/347351441

### DIFF
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/_client_macros.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/_client_macros.j2
@@ -31,6 +31,9 @@
             retry: OptionalRetry = gapic_v1.method.DEFAULT,
             timeout: Union[float, object] = gapic_v1.method.DEFAULT,
             {{ shared_macros.client_method_metadata_argument()|indent(8) }} = {{ shared_macros.client_method_metadata_default_value() }},
+            {% if not method.void and not method.lro and not method.client_streaming and not method.server_streaming %}
+            raw_response_callback: Optional[callable] = None,
+            {% endif %}
             {% if method.extended_lro and not full_extended_lro %}{# This is a hack to preserve backwards compatibility with the "unary" surfaces #}
             ) -> {{ method.extended_lro.operation_type.ident }}:
             {% elif not method.server_streaming %}
@@ -69,6 +72,12 @@
                 should be retried.
             timeout (float): The timeout for this request.
             {{ shared_macros.client_method_metadata_argument_doc()|indent(4) }}
+            {% if not method.void and not method.lro and not method.client_streaming and not method.server_streaming %}
+            raw_response_callback (Optional[callable]): Adds a callback that
+                will be called if the request succeeds. The callback will have
+                a `raw_response` argument which is of type
+                `Union[grpc.Call, requests.models.Response]`.
+            {% endif %}
         {% if not method.void %}
 
         Returns:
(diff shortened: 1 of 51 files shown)


noxfile.py
```diff
@@ -418,7 +419,7 @@ def showcase_alternative_templates(session):
         session,
         templates=templates,
         other_opts=("old-naming",),
-        env={"GAPIC_PYTHON_ASYNC": "False", "IGNORE_FILE": "test_universe_domain.py"},
+        env={"GAPIC_PYTHON_ASYNC": "False", "IGNORE_FILES": "test_universe_domain.py,test_request_metadata.py"},
```
parthea, line 422: This will be removed once the design is approved. I excluded ads templates during the prototype step.

gapic/templates/%namespace/%name_%version/%sub/services/%service/_client_macros.j2
```diff
@@ -141,7 +146,11 @@
 
         # Wrap the RPC method; this adds retry and timeout information,
         # and friendly error handling.
-        rpc = self._transport._wrapped_methods[self._transport.{{ method.transport_safe_name|snake_case}}]
+        if raw_response_callback:
+            call_type = "with_call"
+        else:
+            call_type = "default"
+        rpc = self._transport._wrapped_methods[self._transport.{{ method.transport_safe_name|snake_case}}][call_type]
```
daniel-sanche, line 153: FWIW, I'm still against putting a dict in `_wrapped_methods`. I think we could just make it always do `with_call`, and optionally drop the `call` argument. Let me know if I'm missing why that would cause issues

If we have to have two _wrapped_methods, I'd suggest storing it as a separate `wrapped_methods_with_call` variable, to avoid complexities around the nested dictionary lookup

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
```diff
@@ -361,7 +367,16 @@ class {{ service.async_client_name }}:
 
         # Wrap the RPC method; this adds retry and timeout information,
         # and friendly error handling.
-        rpc = self._client._transport._wrapped_methods[self._client._transport.{{ method.transport_safe_name|snake_case }}]
+        if raw_response_callback:
+            if "with_call" in inspect.signature(gapic_v1.method_async.wrap_method).parameters:
```
daniel-sanche, line 371: Is this expensive? Seems like it could add quite a bit of latency to each call

Maybe directly check the __version__?

gapic/templates/%namespace/%name_%version/%sub/services/%service/_client_macros.j2
parthea: I'll include another prototype where we have `with_call` unconditionally. I'll update this prototype to use a separate `wrapped_methods_with_call` variable

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
parthea: Good catch!

gapic/templates/%namespace/%name_%version/%sub/services/%service/_client_macros.j2
parthea: I updated this prototype to use `wrapped_methods_with_call`. I'l investigate using `with_call` unconditionally in a separate prototype

Issue #2250, opened 2024-11-15
Wrong type annotation for `metadata`, instead of `Sequence[Tuple[str, str | bytes]]`, should be `Sequence[Tuple[str, Union[str, bytes]]]`


gabrysiaolsz:
#### Environment details

  - OS type and version: Debian GNU/Linux rodete
  - Python version: 3.10.15
  - pip version: 23.0.1
  - `gapic-generator` version: N/A, bug in already generated library

#### Steps to reproduce

  1. Install `pip install google-cloud-datacatalog-lineage==0.3.9` (as a quick way of getting a generated client)
  2.  Create a generated `LineageClient`
  3.  Annotated type for `metadata` in API methods is `Sequence[Tuple[str, str]]`, but in reality it should be `Sequence[Tuple[str, Union[str, bytes]]]`. 

#### Failing Code example
Sending a pair of `str`, when the first one has a suffix of "-bin" causes errors:
```python
client.process_open_lineage_run_event(
    request,
    metadata=[
        ('x-goog-ext-512598505-bin', '\n\x08BIGQUERY\x12\x05LOAD1'),
    ]
)
```

#### Stack trace
```
E   TypeError: Binary metadata key="x-goog-ext-512598505-bin" expected bytes, got <class 'str'>

src/python/grpcio/grpc/_cython/_cygrpc/metadata.pyx.pxi:45: TypeError
```

#### Working Code example
Sending a pair of `str, bytes` succeeds:
```python
client.process_open_lineage_run_event(
    request,
    metadata=[
        ('x-goog-ext-512598505-bin', b'\n\x08BIGQUERY\x12\x05LOAD1'),
    ]
)
```

So the type annotation should be `Sequence[Tuple[str, Union[str, bytes]]]`, to avoid confusion and time wasted on debbuging. 
The [examples here](https://github.com/grpc/grpc/blob/master/examples/python/metadata/metadata_client.py) also show to send `str, bytes`.

PR #2253, opened 2024-11-19
chore: add template for metadata

parthea:
Towards https://github.com/googleapis/gapic-generator-python/issues/2250


gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -631,7 +631,7 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
         *,
         retry: OptionalRetry = gapic_v1.method.DEFAULT,
         timeout: Union[float, object] = gapic_v1.method.DEFAULT,
-        metadata: Sequence[Tuple[str, str]] = (),
+        {{ shared_macros.client_method_metadata_argument()|indent(8) }} = {{ shared_macros.client_method_metadata_default_value() }},
```
ohmayr, line 634: I think these macros are no longer used and are adding a toil to whenever a change needs to be made to the client methods i.e. see the linked comment above.

I'd suggest to get rid of these macros.

ohmayr (approved):

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
parthea: I filed https://github.com/googleapis/gapic-generator-python/issues/2255 for this clean up step. We still need to remove usage here: https://github.com/search?q=repo%3Agoogleapis%2Fgoogleapis%20add-iam-methods&type=code

Issue #2191, opened 2024-09-23
docs: Add documentation for request / response interceptors for `gRPC` transport.


ohmayr:
request / response interceptors are currently not supported for `gRPC`. We need to add support for this and also update the relevant unit tests to ensure that we have code coverage.

parthea:
I've switched this issue from `bug` to `docs` as interceptors are a feature of gRPC and are already implemented. We need to add the relevant tests and documentation.

parthea:
This should be part of product agnostic documentation (b/381043501)

PR #2256, opened 2024-11-19
fix: update return annotation for streaming calls

parthea:
Towards b/347351441

 `GrpcStream[Response]` exposes the underlying `grpc.Call` for streaming calls, there is currently no way to use these `grpc.Call` methods without mypy errors, because of the restrictive return type annotation used. See https://github.com/googleapis/python-api-core/pull/554#issue-1998080923 for context.

Adding the `GrpcStream[Response]` and `Awaitable[AsyncIterable[Response]]` response types for streaming calls will satisfy type checkers. 

BEGIN_COMMIT_OVERRIDE
fix(deps): require google-api-core >= 2.17.1
END_COMMIT_OVERRIDE

`google-api-core>=2.15.0` is the minimum version which supports types `GrpcAsyncStream` and `GrpcStream`
`google-api-core>=2.17.1` is needed which fixes an issue with REST Streaming which can be seen in this [build log](https://github.com/googleapis/gapic-generator-python/actions/runs/11922672625/job/33229457759)

### DIFF
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/_client_macros.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/_client_macros.j2
@@ -36,7 +36,7 @@
             {% elif not method.server_streaming %}
             ) -> {{ method.client_output.ident }}:
             {% else %}
-            ) -> Iterable[{{ method.client_output.ident }}]:
+            ) -> Union[Iterable[{{ method.client_output.ident }}], GrpcStream[{{ method.client_output.ident }}]]:
             {% endif %}
         r"""{{ method.meta.doc|rst(width=72, indent=8)|trim }}
 
@@ -75,7 +75,7 @@
             {% if not method.server_streaming %}
             {{ method.client_output.ident.sphinx }}:
             {% else %}
-            Iterable[{{ method.client_output.ident.sphinx }}]:
+            Union[Iterable[{{ method.client_output.ident }}], GrpcStream[{{ method.client_output.ident }}]]:
             {% endif %}
                 {{ method.client_output.meta.doc|rst(width=72, indent=16, source_format="rst") }}
         {% endif %}
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
@@ -17,6 +17,13 @@ import warnings
 {% set package_path = api.naming.module_namespace|join('.') + "." + api.naming.versioned_module_name %}
 from {{package_path}} import gapic_version as package_version
 
+{% set import_ns = namespace(has_streaming_methods=false) %}
+{%- for method in service.methods.values() if method.server_streaming %}
+{% set import_ns.has_streaming_methods = True %}
+{% endfor -%}
+{% if import_ns.has_streaming_methods == True %}
+from google.api_core.grpc_helpers_async import GrpcAsyncStream
+{% endif %}
 from google.api_core.client_options import ClientOptions
 from google.api_core import exceptions as core_exceptions
 from google.api_core import gapic_v1
@@ -263,7 +270,7 @@ class {{ service.async_client_name }}:
             {% if not method.server_streaming %}
             ) -> {{ method.client_output_async.ident }}:
             {% else %}
-            ) -> Awaitable[AsyncIterable[{{ method.client_output_async.ident }}]]:
+            ) -> Union[Awaitable[AsyncIterable[{{ method.client_output_async.ident }}]], Awaitable[GrpcAsyncStream[{{ method.client_output_async.ident }}]]]:
             {% endif %}
         r"""{{ method.meta.doc|rst(width=72, indent=8)|trim }}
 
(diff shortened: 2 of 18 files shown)


gapic/templates/%namespace/%name_%version/%sub/services/%service/_client_macros.j2
```diff
@@ -36,7 +36,7 @@
             {% elif not method.server_streaming %}
             ) -> {{ method.client_output.ident }}:
             {% else %}
-            ) -> Iterable[{{ method.client_output.ident }}]:
+            ) -> Union[Iterable[{{ method.client_output.ident }}], GrpcStream[{{ method.client_output.ident }}]]:
```
daniel-sanche, line 39: This should be redundant, since `GrpcStream` is [just an alias](https://github.com/googleapis/python-api-core/blob/e298ebdfbbb839d64c2da570b454c6c0f6c4e00c/google/api_core/grpc_helpers.py#L149) for `_StreamingResponseIterator`, which is [already Iterable](https://github.com/googleapis/python-api-core/blob/e298ebdfbbb839d64c2da570b454c6c0f6c4e00c/google/api_core/grpc_helpers.py#L101). So we should just be able to use `GrpcStream` here

returning a union type can be ugly, since any attempt to use the returned instance will cause typing issues, since mypy won't know which of the types it is

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
```diff
@@ -263,7 +270,7 @@ class {{ service.async_client_name }}:
             {% if not method.server_streaming %}
             ) -> {{ method.client_output_async.ident }}:
             {% else %}
-            ) -> Awaitable[AsyncIterable[{{ method.client_output_async.ident }}]]:
+            ) -> Union[Awaitable[AsyncIterable[{{ method.client_output_async.ident }}]], Awaitable[GrpcAsyncStream[{{ method.client_output_async.ident }}]]]:
```
daniel-sanche, line 273: same comment here: it would be great if we could avoid the Union return

daniel-sanche (changes requested):

gapic/templates/%namespace/%name_%version/%sub/services/%service/_client_macros.j2
daniel-sanche, line 39: Did you find a solution for how to deal with `api_core.retry.Retry` in this?

The retry object wraps the rpc iterator, and it doesn't currently expose `grc.Call` methods, so anything involving retires won't return a GrpcStream. It's not exactly clear to me how we should expose metadata at the retry level (do we want to give metadata for each failed call, or only the final successful one? I think each failed request could be needed?)

I don't think this is relevant for bigtable, since we extend the retry logic. But it's probably needed in other contexts

parthea: For REST client, `GrpcStream` would be the response type. My initial thought is that we a type which doesn't have `Grpc` in the name. Perhaps we could use a new `StreamResponse` instead of `GrpcStream`?

parthea: Given that this response type is also used for REST streaming, I'm not sure that we can avoid the union here. `_StreamingResponseIterator` has a dependency on gRPC. We would need to update `_StreamingResponseIterator` if we want to use it for REST

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
parthea: I don't see any way to avoid this as this is also the response type for REST. Let's decide on the next steps for https://github.com/googleapis/gapic-generator-python/pull/2256#discussion_r1851135593 and the same solution could be applied to async.

gapic/templates/%namespace/%name_%version/%sub/services/%service/_client_macros.j2
parthea: We could consider adding this as future work, given that it's more than just changing the response type.

parthea: As an alternative, we could add a callback similar to what we have for unary requests in https://github.com/googleapis/gapic-generator-python/pull/2243 as a callback could be used to solve the feature request in https://github.com/googleapis/gapic-generator-python/pull/2256#discussion_r1851154197

daniel-sanche: That's a good point, I wasn't considering the rest side. We could make a `StreamResponse` as a mixin class, that we could add to both `_StreamingResponseIterator` and `rest_streaming.ResponseIterator`?

Yeah, using a callback in all cases could be a good way to resolve some of this complexity. It seems like we'd have some of the same typing issues there though, where we still need a way to represent results for either grpc or rest, right? I might have to dig deeper into the rest side

daniel-sanche: Ok, sounds good

FWIW, I think it would be a good idea to still design a solution for failed attempts, so we don't paint ourselves into a corner here. (maybe we can use the retry.on_error callback for this?) But yeah, I'll assume we just want one set of metadata for now

parthea:
Closing as obsolete. The proposed solution in b/347351441 is to use gRPC/REST interceptors. https://github.com/googleapis/gapic-generator-python/issues/2191 tracks improving documentation to clarify how gRPC/REST interceptors can be used to obtain metadata.

Issue #2015, opened 2024-04-24
`license` should not be a reserved field name


Ark-kun:
I've just noticed that the generated classes mangle the `license` field name:
```py
class Citation(proto.Message):
    r"""Source attributions for content."""
    ...
    license_: str = proto.Field(
        proto.STRING,
        number=5,
    )
```

https://github.com/googleapis/python-aiplatform/blob/c56dd50310ecc42d6a3b0cd2f30db71e8cad0cf6/google/cloud/aiplatform_v1/types/content.py#L572

I think this is incorrect. `license` is not a Python reserved keyword and it can definitely be used as an attribute name.

parthea:
It exists as a built-in constant:  https://docs.python.org/3/library/constants.html#constants-added-by-the-site-module

```
>>> license
Type license() to see the full license text
```

```
>>> license()
A. HISTORY OF THE SOFTWARE
==========================
```

parthea:
I'm going to close this issue but please feel free to open a new issue with more information.

Ark-kun:
>It exists as a built-in constant

This does not prevent it from being used as a field name.

parthea:
A decision was made to

parthea:
I the example from https://github.com/googleapis/gapic-generator-python/issues/2015#issue-2260556390

```
class Citation(proto.Message):
    r"""Source attributions for content."""
    ...
    license: str = proto.Field(
        proto.STRING,
        number=5,
    )
```
We would be clobbering the built-in constant `license`.  It can be done technically but It is generally bad practice. 

Nonetheless, I'll keep this open as a feature request so we can evaluate when we are considering the next breaking change.

(Regardless of whether it can be done technically, it would be a breaking change to change the reserved words list. )

PR #2265, opened 2024-11-27
feat: add debug log when creating client

ohmayr:
Fixes b/381099246


gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
```diff
@@ -243,6 +251,9 @@ class {{ service.async_client_name }}:
 
         )
 
+        if CLIENT_LOGGING_SUPPORTED:
```
vchudnov-g, line 254: I don't think we should have an extra line at each logging site to check the logging enablement. Rather, I suggest something like what I propose in prototype R-0 in in [go/cloudsdk-python-logging-performant](http://goto.google.com/cloudsdk-python-logging-performant), to abstract away any checking we may want to do while providing an interface (only used by us, not our users) that is parallel to the standard loggers.

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
```diff
@@ -243,6 +251,9 @@ class {{ service.async_client_name }}:
...
+            _LOGGER.debug(" {{ service.async_client_name }} client is successfully constructed.")
```
parthea, line 255: This log also appears when running async code. Is that expected?

```
DEBUG    google.showcase_v1beta1.services.echo.client:client.py:605  EchoClient client is successfully constructed.
DEBUG    google.showcase_v1beta1.services.echo.async_client:async_client.py:258  EchoAsyncClient client is successfully constructed.
```

parthea: The ruby logs have the package name in the logs, whereas we have the module name. Is there a benefit to including the PyPI package name as well?

```
google-cloud-secret_manager-v1: Created client for google.cloud.secretmanager.v1.SecretManagerService 
```

parthea, line 255: The `ruby` logs says `Created client for google.cloud.secretmanager.v1.SecretManagerService` which matches the proto: https://github.com/googleapis/googleapis/blob/072ae46e493035d1419caad40bccc1a4f1b977bf/google/cloud/secretmanager/v1/service.proto#L46

The equivalent log for showcase would be `Created client for google.showcase.v1beta1.Echo` instead of what we currently have `EchoClient client is successfully constructed.` 
https://github.com/googleapis/gapic-showcase/blob/625d292c5574c703f42bf50f7ea7b92d47f4918e/schema/google/showcase/v1beta1/echo.proto#L42

I would prefer `Created client for google.showcase.v1beta1.Echo`. WDYT?

parthea, line 255: The ruby logs include information about which credentials will be used with the client which should be included here as well.
```
I, [2024-09-30T16:52:15.809067 #1917313]  INFO -- google-cloud-secret_manager-v1: Created client for google.cloud.secretmanager.v1.SecretManagerService -- {"clientId":3220,"system":"google-cloud-secret_manager-v1","serviceName":"google.cloud.secretmanager.v1.SecretManagerService","credentialsId":3260,"credentialsType":"Google::Auth::ServiceAccountCredentials","credentialsScope":["https://www.googleapis.com/auth/cloud-platform"],"useSelfSignedJWT":false,"universeDomain":"googleapis.com"}
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -568,6 +581,8 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
                     host=self._api_endpoint,
                     client_info=client_info,
                 )
+                if CLIENT_LOGGING_SUPPORTED:
+                    _LOGGER.debug(" {{ service.client_name }} client is successfully constructed.")
```
parthea, line 585: Remove this log since this is under the `if "rest_asyncio" in str(transport_init):` code block and there is a separate log for the async client.

parthea: Fixed in https://github.com/googleapis/gapic-generator-python/pull/2265/commits/c4696b3f307703395ff64582be9b689c632be10f

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
parthea: > The log for the sync client appears when running async code. Is that expected?

This was fixed in https://github.com/googleapis/gapic-generator-python/pull/2265/commits/6f2bbee80037d39c0ce8fefe0b59b3b14875cd8f .

Disregard my comment about the package name. The module name provide sufficient information

parthea: I added logging for `serviceName` via https://github.com/googleapis/gapic-generator-python/pull/2265/commits/a7158546cc47eb8b78174213732733a75c008c38 but it only appears when using structured logging

```
{"timestamp": "2024-11-28 19:30:07,202", "severity": "DEBUG", "name": "google.showcase_v1beta1.services.echo.client", "message": "Created client `google.showcase_v1beta1.EchoClient`.", "serviceName": "google.showcase.v1beta1.Echo"}
```

```
DEBUG:google.showcase_v1beta1.services.echo.client:Created client `google.showcase_v1beta1.EchoClient`.
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
```diff
@@ -243,6 +251,24 @@ class {{ service.async_client_name }}:
...
+            credential_info = None
+            # TODO: Remove this condition once the minimum version of google-auth is 2.35.0
+            if hasattr(self.transport._credentials, "get_cred_info"):
+                credential_info = self.transport._credentials.get_cred_info()
+
+            _LOGGER.debug(
+                "Created client `{{ service.meta.address.proto_package_versioned }}.{{ service.async_client_name }}`.",
+                extra = {
+                    "serviceName": "{{ service.meta.address.proto }}",
+                    "universeDomain": getattr(self._transport._credentials, "universe_domain", ""),
```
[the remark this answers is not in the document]
parthea: `google.auth.credentials.get_cred_info()` is the only thing that could potentially contain sensitive information but reading the description of this function, it is intended to be included in logs.
https://github.com/googleapis/google-auth-library-python/blob/bfab59288fb5e4a9d86f5fff55307a72a0b86e47/google/auth/credentials.py#L131

```
The credential information will be added to auth related error messages by client library.
```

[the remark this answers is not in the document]
vchudnov-g: We should still check what that "credential information" is.

As to the other fields, yeah, I would tend to agree they should be safe. (In general, I'm worried about metadata that, while not strictly secret, could narrow the attack surface for adversaries. But looking into the semantics of those fields, that seems irrelevant here.)

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -510,6 +519,10 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
         # Initialize the universe domain validation.
         self._is_universe_domain_valid = False
 
+        if CLIENT_LOGGING_SUPPORTED:  # pragma: NO COVER
```
vchudnov-g, line 522: Here and elsewhere we log:

```suggestion
        if CLIENT_LOGGING_SUPPORTED and _LOGGER.isEnabledFor(logging.DEBUG)::  # pragma: NO COVER
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
```diff
@@ -243,6 +251,24 @@ class {{ service.async_client_name }}:
 
         )
 
+        if CLIENT_LOGGING_SUPPORTED and _LOGGER.isEnabledFor(std_logging.DEBUG):  # pragma: NO COVER
+            # TODO: Make this performant when logging is not enabled
```
parthea, line 255:
```suggestion
```

Remove this TODO comment as it was from before we added the guard for `_LOGGER.isEnabledFor(std_logging.DEBUG)`

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
```diff
@@ -243,6 +251,24 @@ class {{ service.async_client_name }}:
...
+
+            # TODO: Remove this condition once the minimum version of google-auth is 2.35.0
```
parthea, line 257: File a bug for the TODO comment

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
```diff
@@ -243,6 +251,24 @@ class {{ service.async_client_name }}:
...
+            if callable(credential_info):
+                credential_info = credential_info()
+
+            _LOGGER.debug(
+                "Created client `{{ service.meta.address.proto_package_versioned }}.{{ service.async_client_name }}`.",
+                extra = {
+                    "serviceName": "{{ service.meta.address.proto }}",
+                    "universeDomain": getattr(self._client._transport._credentials, "universe_domain", ""),
+                    "credentialType": f"{type(self._client._transport._credentials).__module__}.{type(self._client._transport._credentials).__qualname__}",
+                    "credentialInfo": credential_info,
```
vchudnov-g, line 268: Does `credential_info` contain actual secrets? In that case, I would use as the value here `f"[hash:{hash(credential_info}]"` or probably even better, `f"[hash:{hashlib.sha256(credential_info}]"`

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
```diff
@@ -4,6 +4,7 @@
 {% import "%namespace/%name_%version/%sub/services/%service/_client_macros.j2" as macros %}
 {% import "%namespace/%name_%version/%sub/services/%service/_shared_macros.j2" as shared_macros %}
 
+import logging as std_logging
```
vchudnov-g, line 7: I'm curious: why are you renaming the standard import? We could simply use non-conflicting names for our own constructs/modules.

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
```diff
@@ -596,6 +609,24 @@ class {{ service.client_name }}(metaclass={{ service.client_name }}Meta):
...
+                if callable(credential_info):
+                    credential_info = credential_info()
+
+                _LOGGER.debug(
+                    "Created client `{{ service.meta.address.proto_package_versioned }}.{{ service.client_name }}`.",
+                    extra = {
+                        "serviceName": "{{ service.meta.address.proto }}",
+                        "universeDomain": getattr(self._transport._credentials, "universe_domain", ""),
+                        "credentialType": f"{type(self._transport._credentials).__module__}.{type(self._transport._credentials).__qualname__}",
+                        "credentialInfo": credential_info,
```
vchudnov-g, line 627: Ditto on hashing this value.

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
ohmayr: There was a naming conflict in the logging goldens:

```
import logging
from google.cloud.logging_v2.types import logging
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
ohmayr: `credentials_info` was added here: https://github.com/googleapis/google-auth-library-python/commit/6f75dd5de9ee1da4509306ff2e6420b3d88f9d00.

It looks like it can possibly contain a source (path), type, and principal (email). So we're okay with not hashing it?

gapic/templates/%namespace/%name_%version/%sub/services/%service/client.py.j2
ohmayr: See my other comment.

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
ohmayr: I've removed the condition so a bug is no longer needed.

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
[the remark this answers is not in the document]
ohmayr: Closing this comment since we have a separate thread to discuss.

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
ohmayr: Extra fields will only be included in the structured logging format. We can support this as a follow up: https://github.com/googleapis/gapic-generator-python/issues/2268.

parthea (approved):

gapic/templates/%namespace/%name_%version/%sub/services/%service/async_client.py.j2
parthea: I files https://github.com/googleapis/gapic-generator-python/issues/2269 for the long term fix which is to avoid generating modules, [like this](https://github.com/googleapis/gapic-generator-python/blob/main/tests/integration/goldens/logging/google/cloud/logging_v2/types/logging.py), which conflict with [Python3 built-in modules](https://docs.python.org/3/py-modindex.html)

parthea: One alternative would be to add `logging` to the reserved names list [here](https://github.com/googleapis/gapic-generator-python/blob/d87480974bf230647e77343f30b1c82a08773a03/gapic/utils/reserved_names.py#L32).

However doing so would result in a breaking change in other libraries such as `google-cloud-build` where field with the name `logging` exists.
https://github.com/googleapis/google-cloud-python/blob/adea041c1d43b94d11ff9d4d5ba13c0d7898cac2/packages/google-cloud-build/google/cloud/devtools/cloudbuild_v1/types/cloudbuild.py#L3368

The field `logging` would be renamed to `logging_`, similar to how `license` was renamed to `license_`
https://github.com/googleapis/gapic-generator-python/issues/2015

The reason is that the reserved terms list is tightly coupled with the logic to alias import statements, and avoid collisions in field names. IOW, we don't have separate lists for just adding an alias to import statements.

https://github.com/googleapis/gapic-generator-python/blob/d87480974bf230647e77343f30b1c82a08773a03/gapic/schema/api.py#L175-L181

https://github.com/googleapis/gapic-generator-python/blob/d87480974bf230647e77343f30b1c82a08773a03/gapic/schema/wrappers.py#L78-L82

PR #2271, opened 2024-12-03
feat: add debug log when sending requests via gRPC

parthea:
Towards b/381100907

The changes in this PR are only for unary-unary calls. unary-stream calls still need to be added.

### DIFF
--- a/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
+++ b/gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
@@ -55,47 +55,46 @@ _LOGGER = logging.getLogger(__name__)
 class _LoggingClientInterceptor(grpc.UnaryUnaryClientInterceptor):  # pragma: NO COVER
     def intercept_unary_unary(self, continuation, client_call_details, request):
         request_metadata = client_call_details.metadata
-        if CLIENT_LOGGING_SUPPORTED and _LOGGER.isEnabledFor(logging.DEBUG):
-            request_payload = {% if method.output.ident.is_proto_plus_type %}{type(request).to_json(request){% else %}json_format.MessageToJson(request){% endif %}
-            {% else %}
-            grpc_request = {
-              "payload":   request_payload,
-              "requestMethod": "grpc",
-              "metadata": dict(request_metadata),
-            }
-            _LOGGER.debug(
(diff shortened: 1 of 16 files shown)


gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
```diff
@@ -42,6 +43,58 @@ from google.longrunning import operations_pb2 # type: ignore
...
+        response = continuation(client_call_details, request)
+        response_metadata = response.trailing_metadata()
+        # Convert gRPC metadata `<class 'grpc.aio._metadata.Metadata'>` to list of tuples
+        metadata = [(k, v) for k, v in response_metadata]
+        result = response.result()
+        if CLIENT_LOGGING_SUPPORTED and _LOGGER.isEnabledFor(logging.DEBUG):
+            grpc_response = {
+                "payload": type(result).to_json(result),
+                "metadata": dict(metadata),
+                "status": "OK",
```
vchudnov-g, line 84: IIUC, if we get here the status is `OK`. In case of a gRPC error, an exception is thrown...or do we catch in the lower levels? If so, how do we get the status?

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
```diff
@@ -56,13 +56,19 @@ class MetadataClientInterceptor(grpc.UnaryUnaryClientInterceptor):
     def intercept_unary_unary(self, continuation, client_call_details, request):
         request_metadata = client_call_details.metadata
         if CLIENT_LOGGING_SUPPORTED and _LOGGER.isEnabledFor(logging.DEBUG):
+            grpc_request = {
+              "payload":   type(request).to_json(request),
+              "requestMethod": "grpc",
```
ohmayr, line 61: I don't think we need this for `gRPC`. `method` and `url` are rest only fields.

vchudnov-g: But how will the user looking at the logs know (from the logs only) that this was a gRPC request? They have to reason from not having a URL or method field, or from looking at the metadata. We might as well make it explicit for convenience.

No need to block on this; we can change it later.

vchudnov-g:
As per internal discussions, we'll be splitting this PR into smaller PRs

PR #2272, opened 2024-12-04
feat: Add support for reading selective GAPIC generation methods from service YAML

gkevinzheng:
This PR implements Selective GAPIC generation for Python. For more details, see:

Overall design doc: [go/selective-gapic-generation](http://go/selective-gapic-generation)
Implementation doc for Python: [go/selective-gapic-gen-python](http://go/selective-gapic-gen-python)


parthea (commented):
LGTM but holding off on formal approval until tomorrow after the next release is shipped

parthea (approved):

parthea:
Requesting review from @vchudnov-g as well

gapic/schema/wrappers.py
```diff
@@ -1679,6 +1779,51 @@ def with_context(self, *,
...
+        address_allowlist.add(self.ident)
+
+        if self.lro:
+            self.lro.build_address_allowlist_for_selective_gapic(address_allowlist=address_allowlist,
+                                                                 resource_messages=resource_messages)
+
+        if self.extended_lro:
+            # We need to add the service/method pointed to by self.operation_service to
+            # the allowlist, as it might not have been specified by selective_gapic_generation.
+            # We assume that the operation service lives in the same proto as this one.
```
vchudnov-g, line 1802: Is this a good assumption? (I honestly don't know.) Will it work with operation mixins?

gapic/schema/wrappers.py
```diff
@@ -1994,3 +2139,43 @@ def with_context(self, *,
...
+        """Builds a set of Addresses of wrapper objects to be included in selective GAPIC generation.
+
+        This method is used to create an allowlist of addresses to be used to filter out unneeded
+        services, methods, messages, and enums at a later step.
+        """
+
+        for method in self.methods.values():
+            if method.ident.proto in method_names:
+                # Include this service if there are any types/methods in selective gapic for this service.
+                address_allowlist.add(self.meta.address)
```
vchudnov-g, line 2158: Isn't this allow-listing this service it it has any methods at all, not just allowlisted ones?

Should the logic be something like this pseudocode?
```python
for method in self.methods.values():
  if method in allowlist:
    allowlist.add(self)
    allowlist.add(method)
    method.build_allowlist()

File: gapic/schema/wrappers.py
Comment (vchudnov-g) [inline gapic/schema/wrappers.py:1]: Looking at this file, it seems to me that the `build_address_allowlist` name is descriptive for the method you're adding to the `Service` class, but the other classes should have their new methods names something like `add_to_address_allowlist`, to emphasize that they themselves are not checking for the RPC allowlist in the YAML, but are being included transitively. WDYT?

**Review by vchudnov-g — COMMENTED** [review]
I think this is looking good, but I have some questions/suggestions. PTAL.

File: gapic/schema/wrappers.py (hunk 1 above)
Comment (gkevinzheng) [reply gapic/schema/wrappers.py:1802]: I have to make this assumption, otherwise the lookups won't work. The precedent for this is here: https://github.com/googleapis/gapic-generator-python/blob/e050f4eb3eddfe150f028a2a2bd901899afb965a/gapic/schema/api.py#L1238-L1250

File: gapic/schema/wrappers.py (hunk 2 above)
Comment (gkevinzheng) [reply gapic/schema/wrappers.py:2158]: `method_names` is the allowlist. I'll make that clearer by renaming it to `method_allowlist` or `method_names_allowlist`.

File: gapic/schema/api.py [hunk 3]
```diff
@@ -365,10 +428,49 @@ def disambiguate_keyword_sanitize_fname(
...
+                for proto in api.protos.values():
+                    proto.build_address_allowlist_for_selective_gapic(method_names=selective_gapic_methods,
+                                                                      address_allowlist=address_allowlist,
+                                                                      resource_messages=all_resource_messages)
+
+                new_protos = {}
+
+                # We only prune services/messages/enums from protos that are not depepdencies.
+                for name, proto in api.all_protos.items():
+                    if name not in api.protos:
```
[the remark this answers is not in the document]
gkevinzheng: Definition here:
https://github.com/googleapis/gapic-generator-python/blob/e050f4eb3eddfe150f028a2a2bd901899afb965a/gapic/schema/api.py#L406-L417

Here I'm trying to prune only the proto objects corresponding to the files in this API package.

gapic/schema/api.py
```diff
@@ -237,6 +237,69 @@ def disambiguate(self, string: str) -> str:
...
+        """
+        services = {
+            k: v.prune_messages_for_selective_gapic(
+                address_allowlist=address_allowlist)
+            for k, v in self.services.items()
+            if v.meta.address in address_allowlist
+        }
+
+        # For messages and enums we should only be pruning them from all_messages if they
+        # are proto plus types. This should apply to the Protos we are pruning from, but might
```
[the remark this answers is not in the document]
gkevinzheng: Are we trying to move away from proto plus types? I'll remove the "in the future part" regardless

vchudnov-g:
Meta-comment: please do not merge `main` while a PR is being reviewed; it makes it much harder to see the incremental changes relevant only to the PR. I suggest merging `main` before sending out for the initial review, and after the PR is approved when you're ready to commit the PR to `main` (and if the merge changes are non-trivial, a second round of review could be reasonable, but that seems to be rare).

tests/integration/redis_v1.yaml
vchudnov-g, line 1: Would it be possible to have this integration test not clobber the existing redis test? We have been using redis as the only place to do a golden test for a bunch of things, and I want to ensure we don't accidentally exclude something useful, or and that we're not puzzled when a new feature does not show up. So could we rename the source and the output of this test to be something like `redis_selective_v1`? (I like the idea of the source proto/config being subsets of their full redis counterparts, because that way we can easily diff the two.)

gapic/schema/api.py
```diff
@@ -237,6 +237,68 @@ def disambiguate(self, string: str) -> str:
...
+        This method is used to create an allowlist of addresses to be used to filter out unneeded
+        services, methods, messages, and enums at a later step.
+        """
+        for service in self.services.values():
+            # The method.operation_service for an extended LRO is not fully qualified, so we
+            # truncate the service names accordingly so they can be found in
+            # method.add_to_address_allowlist
+            services_in_proto = {
+                service.name: service for service in self.services.values()
+            }
```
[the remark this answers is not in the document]
gkevinzheng: You're right, this should be outside the for loop, as it never changes and is used as a reference for the `service.add_to_address_allowlist` method.

gapic/schema/api.py
```diff
@@ -237,6 +237,68 @@ def disambiguate(self, string: str) -> str:
...
+        return None.
+        """
+        services = {
+            k: v.prune_messages_for_selective_generation(
+                address_allowlist=address_allowlist)
+            for k, v in self.services.items()
+            if v.meta.address in address_allowlist
+        }
+
+        # For messages and enums we should only be pruning them from all_messages if they
```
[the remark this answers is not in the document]
gkevinzheng: Does filtering out non-protoplus messages in the schema prevent them from being generated? For Selective GAPIC Phase 1, we should be ignoring things that get generated by protoc. I think that other languages use protoc generated types, but since we generate our own types, we need to filter things out.

Do I need the `is_proto_plus` check in that case?

gapic/schema/wrappers.py
```diff
@@ -1994,3 +2140,43 @@ def with_context(self, *,
...
+    def prune_messages_for_selective_generation(self, *,
+                                                address_allowlist: Set['metadata.Address']) -> 'Service':
+        """Returns a truncated version of this Service.
+
+        Only the methods, messages, and enums contained in the address allowlist
+        are included in the returned object.
+        """
+        # Since the API/Proto level dataclass attributes only go down to the method
+        # level from the top level services, and not any deeper in the schema tree
+        # (all_messages and all_enums are different objects), we can stop here.
```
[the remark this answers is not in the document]
gkevinzheng: I reread my comment and I think what I mean is that all the pruning that needs to happen happens in 3 places:

1. At the `Proto` level when we filter out `Proto.all_messages` and `Proto.all_enums`
2. At the `Service` level when we filter out the methods.

These 2 steps should be sufficient to truncate the schema down to what we need.

gapic/schema/api.py
```diff
@@ -237,6 +237,68 @@ def disambiguate(self, string: str) -> str:
             return self.disambiguate(f'_{string}')
         return string
 
+    def add_to_address_allowlist(self, *,
+                                 method_allowlist: Iterable[str],
+                                 address_allowlist: Set['metadata.Address'],
+                                 resource_messages: Dict[str, 'wrappers.MessageType'],
+                                 ) -> None:
```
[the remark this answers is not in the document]
gkevinzheng: 1.b holds. I'm passing around `address_allowlist` because it seems simpler/more efficient to extend one set that gets modified than to initialize a new set, add to it via set union of other sets that get returned in recursive function calls, then return that set, every stack frame.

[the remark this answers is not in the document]
vchudnov-g: OK, that makes sense. That seems simpler than passing to the various methods in various classes a reference to a singleton they all modify. Thanks!

gapic/schema/api.py
```diff
@@ -237,6 +237,68 @@ def disambiguate(self, string: str) -> str:
...
+        """
+        services = {
+            k: v.prune_messages_for_selective_generation(
+                address_allowlist=address_allowlist)
+            for k, v in self.services.items()
+            if v.meta.address in address_allowlist
+        }
+
+        # For messages and enums we should only be pruning them from all_messages if they
+        # are proto plus types. This should apply to the Protos we are pruning from.
```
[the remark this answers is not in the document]
vchudnov-g: OK, the new explanation is much clearer. Thanks!

gapic/schema/api.py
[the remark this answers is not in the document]
vchudnov-g: > For Selective GAPIC Phase 1, we should be ignoring things that get generated by protoc. 

If by "ignoring" you mean that we are not excluding protoc generated types even if they're not used by the selective generation, then the code you have below satisfies that intent, but I would clarify it in comments ("we only exclude proto-plus types not in the allow-list. We do not exclude any protoc-generated types.")

Now, it seems to me that if we're not referencing a proto-plus type because it's not in the allowlist, then its corresponding protoc type would also not be referenced so, semantically, it ought to be safe to remove it. But I do wonder how complicated/needed it is to remove some of the protoc-generated objects. Is this why this is relegated to a later phase of the project? One would think that if the point of selective generation is to not expose some proto objects to in the libraries, (eventually) removing the non-allowlisted protoc-generated types would be necessary.

tests/unit/schema/test_api.py
```diff
@@ -2745,11 +2745,813 @@ def test_read_empty_python_settings_from_service_yaml():
...
+                make_field_pb2(name="thing", number=1, type=9),
+            ),
+        )
+        response_message = make_message_pb2(name=response_message_name)
+
+        # Set up the resource
+        resource_message_opts = resource_message.options.Extensions[resource_pb2.resource]
+        resource_message_opts.type = resource_type
+        resource_message_opts.pattern.append(
+            "foo/{foo}/bar/{bar}")
```
vchudnov-g, line 3386: Am I correct in thinking these `foo` and `bar` are not required to be the same as the various `Foo` and `Bar` messages and fields you declare? If so, use some other placeholders, like library terms or animal terms.

tests/integration/redis_v1.yaml
vchudnov-g: Thanks for doing this!

tests/integration/goldens/redis_selective/.flake8
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2020 Google LLC
```
vchudnov-g, line 3:
```suggestion
# Copyright 2025 Google LLC
```

vchudnov-g (approved):
This looks great! Please address open comments from previous reviews before you merge. Thank you so much for doing this!

tests/integration/goldens/redis_selective/.flake8
gkevinzheng: I tried to edit the year, but the golden test for `redis_selective` fails on it expecting the year 2020:

```
$ cat 
Executing tests from //tests/integration:redis_selective
-----------------------------------------------------------------------------
diff -r codegen_tmp/.flake8 
3c3
< # Copyright 2020 Google LLC
---
> # Copyright 2025 Google LLC
```

PR #2284, opened 2024-12-10
feat: add client logging support for sync gRPC

ohmayr:
excludes support for mixins and streaming calls.


gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
```diff
@@ -42,6 +44,66 @@ from google.longrunning import operations_pb2 # type: ignore
...
+                "requestMethod": "grpc",
+                "metadata": dict(request_metadata),
+            }
+            _LOGGER.debug(
+                f"Sending request for {client_call_details.method}",
+                extra = {
+                    "serviceName": "{{ service.meta.address.proto }}",
+                    "rpcName": client_call_details.method,
+                    "request": grpc_request,
+                    {# TODO(https://github.com/googleapis/gapic-generator-python/issues/2275): logging `metadata` seems repetitive and may need to be cleaned up. We're including it within "request" for consistency with REST transport.' #}
```
vchudnov-g, line 75:
```suggestion
                    {# TODO(https://github.com/googleapis/gapic-generator-python/issues/2275): logging `metadata` seems repetitive and may need to be cleaned up. We're including it within "request" for consistency with REST transport. #}
```

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
```diff
@@ -42,6 +44,66 @@ from google.longrunning import operations_pb2 # type: ignore
...
+                "metadata": metadata,
+                "status": "OK", 
+            }
+            _LOGGER.debug(
+                f"Received response for {client_call_details.method}.",
+                extra = {
+                    "serviceName": "{{ service.meta.address.proto }}",
+                    "rpcName": client_call_details.method,
+                    "response": grpc_response,
+                    {# TODO(https://github.com/googleapis/gapic-generator-python/issues/2275): logging `metadata` seems repetitive and may need to be cleaned up. We're including it within "request" for consistency with REST transport.' #}
```
vchudnov-g, line 101:
```suggestion
                    {# TODO(https://github.com/googleapis/gapic-generator-python/issues/2275): logging `metadata` seems repetitive and may need to be cleaned up. We're including it within "request" for consistency with REST transport. #}
```

tests/integration/goldens/credentials/google/iam/credentials_v1/services/iam_credentials/transports/grpc.py
```diff
@@ -21,12 +22,71 @@
...
+                    "metadata": grpc_request["metadata"],
+                },
+            )
+
+        response = continuation(client_call_details, request)
+        response_metadata = response.trailing_metadata()
+        # Convert gRPC metadata `<class 'grpc.aio._metadata.Metadata'>` to list of tuples
+        metadata = dict([(k, v) for k, v in response_metadata]) if response_metadata else None
+        result = response.result()
+        if CLIENT_LOGGING_SUPPORTED and _LOGGER.isEnabledFor(std_logging.DEBUG):  # pragma: NO COVER
```
vchudnov-g, line 69: I think you can move most of these lines inside the logging guard, so only the `continuation` line stays outside, since all we return is the response.

```suggestion
        if CLIENT_LOGGING_SUPPORTED and _LOGGER.isEnabledFor(std_logging.DEBUG):  # pragma: NO COVER
                response_metadata = response.trailing_metadata()
                # Convert gRPC metadata `<class 'grpc.aio._metadata.Metadata'>` to list of tuples
               metadata = dict([(k, v) for k, v in response_metadata]) if response_metadata else None
               result = response.result()
```

tests/integration/goldens/credentials/google/iam/credentials_v1/services/iam_credentials/transports/grpc.py
```diff
@@ -21,12 +22,71 @@
...
+    CLIENT_LOGGING_SUPPORTED = False
+
+_LOGGER = std_logging.getLogger(__name__)
+
+
+class _LoggingClientInterceptor(grpc.UnaryUnaryClientInterceptor):  # pragma: NO COVER
+    def intercept_unary_unary(self, continuation, client_call_details, request):
+        request_metadata = client_call_details.metadata
+        if CLIENT_LOGGING_SUPPORTED and _LOGGER.isEnabledFor(std_logging.DEBUG):
+            try:
```
vchudnov-g, line 45: `request_metadata` is only used for logging, right?

```suggestion
        if CLIENT_LOGGING_SUPPORTED and _LOGGER.isEnabledFor(std_logging.DEBUG):
                       request_metadata = client_call_details.metadata
                        try:
```

tests/integration/goldens/credentials/google/iam/credentials_v1/services/iam_credentials/transports/grpc.py
```diff
@@ -186,6 +246,11 @@ def __init__(self, *,
                 ],
             )
 
+        self._grpc_channel = self._base_channel
+        if CLIENT_LOGGING_SUPPORTED and _LOGGER.isEnabledFor(std_logging.DEBUG):  # pragma: NO COVER
```
vchudnov-g, line 250: I suggest checking here whether any logging levels are handled at all, so using either `std_logging.NOTSET` or the next highest priority, the numeric `1` ([ref](https://docs.python.org/3/library/logging.html#levels))

tests/integration/goldens/credentials/google/iam/credentials_v1/services/iam_credentials/transports/grpc.py
```diff
@@ -238,7 +303,7 @@ def create_channel(cls,
     def grpc_channel(self) -> grpc.Channel:
         """Return the channel designed to connect to this service.
         """
-        return self._grpc_channel
+        return self._base_channel
```
vchudnov-g, line 306: Why not `_grpc_channel` here, especially if you're concerned about backwards compatibility?

We could define a `base_channel` method to return `_base_channel`.

This would also reduce naming confusion.

vchudnov-g (commented):
Looks good. One non-trivial question.suggestion about the accessor.

tests/integration/goldens/credentials/google/iam/credentials_v1/services/iam_credentials/transports/grpc.py
ohmayr: If logging is turned on `_grpc_channel` becomes a channel of type `_interceptor.Channel` which can break the user workflow if they turn on logging. `_base_channel` is the previous `_grpc_channel`.

vchudnov-g: Then how about renaming your current `_grpc_channel` to `_logged_channel` and your current `_base_channel` to `_grpc_channel`? That will keep the naming clearer, particularly with this accessor.

ohmayr: Take a look here for the async implementation (especially the mixins templated file which is used for both sync and async https://github.com/googleapis/gapic-generator-python/pull/2291/files#diff-da6458fa34d168ea1f5b9973cee7264ccca58e61d4a6aa48dc5c061786daf2a6)

If I rename channels as per your suggestion in `grpc.py.j2`, I'll have to update `_grpc_channel` to `_logged_channel` in `grpc_asyncio.j2` (something that I haven't touched) and in `_mixins.py.j2` (where I've simply changed the public accessor `grpc_channel` to the private accessor `_grpc_channel`.

Let me know what you think

tests/integration/goldens/credentials/google/iam/credentials_v1/services/iam_credentials/transports/grpc.py
ohmayr: done

tests/integration/goldens/credentials/google/iam/credentials_v1/services/iam_credentials/transports/grpc.py
ohmayr: addressed.

tests/integration/goldens/credentials/google/iam/credentials_v1/services/iam_credentials/transports/grpc.py
ohmayr: I'm open to renaming the channels since I do agree that the current names aren't very appropriate. (Another thing that I'd like to highlight):

```
self._grpc_channel = self._base_channel
if CLIENT_LOGGING_SUPPORTED and not _LOGGER.isEnabledFor(std_logging.NOTSET):  # pragma: NO COVER
    self._interceptor = _LoggingClientInterceptor()
    self._grpc_channel =  grpc.intercept_channel(self._base_channel, self._interceptor)
```

`self._grpc_channel` is the base channel if logging isn't enabled and is otherwise the intercepted channel. So renaming it to `self._logged_channel` won't be too appropriate either (unless we change our logic).

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
```diff
@@ -258,7 +258,7 @@ class {{ service.name }}GrpcTransport({{ service.name }}Transport):
             )
 
         self._grpc_channel = self._base_channel
-        if CLIENT_LOGGING_SUPPORTED and _LOGGER.isEnabledFor(std_logging.DEBUG):  # pragma: NO COVER
+        if CLIENT_LOGGING_SUPPORTED and not _LOGGER.isEnabledFor(std_logging.NOTSET):  # pragma: NO COVER
```
vchudnov-g, line 261: (see my other comment on this code line first)

We do NOT want the `not` in  `not _LOGGER.isEnabledFor(std_logging.NOTSET)`: If this logger does not have a level specified, `isEnabledFor` will traverse up the chain until it finds a handler ([ref](https://docs.python.org/3/library/logging.html#logging.Logger.isEnabledFor)). So this use case would fail: `_LOGGER`  has no handlers (since it is `NOTSET`) but its parent is configured with, say, `DEBUG`. Then `isEnabledFor` would return `True`, the `not` will make it `False`, and the user sees no log. This runs counter to the idiomatic expectations of logging in Python.

We should remove the `not`. If checking for `NOTSET` causes an issue, then replace the argument with the value 1, which is the next highest debug level ([ref](https://docs.python.org/3/library/logging.html#levels)).

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
```diff
@@ -55,8 +55,8 @@ _LOGGER = std_logging.getLogger(__name__)
 
 class _LoggingClientInterceptor(grpc.UnaryUnaryClientInterceptor):  # pragma: NO COVER
     def intercept_unary_unary(self, continuation, client_call_details, request):
-        request_metadata = client_call_details.metadata
-        if CLIENT_LOGGING_SUPPORTED and _LOGGER.isEnabledFor(std_logging.DEBUG):
+        if CLIENT_LOGGING_SUPPORTED and not _LOGGER.isEnabledFor(std_logging.NOTSET):  # pragma: NO COVER
```
vchudnov-g, line 58: We should either revert these checks in this function to be the explicit checks for `DEBUG`, since we know that's what we're logging, or we should apply here the fix in my comment at the third logging-check location in this file, outside this function.

vchudnov-g, line 58: Since we're doing this the same check twice in this function, we should do it once and save the result in a variable, and use the variable in the two `if` statements.

[That said, we're only adding the interceptor if the check passes (line 261), so having the check both there and here is redundant. But I think it's worth keeping this check.]

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
vchudnov-g, line 261: We only set up these channels when we set up the transport, right, and they're used for various RPC calls?

Then we will run into a problem if the user changes their logging config at run time. Specifically:
1. The user sets things up and does their first RPC call with no logging.
2. The user code then sets up logging either for this module or one of its ancestors.
3. The user makes an RPC call, expecting it to be logged, but since the transport was set up without logging, this interceptor never gets called.

I think we should always add the logging interceptor, and have it do the check when it is called. If logging is off, it's a single check and a pass-through. And if the logging config gets changes in between RPC calls, it will pick it up. This means removing the logging check here, so my other comment about changing it here does not apply.

vchudnov-g (changes requested):

tests/integration/goldens/credentials/google/iam/credentials_v1/services/iam_credentials/transports/grpc.py
vchudnov-g: As per my comments on the latest review (2024-12-11 @ 09:04 PT), I think there's a good argument to always have the logging interceptor. In the worst case it's just a pass-through. Always having it enables us to respond to logging config changes at run-time.

I think renaming the channel names will make things clearer and reduce confusion.

ohmayr: addressed.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
ohmayr: addressed.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
ohmayr: I've reverted the statements for now but I do have a question:

> _LOGGER has no handlers (since it is NOTSET) but its parent is configured with, say, DEBUG. Then isEnabledFor would return True, the not will make it False, and the user sees no log. This runs counter to the idiomatic expectations of logging in Python.

If a level is configured against the parent, then isEnabledFor(NOTSET) should return False and the not should make it True.

The isEnabledFor should check for the level configured against all loggers up in the chain.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
ohmayr: > the third logging-check location in this file, outside this function

Is that the guard for the logging channel? I just removed it based on the other comment.

gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2
ohmayr: https://docs.python.org/3/library/logging.html#logging.Logger.isEnabledFor

The idea for checking `not IsEnabledFor(NOTSET)` and `isEnabledFor(DEBUG)` is similar. The former imo would be better if we want to emit our logs if let's say someone has configured a logging level > DEBUG.

vchudnov-g: > If a level is configured against the parent, then isEnabledFor(NOTSET) should return False and the not should make it True.

As per https://docs.python.org/3/library/logging.html#logging.Logger.isEnabledFor, if this logger is `NOTSET`, the check will go up the hierarchy.

> The isEnabledFor should check for the level configured against all loggers up in the chain.

Almost. Our intent is that it do so, agreed, but I noticed yesterday that the semantics are not quite that, so I filed #2292.

vchudnov-g (commented):
LGTM. I can do a final pass once you have the if/elif stuff down.

vchudnov-g (approved):
One comment

vchudnov-g (approved):

PR #2288, opened 2024-12-10
fix(log): preserve dict of rest async response headers

vchudnov-g:


vchudnov-g:
OK, I was able to update the goldens!

ohmayr (approved):

PR #2291, opened 2024-12-10
feat: add client debug logging support for async gRPC

### LINKED ISSUE #16242: [logging] add client logging support for gRPC streaming RPCs
add client logging support for gRPC streaming RPCs

ohmayr:
Fixes googleapis/google-cloud-python#16242 🦕

### DIFF
--- a/tests/integration/goldens/credentials/google/iam/credentials_v1/services/iam_credentials/transports/grpc_asyncio.py
+++ b/tests/integration/goldens/credentials/google/iam/credentials_v1/services/iam_credentials/transports/grpc_asyncio.py
@@ -57,6 +57,11 @@ class _LoggingClientAIOInterceptor(grpc.aio.UnaryUnaryClientInterceptor):  # pra
                 request_payload = MessageToJson(request)
             else:
                 request_payload = f"{type(request).__name__}: {pickle.dumps(request)}"
+
+            request_metadata = {
+                key: value.decode("utf-8") if isinstance(value, bytes) else value
+                for key, value in request_metadata
+            }
             grpc_request = {
                 "payload": request_payload,
                 "requestMethod": "grpc",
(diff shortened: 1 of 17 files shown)


tests/integration/goldens/credentials/google/iam/credentials_v1/services/iam_credentials/transports/grpc_asyncio.py
```diff
@@ -232,8 +300,11 @@ def __init__(self, *,
                 ],
             )
 
-        # Wrap messages. This must be done after self._grpc_channel exists
+        self._interceptor = _LoggingClientAIOInterceptor()
+        self._grpc_channel._unary_unary_interceptors.append(self._interceptor)
+        self._logged_channel = self._grpc_channel
```
vchudnov-g, line 305: Why is `_logged_channel` the same as `_grpc_channel`? Can we do what we do in the sync case and add the logging interceptor only to `_logged_channel`?

ohmayr: We can't. The implementation for sync vs async channels (i.e. adding interceptors) is different. sync provides a new channel class for interceptors vs async allows us to append the interceptor directly to the original class.

parthea (approved):

tests/integration/goldens/credentials/google/iam/credentials_v1/services/iam_credentials/transports/grpc_asyncio.py
vchudnov-g: OK, so both identifiers point to the same object? Then I would suggest, for clarity, that where you changed `self.grpc_channel` to `self._grpc_channel`, you instead use `self._logged_channel`. It's functionally the same, but it's a good reminder to people reading this in the future that the channel is being logged. Also makes it clearer given that `_mixins.py.j2` uses `logged_channel`.

vchudnov-g (approved):
LGTM, but please consider the renaming in my comment for clarity.

tests/integration/goldens/credentials/google/iam/credentials_v1/services/iam_credentials/transports/grpc_asyncio.py
ohmayr: Yep both point to the same object. It just makes it easier to use the same template for mixins. I've updated the instances of `_grpc_channel` to `_logged_channel` in RPC calls.

PR #2300, opened 2024-12-17
tests: add showcase tests for reading grpc/rest response metadata

parthea:
This PR adds showcase integration tests for reading response metadata using the following matrix: [(`gRPC`, `REST`), (`sync` , `async`)]
This PR depends on https://github.com/googleapis/gapic-generator-python/pull/2370.


tests/system/conftest.py
```diff
@@ -276,12 +295,80 @@ def intercept_stream_stream(
...
+    grpc.aio.StreamUnaryClientInterceptor,
+    grpc.aio.StreamStreamClientInterceptor,
+):
+    def __init__(self, key, value):
+        self._key = key
+        self._value = value
+        self.request_metadata = []
+        self.response_metadata = []
+
+    async def _add_metadata(self, client_call_details):
```
vchudnov-g, line 310:
```suggestion
    async def _add_request_metadata(self, client_call_details):
```

tests/system/conftest.py
```diff
@@ -276,12 +295,80 @@ def intercept_stream_stream(
...
+        metadata = [(k, str(v)) for k, v in await response.trailing_metadata()]
+        self.response_metadata = metadata
+        return response
+
+    async def intercept_unary_stream(self, continuation, client_call_details, request):
+        self._add_metadata(client_call_details)
+        response_it = continuation(client_call_details, request)
+        return response_it
+
+    async def intercept_stream_unary(
```
vchudnov-g, line 327: We should use more descriptive names to differentiate this method from the previous one. I realize previously existing classes established this pattern, but we should clarify.

tests/system/conftest.py
```diff
@@ -276,12 +295,80 @@ def intercept_stream_stream(
...
+
+    async def intercept_stream_stream(
+        self, continuation, client_call_details, request_iterator
+    ):
+        self._add_metadata(client_call_details)
+        response_it = continuation(client_call_details, request_iterator)
+        return response_it
+
+
+class MetadataClienRestInterceptor(EchoRestInterceptor):
```
vchudnov-g, line 342: I would clarify in the name that this is for the echo service

```suggestion
class EchoMetadataClienRestInterceptor(EchoRestInterceptor):
```

tests/system/conftest.py
```diff
@@ -293,4 +380,39 @@ def intercepted_echo(use_mtls):
...
+        credentials=ga_credentials.AnonymousCredentials(),
+        host="localhost:7469",
+        url_scheme="http",
+        interceptor=interceptor,
+    )
+    return EchoClient(transport=transport), interceptor
+
+
+@pytest.fixture
+def intercepted_echo_grpc_async():
```
vchudnov-g, line 403: Put the two grpc fixtures next to each other?

tests/system/conftest.py
```diff
@@ -293,4 +380,39 @@ def intercepted_echo(use_mtls):
         credentials=ga_credentials.AnonymousCredentials(),
         channel=intercept_channel,
     )
-    return EchoClient(transport=transport)
+    return EchoClient(transport=transport), interceptor
+
+
+@pytest.fixture
+def intercepted_echo_rest():
```
vchudnov-g, line 387: We're waiting for REST async or can we test that now?

tests/system/test_grpc_interceptor_streams.py
```diff
@@ -21,9 +21,10 @@
 intercepted_metadata = (('showcase-trailer', 'intercepted'),)
 
 
-def test_unary_stream(intercepted_echo):
+def test_unary_stream(intercepted_echo_grpc):
```
vchudnov-g, line 24: Do we want to test async grpc here?

tests/system/test_response_metadata.py
```diff
@@ -0,0 +1,77 @@
...
+# limitations under the License.
+
+import os
+import pytest
+
+from google import showcase
+
+
+@pytest.mark.parametrize(
+    "transport,response_metadata",
```
vchudnov-g, line 22: How do the interceptors used int his file know about the response metadata they should be injecting? Or  is this real response metadata from Showcase? I'm not finding it by doing a search n the Showcase repo.

tests/system/conftest.py
parthea: fixed in https://github.com/googleapis/gapic-generator-python/pull/2300/commits/e302dc71914d19b9f106351378c18ea9f729c3b4

tests/system/conftest.py
parthea: This is not something that we can change as it's an abstract method in gRPC that we're overriding

https://grpc.github.io/grpc/python/grpc.html#grpc.StreamUnaryClientInterceptor.intercept_stream_unary

tests/system/conftest.py
parthea: Fixed in https://github.com/googleapis/gapic-generator-python/pull/2300/commits/1ae8aaaf6076170af9f9338d681b311623898a66

tests/system/conftest.py
parthea: Fixed in https://github.com/googleapis/gapic-generator-python/pull/2300/commits/7147a581de1d892de471286c46f28f215b64740a

tests/system/test_grpc_interceptor_streams.py
parthea: Fixed in 05f6c65299187262b53df727e8a6bbbf45e60a66

tests/system/conftest.py
parthea: Fixed in https://github.com/googleapis/gapic-generator-python/pull/2300/commits/05f6c65299187262b53df727e8a6bbbf45e60a66

tests/system/test_response_metadata.py
parthea: The interceptors aren't injecting anything here. There is a feature of the echo `rpc` which adds response headers(metadata) automatically for `REST` with the `X-Showcase-Request` prefix which was added in https://github.com/googleapis/gapic-showcase/pull/1509. For `gRPC`, I believe this feature was added in https://github.com/googleapis/gapic-showcase/pull/970

vchudnov-g (approved):
One q.

PR #2308, opened 2025-01-06
chore: add documentation for logging

ohmayr:
Fixes b/384047953

### DIFF
--- a/gapic/templates/README.rst.j2
+++ b/gapic/templates/README.rst.j2
@@ -85,7 +85,7 @@ Examples
 
     export GOOGLE_SDK_PYTHON_LOGGING_SCOPE=google
 
-- Enabling the default handler for a specific Google module (for a client library called :code:`Library`):
+- Enabling the default handler for a specific Google module (for a client library called :code:`library_v1`):
 
 .. code-block:: console
 
@@ -134,7 +134,7 @@ Logging details
    :code:`logging.getLogger("google").propagate = True` in your code.
 #. You can mix the different logging configurations above for different Google modules. For example, you may want use a code-based logging configuration for
    one library, but decide you need to also set up environment-based logging configuration for another library.
-   1.  If you attempt to use both code-based and environment-based configuration for the same module, the environment-based configuration will be ineffectual
+   #. If you attempt to use both code-based and environment-based configuration for the same module, the environment-based configuration will be ineffectual
    if the code -based configuration gets applied first.
 #. The Google-specific logging configurations (default handlers for environment-based configuration; not propagating logging events to the root logger) get
    executed the first time *any* client library is instantiated in your application, and only if the affected loggers have not been previously configured.
--- a/gapic/templates/README.rst.j2
+++ b/gapic/templates/README.rst.j2
@@ -140,4 +140,4 @@ Logging details
 
 #. The Google-specific logging configurations (default handlers for environment-based configuration; not propagating logging events to the root logger) get
    executed the first time *any* client library is instantiated in your application, and only if the affected loggers have not been previously configured.
-   (This is the reason for 2.1 above.)
+   (This is the reason for 2.i. above.)
--- a/gapic/templates/README.rst.j2
+++ b/gapic/templates/README.rst.j2
@@ -134,8 +134,10 @@ Logging details
    :code:`logging.getLogger("google").propagate = True` in your code.
 #. You can mix the different logging configurations above for different Google modules. For example, you may want use a code-based logging configuration for
    one library, but decide you need to also set up environment-based logging configuration for another library.
+
    #. If you attempt to use both code-based and environment-based configuration for the same module, the environment-based configuration will be ineffectual
       if the code -based configuration gets applied first.
+
 #. The Google-specific logging configurations (default handlers for environment-based configuration; not propagating logging events to the root logger) get
    executed the first time *any* client library is instantiated in your application, and only if the affected loggers have not been previously configured.
    (This is the reason for 2.1 above.)
--- a/gapic/templates/README.rst.j2
+++ b/gapic/templates/README.rst.j2
@@ -49,7 +49,7 @@ Windows
     <your-env>\Scripts\pip.exe install \path\to\library
 
 
-LOGGING
+Logging
 -------
 
 This library uses the standard Python :code:`logging` functionality to log some RPC events that could be of interest for debugging and monitoring purposes.
--- a/gapic/templates/README.rst.j2
+++ b/gapic/templates/README.rst.j2
@@ -55,7 +55,7 @@ LOGGING
 This library uses the standard Python :code:`logging` functionality to log some RPC events that could be of interest for debugging and monitoring purposes.
 Note the following:
 
-#. Logs may contain sensitive information. Take care to **restrict access to the logs** if they are saved, whether it be on local storage or Google Cloud Logging.
+#. Logs may contain sensitive information. Take care to **restrict access to the logs** if they are saved, whether it be on local storage or on Google Cloud Logging.
 #. Google may refine the occurrence, level, and content of various log messages in this library without flagging such changes as breaking. **Do not depend on immutability of the logging events**.
 #. By default, the logging events from this library are not handled. You must **explicitly configure log handling** using one of the mechanisms below.
 
--- a/gapic/templates/README.rst.j2
+++ b/gapic/templates/README.rst.j2
@@ -48,3 +48,49 @@ Windows
     <your-env>\Scripts\activate
     <your-env>\Scripts\pip.exe install \path\to\library
 
+
+LOGGING
+-------
+
+This library has support to enable logging for debugging and monitoring purposes. Note that logs may contain sensitive information and care should be
+taken to restrict access to the logs if they are saved, whether it be on local storage or Google Cloud Logging.
+
+Users must **explicitly opt-in** to enable logging by configuring the `GOOGLE_SDK_PYTHON_LOGGING_SCOPE` environment variable with a valid logging scope.
+
+A logging scope is a namespace that begins with :code:`google`.
+
+- Valid logging scopes: :code:`google`, :code:`google.cloud.asset.v1`, :code:`google.api`, :code:`google.auth`, etc.
+- Invalid logging scopes: :code:`foo`, :code:`123`, etc.
+
+**NOTE**: If an invalid logging scope is configured, we do not act on the corresponding logger.
+
+To set a handler that applies to all Google-based loggers:
+
+.. code-block:: console
+
+    export GOOGLE_SDK_PYTHON_LOGGING_SCOPE=google
+
+To enable a handler for a specific GAPIC:
+
+.. code-block:: console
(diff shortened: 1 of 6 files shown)


gapic/templates/README.rst.j2
```diff
@@ -48,3 +48,49 @@ Windows
...
     <your-env>\Scripts\pip.exe install \path\to\library
 
+
+LOGGING
+-------
+
+This library has support to enable logging for debugging and monitoring purposes. Note that logs may contain sensitive information and care should be
+taken to restrict access to the logs if they are saved, whether it be on local storage or Google Cloud Logging.
+
+Users must **explicitly opt-in** to enable logging by configuring the `GOOGLE_SDK_PYTHON_LOGGING_SCOPE` environment variable with a valid logging scope.
```
parthea, line 58:
```suggestion
By default this feature is not enabled. Users may opt-in to enable logging by configuring the `GOOGLE_SDK_PYTHON_LOGGING_SCOPE` environment variable with a valid logging scope.
```

gapic/templates/README.rst.j2
```diff
@@ -48,3 +48,49 @@ Windows
...
+
+A logging scope is a namespace that begins with :code:`google`.
+
+- Valid logging scopes: :code:`google`, :code:`google.cloud.asset.v1`, :code:`google.api`, :code:`google.auth`, etc.
+- Invalid logging scopes: :code:`foo`, :code:`123`, etc.
+
+**NOTE**: If an invalid logging scope is configured, we do not act on the corresponding logger.
```
parthea, line 65:
```suggestion
**NOTE**: If an invalid logging scope is configured, debug logging will remain disabled.
```

gapic/templates/README.rst.j2
```diff
@@ -48,3 +48,49 @@ Windows
...
+
+To set a handler that applies to all Google-based loggers:
+
+.. code-block:: console
+
+    export GOOGLE_SDK_PYTHON_LOGGING_SCOPE=google
+
+To enable a handler for a specific GAPIC:
```
parthea, line 73:
```suggestion
To enable a handler for a specific module:
```

gapic/templates/README.rst.j2
```diff
@@ -48,3 +48,49 @@ Windows
...
+
+.. code-block:: console
+
+    export GOOGLE_SDK_PYTHON_LOGGING_SCOPE=google.cloud.abc.v1
```
parthea, line 77:
```suggestion
    export GOOGLE_SDK_PYTHON_LOGGING_SCOPE=google.cloud.abc_v1
```

gapic/templates/README.rst.j2
```diff
@@ -48,3 +48,49 @@ Windows
...
+
+:code:`GOOGLE_SDK_PYTHON_LOGGING_SCOPE` allows users to enable or disable logs by configuring a log system; it does not let them configure log levels, handlers, formatters, etc.
```
parthea, line 79:
```suggestion
:code:`GOOGLE_SDK_PYTHON_LOGGING_SCOPE` allows users to enable or disable logs by configuring a log system; however it is not currently possible to customize log levels, handlers, formatters, etc.
```

gapic/templates/README.rst.j2
```diff
@@ -48,3 +48,49 @@ Windows
...
+
+Loggers so configured will handle log messages at level DEBUG or higher, outputting them to stderr. The default log format for these log messages is structured log format.
```
parthea, line 81:
```suggestion
Configured loggers will handle log messages at level DEBUG or higher, outputting them to stderr. The default log format for these log messages is structured log format.
```

parthea, line 81: Would it be helpful to provide a one line sample of the log output so that users can see what debug logs will look like?

parthea (approved):
Added minor observations, otherwise LGTM.

gapic/templates/README.rst.j2
vchudnov-g, line 1: I found it easier to make changes separately, and this is the text I suggest. WDYT?

Some notes:

1. If you prefer to discuss the text in a Google Doc first, I can easily export it to one and we can discuss the content there.
2. It would be great to not use a generic example library in this file (in the content below, `google.cloud.library_v1`) but rather the actual name of the library we're generating. This should be simple because we're in a template, but if it's not, let's file a follow-up issue to get back to this.

(sorry I couldn't do this directly in RST. Once we agree on the content, I should be able to export to RST with minimal set-up on my part, if that's helpful.)

SUGGESTED CONTENT FOLLOWS

-----

# Logging

This library uses the standard Python `logging` functionality to log some RPC events that could be of interest for debugging and monitoring purposes. Note the following:

1.  Logs may contain sensitive information. Take care to **restrict access to the logs** if they are saved, whether it be on local storage or Google Cloud Logging.
4.  Google may refine the occurrence, level, and content of various log messages in this library without flagging such changes as breaking. **Do not depend on immutability of the logging events**.
5.  By default, the logging events from this library are not handled. You must **explicitly configure log handling** using one of the mechanisms below in order for the logging events to be handled.



## Simple, environment-based configuration

To enable logging for this library without any changes in your code, set the `GOOGLE_SDK_PYTHON_LOGGING_SCOPE` environment variable to a valid Google logging scope. This configures handling of logging events (at level `logging.DEBUG` or higher) from this library in a default manner, emitting the logged messages in a structured format. It does not currently allow customizing the logging levels captured nor the handlers, formatters, etc. used for any logging event.

A logging scope is a period-separated namespace that begins with `google`, identifying the Python module or package to log.

-   Valid logging scopes: `google`, `google.cloud.asset.v1`, `google.api`, `google.auth`, etc.
-   Invalid logging scopes: `foo`, `123`, etc.

****NOTE****: If the logging scope is invalid, the library does not set up any logging handlers.



### Examples

-   Enabling the default handler for all Google-based loggers

        export GOOGLE_SDK_PYTHON_LOGGING_SCOPE=google

-   Enabling the default handler for a specific Google module

    In the case of a client library called `Library`:
    
        export GOOGLE_SDK_PYTHON_LOGGING_SCOPE=google.cloud.library_v1


<a id="org3ee21c8"></a>

## Advanced, code-based configuration

You can also configure a valid logging scope using Python's standard `logging` mechanism.


<a id="org4baf081"></a>

### Examples

-   Configuring a handler for all Google-based loggers

        import logging
        
        from google.cloud.translate_v3 import translate
        
        base_logger = logging.getLogger("google")
        base_logger.addHandler(logging.StreamHandler())
        base_logger.setLevel(logging.DEBUG)

-   Configuring a handler for a specific Google module

    In the case of a client library called `Library`:
    
        import logging
        
        from google.cloud.translate_v3 import translate
        
        base_logger = logging.getLogger("google.cloud.library_v1")
        base_logger.addHandler(logging.StreamHandler())
        base_logger.setLevel(logging.DEBUG)


<a id="org306b2b5"></a>

## Logging details

1.  Regardless of which of the mechanisms above you use to configure logging for this library, by default logging events are not propagated up to the root logger from the `google`-level logger. If you need the events to be propagated to the root logger, you must explicitly set `logging.getLogger("google").propagate = True` in your code.
2.  You can mix the different logging configurations above for different Google modules. For example, you may want use a code-based logging configuration for one library, but decide you need to also set up environment-based logging configuration for another library.
    1.  If you attempt to use both code-based and environment-based configuration for the same module, the environment-based configuration will be ineffectual if the code -based configuration gets applied first.
4.  The Google-specific logging configurations (default handlers for environment-based configuration; not propagating logging events to the root logger) get executed the first time *any* client library is instantiated in your application, and only if the affected loggers have not been previously configured. (This is the reason for 2.1 above.)

gapic/templates/README.rst.j2
```diff
@@ -48,3 +48,49 @@ Windows
     <your-env>\Scripts\activate
     <your-env>\Scripts\pip.exe install \path\to\library
 
+
+LOGGING
+-------
+
+This library has support to enable logging for debugging and monitoring purposes. Note that logs may contain sensitive information and care should be
```
vchudnov-g, line 55: I'm not sure it's clear or accurate to say we support "enabling" logging. We log, in the sense we emit logging events. Users can set up handlers for those events, either explicitly via code or with the help of our environment-based config, which we describe below. I tried to clear this up in the suggested text.

gapic/templates/README.rst.j2
vchudnov-g: - we shouldn't call this "debug logging" externally (see internal doc comment, for example)
- "will remain disabled" is misleading, because the user could set up handlers a different way

I tried to clear this up in the suggested text.

gapic/templates/README.rst.j2
vchudnov-g: I'm not sure how much value that adds. Plus, since we're giving ourselves the leeway to change the log messages, this could easily become stale, depending on how much information we provide here.

tests/integration/goldens/credentials/README.rst
```diff
@@ -47,3 +47,95 @@ Windows
...
+
+Logging details
+~~~~~~~~~~~~~~~
+
+#. Regardless of which of the mechanisms above you use to configure logging for this library, by default logging events are not propagated up to the root
+   logger from the `google`-level logger. If you need the events to be propagated to the root logger, you must explicitly set
+   :code:`logging.getLogger("google").propagate = True` in your code.
+#. You can mix the different logging configurations above for different Google modules. For example, you may want use a code-based logging configuration for
+   one library, but decide you need to also set up environment-based logging configuration for another library.
+   1.  If you attempt to use both code-based and environment-based configuration for the same module, the environment-based configuration will be ineffectual
```
vchudnov-g, line 137: This nested 1 does not rendered correctly (at least according to the GitHub preview if you click on the page icon, the second from the left, in the set of icons for this file shown right above this diff, on the right).
icon

We need to specify it's a sub-bullet.

vchudnov-g (commented):
If it's not trivial to use the info for the library being generated rather than the generic "library" API, could you file an issue to follow up on that? Thanks! (This was suggestion 2 in my [previous comment](https://github.com/googleapis/gapic-generator-python/pull/2308#discussion_r1907703590))

ohmayr:
> If it's not trivial to use the info for the library being generated rather than the generic "library" API, could you file an issue to follow up on that? Thanks! (This was suggestion 2 in my [previous comment](https://github.com/googleapis/gapic-generator-python/pull/2308#discussion_r1907703590))

Filed https://github.com/googleapis/gapic-generator-python/issues/2314

gapic/templates/README.rst.j2
```diff
@@ -48,3 +48,96 @@ Windows
     <your-env>\Scripts\activate
     <your-env>\Scripts\pip.exe install \path\to\library
 
+
+LOGGING
```
vchudnov-g, line 52: For consistency with the other sections:

```suggestion
Logging
```

vchudnov-g (approved):
Just one minor change for consistency.

Thanks so much for doing this, Omair! It looks great!

Issue #2322, opened 2025-02-06
Remove constraint for googleapis-common-protos in requirements.in


parthea:
In PR https://github.com/googleapis/gapic-generator-python/pull/2321, A constraint  for `googleapis-common-protos>=1.67.0rc1` was added to `requirements.in` to unblock development for the selective GAPIC generation. We should change this back to `googleapis-common-protos` once a newer version of `googleapis-common-protos` is published.

PR #2323, opened 2025-02-06
build: Update setup.py with updated version of googleapis-common-protos

gkevinzheng:
The unit tests I have for selective GAPIC are only able to pass with this `setup.py` change. This is because each invocation of `nox -s unit` installs the GAPIC generator in a virtualenv via `pip install -e .`, which uses the dependencies listed under `setup.py`.


gkevinzheng:
Linking this to #2322 as an additional TODO item once a formal release of `googleapis-common-protos` version 1.67.0 is out.

PR #2325, opened 2025-02-10
feat: Added support for internal methods in selective GAPIC generation

gkevinzheng:
This PR implements Selective GAPIC generation Phase 2 for Python, focusing on the generation of internal methods and client classes.

Overall design doc: [go/selective-gapic-generation-p2](http://go/selective-gapic-generation-p2)
Implementation doc for Python: [go/selective-gapic-gen-python-p2](http://go/selective-gapic-gen-python-p2)


setup.py
```diff
@@ -29,7 +29,9 @@
     # templated setup.py.j2: https://github.com/googleapis/gapic-generator-python/blob/main/gapic/templates/setup.py.j2
     "click >= 6.7",
     "google-api-core[grpc] >= 1.34.1, <3.0.0dev,!=2.0.*,!=2.1.*,!=2.2.*,!=2.3.*,!=2.4.*,!=2.5.*,!=2.6.*,!=2.7.*,!=2.8.*,!=2.9.*,!=2.10.*",
-    "googleapis-common-protos >= 1.55.0",
+    # TODO(https://github.com/googleapis/gapic-generator-python/issues/2322): Upgrade the version of googleapis-common-protos once a
+    # formal version of 1.67.0 is published.
+    "googleapis-common-protos >= 1.67.0rc1",
```
parthea, line 34: Note to self: We should release `googleapis-common-protos` prior to merging this PR and update this line

gapic/schema/wrappers.py
```diff
@@ -1928,7 +2003,7 @@ def __getattr__(self, name):
     @property
     def client_name(self) -> str:
         """Returns the name of the generated client class"""
-        return self.name + "Client"
+        return ("Base" if self.is_internal else "") + self.name + "Client"
```
parthea, line 2006: We should have 2 clients , one with the `Base` prefix and one without `Base` prefix, in the case that some rpcs are internal and some aren't

gkevinzheng: I believe that even in the mixed case with a `Service` having a mix of internal and non-internal methods, we would be having one `Base`-prefixed client with both kinds of methods in it.

tests/integration/BUILD.bazel
```diff
@@ -29,6 +29,7 @@ INTEGRATION_TEST_LIBRARIES = [
     "logging",  # Java package remapping in gapic.yaml.
     "redis",  # Has a gapic.yaml.
     "redis_selective", # Selective generation.
+    "logging_internal", # Selective generation with internal client classes.
```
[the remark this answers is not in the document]
gkevinzheng: I did this based off feedback I got from @vchudnov-g for the Selective GAPIC Phase 1 PR.

setup.py
parthea: We can revert this change now(See https://github.com/googleapis/gapic-generator-python/pull/2330)

gapic/schema/wrappers.py
parthea: Acknowledged

tests/integration/BUILD.bazel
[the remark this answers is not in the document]
parthea: SG. I won't block on this. We can remove it later on if we find that it's toilsome to maintain another golden API.

parthea (approved):

setup.py
parthea, line 34:
```suggestion
    "googleapis-common-protos >= 1.55.0",
```

PR #2329, opened 2025-02-12
feat: add ability to remove unversioned modules

parthea:
Googlers see go/gapic-default-imports


tests/unit/generator/test_generator.py
```diff
@@ -145,6 +146,50 @@ def test_get_response_ignore_gapic_metadata():
             assert res.file == CodeGeneratorResponse().file
 
 
+@pytest.mark.parametrize("unversioned_package_disabled", (True, False))
+def test_get_response_ignore_unversioned_package(unversioned_package_disabled):
+    g = make_generator()
```
vchudnov-g, line 151: Please don't use one character variable names. I see the same construct is used in other tests, but I advocate not perpetuating the practice.

```suggestion
    generator = make_generator()
```

tests/unit/generator/test_generator.py
```diff
@@ -145,6 +146,50 @@ def test_get_response_ignore_gapic_metadata():
...
+                )
+            ]
+        )
+    )
+    api = make_api(
+        naming=naming,
+        service_yaml_config=service_config,
+    )
+
+    with mock.patch.object(jinja2.FileSystemLoader, "list_templates") as lt:
```
vchudnov-g, line 177: I'd still suggest avoiding short names if possible, though these are two character and use closed to the definition, which makes it more amenabel IMO.

Conider s/`lt`/`lister`/ and s/`gt`/`getter`/

vchudnov-g (approved):

tests/unit/generator/test_generator.py
parthea: Fixed in https://github.com/googleapis/gapic-generator-python/pull/2329/commits/20f40506c9c6596b73312a083ac2765c616c73b7

tests/unit/generator/test_generator.py
parthea: Fixed in https://github.com/googleapis/gapic-generator-python/pull/2329/commits/6491e53ca0a0a4f5914b705b20778144f0a4588c

PR #2344, opened 2025-02-25
tests: add testing for Python 3.14 Pre-release

### LINKED ISSUE #2239: Add presubmits for Python 3.14 Pre-release
https://www.python.org/download/pre-releases/

parthea:
Fixes #2239

BEGIN_COMMIT_OVERRIDE
fix: add testing for Python 3.14 Pre-release
END_COMMIT_OVERRIDE


noxfile.py
```diff
@@ -44,9 +44,10 @@
     "3.11",
     "3.12",
     "3.13",
+    "3.14",
 )
 
-NEWEST_PYTHON = ALL_PYTHON[-1]
+NEWEST_PYTHON = ALL_PYTHON[-2]
```
ohmayr, line 50: Why not test with `3.14`?

ohmayr (approved):

noxfile.py
parthea: I will add this in a follow up PR which adds formal 3.14 support

PR #2349, opened 2025-03-04
fix: allow proto files to be included in the output of py_gapic_assembly_pkg

parthea:
Towards https://github.com/googleapis/google-cloud-python/issues/13545


parthea:
Wait for cl/733327565

ohmayr (approved):

PR #2362, opened 2025-03-15
build: update constraints file to test latest major versions

parthea:
Updates constraints file to ensure that the latest major versions of dependencies are tested.

This fixes a testing gap where presubmits could pass without testing the latest major version. For example, see build log [here](https://github.com/googleapis/gapic-generator-python/actions/runs/13872559057/job/38820975597?pr=2362) which failed using the changes in this PR prior to releasing protobuf 6.x support in  `googleapis-common-protos`.

### DIFF
--- a/gapic/templates/testing/constraints-3.13.txt.j2
+++ b/gapic/templates/testing/constraints-3.13.txt.j2
@@ -1,5 +1,6 @@
 {% from '_pypi_packages.j2' import pypi_packages %}
-# This constraints file is used to check that the latest
+# We use the constraints file for the latest Python version
+# (currently this file) to check that the latest
 # major versions of dependencies are supported in setup.py.
 # List all library dependencies and extras in this file.
 # Require the latest major version be installed for each dependency.
--- a/gapic/templates/testing/constraints-3.13.txt.j2
+++ b/gapic/templates/testing/constraints-3.13.txt.j2
@@ -1,4 +1,19 @@
-# -*- coding: utf-8 -*-
-{% block constraints %}
-{% include "testing/_default_constraints.j2" %}
-{% endblock %}
+{% from '_pypi_packages.j2' import pypi_packages %}
+# This constraints file is used to check that the latest
+# major versions of dependencies are supported in setup.py.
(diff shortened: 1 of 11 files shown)


gapic/templates/testing/constraints-3.13.txt.j2
```diff
@@ -1,4 +1,19 @@
...
+# List all library dependencies and extras in this file.
+# Require the latest major version be installed for each dependency.
+# e.g., if setup.py has "google-cloud-foo >= 1.14.0, < 2.0.0",
+# Then this file should have google-cloud-foo>=1
+google-api-core>=2
+google-auth>=2
+proto-plus>=1
+protobuf>=6
+{% for package_tuple, package_info in pypi_packages.items() %}
+{# Quick check to make sure the package is different from this setup.py #}
```
vchudnov-g, line 13: Could you clarify this comment? I think you're checking we don't circularly include this package in its own constraints file. Is that right?

gapic/templates/testing/constraints-3.13.txt.j2
```diff
@@ -1,4 +1,19 @@
-# -*- coding: utf-8 -*-
-{% block constraints %}
-{% include "testing/_default_constraints.j2" %}
-{% endblock %}
+{% from '_pypi_packages.j2' import pypi_packages %}
+# This constraints file is used to check that the latest
```
vchudnov-g, line 2: Could we make this comment more general, so it's more obvious that when we 3.14 comes out, we should move this logic there?

```suggestion
# We use the constraints file for the latest Python version (currently this file) to check that the latest
```

vchudnov-g (approved):
Looks good, but I suggest two clarifications.

gapic/templates/testing/constraints-3.13.txt.j2
parthea: Fixed in https://github.com/googleapis/gapic-generator-python/pull/2362/commits/b786cc26b9e42d3cf47decc55576aef638ac7b8b

PR #2372, opened 2025-03-24
feat: adds augmented pagination to account for BQ family of APIs

chalmerlowe:
This is a work in progress and contains proof-of-concept code for discussion based on the [design here](https://docs.google.com/document/d/1YSdk477zuj1b2gXmlQ65XoXI2DrUcJ10kMF7lnd8myo/edit?usp=sharing).
Do not merge.

Adds:
* capability to accept `UInt32Value` and `Int32Value` wrapper types
* notes explaining the purpose for the updates

TODO:
- [ ] update tests to account for new code branches AND edge cases
  - [ ] BQ `JobService.GetQueryResults` (should not paginate)
  - [ ] expected packages (i.e. BQ should paginate)
- [x] ensure coverage passes
- [x] blacken/lint the code

Fixes internal bug: #360114649 🦕


gapic/schema/wrappers.py
```diff
@@ -1845,13 +1852,32 @@ def paged_result_field(self) -> Optional[Field]:
...
+        if page_field_size.type == int or (
+            # The following additional checks are related to several members of the BQ family of
+            # APIs, which use the legacy wrapper types: "UInt32Value" and "Int32Value"}
+            # for max_results.
+            # NOTE:
+            #   bigquery_v2 should be paginated
+            #   but bigquery_connection_v1beta1 should NOT be paginated
+            isinstance(page_field_size.type, MessageType)
+            and page_field_size.type.message_pb.name in {"UInt32Value", "Int32Value"}
+            and package_name in self.PACKAGES_WITH_ALLOWED_WRAPPERS
```
leahecole, line 1877: non blocking nit - move this up before the `isinstance` conditional to fail faster, perhaps?

gapic/schema/wrappers.py
```diff
@@ -1845,13 +1852,32 @@ def paged_result_field(self) -> Optional[Field]:
...
+        # the service package must be in the allowlist.
+        package_name = self.input.meta.address.proto_package
+
+        if page_field_size.type == int or (
+            # The following additional checks are related to several members of the BQ family of
+            # APIs, which use the legacy wrapper types: "UInt32Value" and "Int32Value"}
+            # for max_results.
+            # NOTE:
+            #   bigquery_v2 should be paginated
+            #   but bigquery_connection_v1beta1 should NOT be paginated
```
leahecole, line 1874: nit - do we still need this comment given that list connections is no longer specified?

gapic/schema/wrappers.py
parthea, line 1877: Please remove this line. We can add it back if we find that this is going to break other APIs.
```suggestion
```

gapic/schema/wrappers.py
parthea, line 1874: `bigquery_connection_v1beta1` doesn't exist as far as I can tell.
https://github.com/search?q=org%3Agoogleapis+bigquery_connection_v1beta1&type=code

https://github.com/googleapis/google-cloud-python/tree/main/packages/google-cloud-bigquery-connection/google/cloud

https://github.com/googleapis/google-cloud-python/blob/a37aa650281f4621f323b0ac1237ebddef635fbc/packages/google-cloud-bigquery-connection/.OwlBot.yaml#L20

IOW, we don't need to cater for `bigquery_connection_v1beta1`

parthea:
Ideally https://github.com/aip-dev/google.aip.dev/pull/1496 would be merged ahead of this PR to avoid introducing technical debt of inconsistent behavior between the generator and AIPs

gapic/schema/wrappers.py
```diff
@@ -1845,13 +1852,32 @@ def paged_result_field(self) -> Optional[Field]:
...
+            isinstance(page_field_size.type, MessageType)
+            and page_field_size.type.message_pb.name in {"UInt32Value", "Int32Value"}
+            and package_name in self.PACKAGES_WITH_ALLOWED_WRAPPERS
+        ):
+            pass
```
parthea, line 1879: Add a helper function to check the conditions, if needed, so that we can avoid using `pass`

gapic/schema/wrappers.py
chalmerlowe: OBE. Removing the allowlist check.

gapic/schema/wrappers.py
```diff
@@ -1830,13 +1830,35 @@ def ident(self) -> metadata.Address:
         """Return the identifier data to be used in templates."""
         return self.meta.address
 
+    def _validate_paged_field_size_type(page_field_size):
```
leahecole, line 1833:
```suggestion
    def _validate_paged_field_size_type(self, page_field_size):
```

leahecole: this should help the unit tests! I have absolutely been burned by this before

gapic/schema/wrappers.py
chalmerlowe: Done

gapic/schema/wrappers.py
chalmerlowe: Done

leahecole:
Commenting here as well as in the doc for posterity and for future readers, i think it's appropriate to add fragments and potentially to add BQ V2 as a golden

gapic/schema/wrappers.py
```diff
@@ -1830,13 +1830,35 @@ def ident(self) -> metadata.Address:
...
+        Int32Value.
+        """
+
+        _type = page_field_size.type
+
+        if _type == int or (
+            isinstance(_type, MessageType)
+            and _type.message_pb.name in {"UInt32Value", "Int32Value"}
+        ):
+            return True
```
parthea, line 1848: Avoid using `_type` to avoid confusion with the [built-in function type](https://docs.python.org/3/library/functions.html#built-in-functions). Perhaps use `field_type` instead?

Also add type hints to clarify that the return value should be `bool`

```suggestion
        field_type = page_field_size.type

        return field_type == int or (
            isinstance(field_type, MessageType)
            and field_type.message_pb.name in {"UInt32Value", "Int32Value"}
        )
```

gapic/schema/wrappers.py
```diff
@@ -1830,13 +1830,35 @@ def ident(self) -> metadata.Address:
         """Return the identifier data to be used in templates."""
         return self.meta.address
 
+    def _validate_paged_field_size_type(self, page_field_size):
```
parthea, line 1833:
```suggestion
    def _validate_paged_field_size_type(self, page_field_size) -> bool:
```

tests/unit/schema/wrappers/test_method.py
```diff
@@ -999,3 +1055,65 @@ def test_mixin_rule():
         "city": {},
     }
     assert e == m.sample_request
+
+
+@pytest.mark.parametrize(
+    "field_type, _type, expected",
```
parthea, line 1061: Use `pb_type` instead of `_type` since this is referring to the protobuf type and this avoids confusion with the [built-in function type](https://docs.python.org/3/library/functions.html#built-in-functions)

tests/unit/schema/wrappers/test_method.py
```diff
@@ -189,6 +190,61 @@ def test_method_paged_result_field_no_page_field():
     assert method.paged_result_field is None
 
 
+def test_method_paged_result_field_invalid_wrapper_type():
+    """Validate paged_result_field() returns None if page_size/max_results wrappertypes
+    are not allowed types.
+    """
+
+    # page_size is not allowed wrappertype
+    parent = make_field(name="parent", type=9)  # str
```
parthea, line 199:
```suggestion
    parent = make_field(name="parent", type="TYPE_STRING")
```

Use the enum string rather than the numeric value as it improves readability similar to https://github.com/googleapis/gapic-generator-python/blob/c8b7229ba2865d6a2f5966aa151be121de81f92d/tests/unit/schema/wrappers/test_field.py#L275

You can find all of the types here:
https://github.com/protocolbuffers/protobuf/blob/836fb77aebc3bc4a6e2508b67e4ab7977da3bcc1/src/google/protobuf/descriptor.proto#L241

Applies throughout

parthea (commented):
LGTM. Holding off on formal approval until https://github.com/aip-dev/google.aip.dev/pull/1496 is merged and tests in cl/741139812 pass. 

Should we consider updating [showcase protos](https://github.com/googleapis/gapic-showcase/tree/main/schema/google/showcase/v1beta1) as well? (or since this is legacy behavior, we could add fragments tests instead as suggested by @leahecole )

chalmerlowe:
> LGTM. Holding off on formal approval until [aip-dev/google.aip.dev#1496](https://github.com/aip-dev/google.aip.dev/pull/1496) is merged and tests in cl/741139812 pass.
> 
> Should we consider updating [showcase protos](https://github.com/googleapis/gapic-showcase/tree/main/schema/google/showcase/v1beta1) as well? (or since this is legacy behavior, we could add fragments tests instead as suggested by @leahecole )

Re: cl/741139812
What do I need to do to move this along? It appears to be blocking.

Re: https://github.com/aip-dev/google.aip.dev/pull/1496 ... I accepted Noah's suggestion. Is there more needed from me?

tests/fragments/test_pagination_max_results_and_wrapper.proto
```diff
@@ -0,0 +1,36 @@
+// Copyright (C) 2021  Google LLC
```
parthea, line 1:
```suggestion
// Copyright (C) 2025  Google LLC
```

tests/fragments/test_pagination_page_size.proto
```diff
@@ -0,0 +1,36 @@
+// Copyright (C) 2021  Google LLC
```
parthea, line 1: This fragment proto is redundant because we're already running showcase tests which uses this proto with `page_size`: https://github.com/googleapis/gapic-showcase/blob/5a47589b6ef373fcd55c2d5545b0a902ea570dd9/schema/google/showcase/v1beta1/echo.proto#L289C9-L289C27

parthea (commented):
LGTM, Please also add a test to https://github.com/googleapis/gapic-generator-python/blob/main/gapic/templates/tests/unit/gapic/%25name_%25version/%25sub/test_%25service.py.j2

tests/fragments/test_pagination_page_size.proto
chalmerlowe: Good call. Will remove.

chalmerlowe:
> LGTM, Please also add a test to https://github.com/googleapis/gapic-generator-python/blob/main/gapic/templates/tests/unit/gapic/%25name_%25version/%25sub/test_%25service.py.j2

@parthea What are your thoughts on adding a new macro OR modifying an existing macro to the test_macros.p2 file instead of adding the test to test_service.py.j2? 

I modeled my test off of one of these types of macros:

https://github.com/googleapis/gapic-generator-python/blob/629cf19430d376273e9761eba40870ddeedcd549/gapic/templates/tests/unit/gapic/%25name_%25version/%25sub/test_macros.j2#L686C1-L757C12

```
{% if method.paged_result_field %}
{% if not method.paged_result_field.map %}
def test_{{ method_name }}_pager(transport_name: str = "grpc"):
    client = {{ service.client_name }}(
        transport=transport_name,
    )

    with mock.patch.object(
            type(client.transport.{{ method.transport_safe_name|snake_case }}),
            '__call__') as call:
        # Set the response to a series of pages.
        call.side_effect = (
            {{ method.output.ident }}(
                {{ method.paged_result_field.name }}=[
                    {{ method.paged_result_field.type.ident }}(),
                    {{ method.paged_result_field.type.ident }}(),
                    {{ method.paged_result_field.type.ident }}(),
                ],
                next_page_token='abc',
            ),
            {{ method.output.ident }}(
                {{ method.paged_result_field.name }}=[],
                next_page_token='def',
            ),
            {{ method.output.ident }}(
                {{ method.paged_result_field.name }}=[
                    {{ method.paged_result_field.type.ident }}(),
                ],
                next_page_token='ghi',
            ),
            {{ method.output.ident }}(
                {{ method.paged_result_field.name }}=[
                    {{ method.paged_result_field.type.ident }}(),
                    {{ method.paged_result_field.type.ident }}(),
                ],
            ),
            RuntimeError,
        )

        # REDACTED FOR SPACE
        pager = client.{{ method_name }}(request={}, retry=retry, timeout=timeout)

        # REDACTED FOR SPACE
        results = list(pager)
        assert len(results) == 6
        assert all(isinstance(i, {{ method.paged_result_field.type.ident }})
                   for i in results)
{% endif %}
```

parthea:
Can we tweak the [existing test](https://github.com/googleapis/gapic-generator-python/blob/629cf19430d376273e9761eba40870ddeedcd549/gapic/templates/tests/unit/gapic/%25name_%25version/%25sub/test_macros.j2#L688) to also handle the legacy pagination?

gapic/schema/wrappers.py
chalmerlowe: @leahecole 

Resolved.

gapic/schema/wrappers.py
chalmerlowe: @parthea:
Resolved

gapic/schema/wrappers.py
chalmerlowe: @parthea: resolved.

tests/unit/schema/wrappers/test_method.py
chalmerlowe: @parthea

Resolved.

tests/fragments/test_pagination_page_size.proto
chalmerlowe: @parthea 

Resolved, removed file.

chalmerlowe:
> Can we tweak the [existing test](https://github.com/googleapis/gapic-generator-python/blob/629cf19430d376273e9761eba40870ddeedcd549/gapic/templates/tests/unit/gapic/%25name_%25version/%25sub/test_macros.j2#L688) to also handle the legacy pagination?

The existing test created by macro in `test_macros.j2` executes if `method.paged_result_field` exists (i.e. is not `None`). Beyond that, it does not use or reference anything related to the current OR legacy pagination fields besides `next_page_tokens`.

What it does do is automatically execute *Requests and extracts content from *Responses to create a single list from three sets of rows (pages) gathered sequentially using `next_page_tokens`. It confirms that the final list is the correct length (signifying the paging process succeeded).

This macro generates unit tests using this macro successfully and those tests execute successfully.

gapic/schema/wrappers.py
```diff
@@ -1830,13 +1830,36 @@ def ident(self) -> metadata.Address:
...
+        """
+
+        pb_type = page_field_size.type
+
+        if pb_type == int or (
+            isinstance(pb_type, MessageType)
+            and pb_type.message_pb.name in {"UInt32Value", "Int32Value"}
+        ):
+            return True
+        return False
```
parthea, line 1849:
```suggestion
        return pb_type == int or (
            isinstance(pb_type, MessageType)
            and pb_type.message_pb.name in {"UInt32Value", "Int32Value"}
        )
```

parthea (approved):

PR #2373, opened 2025-03-28
refactor: reduce code duplication

parthea:
Reduce code duplication in `gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc.py.j2`  and `gapic/templates/%namespace/%name_%version/%sub/services/%service/transports/grpc_asyncio.py.j2`

### DIFF
--- a/tests/integration/goldens/asset/google/cloud/asset_v1/services/asset_service/transports/grpc_asyncio.py
+++ b/tests/integration/goldens/asset/google/cloud/asset_v1/services/asset_service/transports/grpc_asyncio.py
@@ -79,6 +79,7 @@ class _LoggingClientAIOInterceptor(grpc.aio.UnaryUnaryClientInterceptor):  # pra
                     "metadata": grpc_request["metadata"],
                 },
             )
+
         response = await continuation(client_call_details, request)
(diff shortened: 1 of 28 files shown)


tests/integration/goldens/asset/google/cloud/asset_v1/services/asset_service/transports/grpc_asyncio.py
```diff
@@ -79,6 +79,7 @@ async def intercept_unary_unary(self, continuation, client_call_details, request
                     "metadata": grpc_request["metadata"],
                 },
             )
+
```
ohmayr, line 82: can we remove this? Given that we're just refactoring macros, I'd expect the generated code to stay the same.

ohmayr (approved):

tests/integration/goldens/asset/google/cloud/asset_v1/services/asset_service/transports/grpc_asyncio.py
parthea: Fixed in https://github.com/googleapis/gapic-generator-python/pull/2373/commits/107fea49ae8e2889e1d2b502942483cbb7894b6a

PR #2423, opened 2025-08-13
chore: depend on stable version of google.shopping.type

vchudnov-g:


vchudnov-g:
OK, l `google-shopping-type` 1.0.0 has been published to PyPI https://pypi.org/project/google-shopping-type/1.0.0/, so this PR is unblocked. We should merge and release the generator.

parthea (approved):

PR #2426, opened 2025-08-27
fix: fix presubmit failures due to async tests

vchudnov-g:
This fixes a generated test that was causing pre-submits to fail because the tests itself is async (and tests async code) but it  wasn't properly marked in order for the test infrastructure to treat it properly.


ohmayr:
**Notice: Migration in Progress**

This repository is being consolidated into the google-cloud-python mono-repo. We are closing this PR here to prevent "split-brain" development during the transition.

**What happens next?**

- Your work has been captured in our migration snapshot.

- We are evaluating all active PRs for manual porting into the new location: packages/gapic-generator-python/.

- You can track the porting status and the technical migration plan here: https://github.com/googleapis/google-cloud-python/issues/16182

PR #2438, opened 2025-09-16
chore: add samples directory to lint paths

parthea:
Add `samples` directory to list of paths to lint


parthea:
Needs fix from https://github.com/googleapis/gapic-generator-python/pull/2441

gkevinzheng (approved):

PR #2454, opened 2025-09-30
fix: deprecate credentials_file

parthea:
See https://github.com/googleapis/google-auth-library-python/pull/1812


chalmerlowe (approved):
LGTM

PR #2474, opened 2025-11-03
tests: use older google-api-core when running tools with older runtimes

parthea:
Warnings emitted from google-api-core starting in `2.28.0` appear to cause issues when running `protoc`. The specific failure is `Plugin output is unparseable`

This PR adds a temporary workaround to use older version of `google-api-core` to unblock development while we identify a long term fix for https://github.com/googleapis/google-cloud-python/issues/16212

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -175,6 +175,13 @@ def fragment(session, use_ads_templates=False):
     )
     session.install("-e", ".")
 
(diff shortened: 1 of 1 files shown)


chalmerlowe (approved):
LGTM. With the understanding that this is to unblock some key work and the underlying issue will be resolved in another PR.

PR #2491, opened 2025-11-27
chore: remove code formatting for noxfile.py

parthea:
We should ensure that generated code is formatted correctly, rather than relying on external code formatting which increases the overall generation time.

This PR removes code formatting as a separate step for `noxfile.py`. There were only 3 changes in the file needed to address formatting issues. 

We already have a `goldens-lint` check which checks that the code is formatted properly. This PR avoid the extra step of formatting `noxfile.py` prior to running the lint check.

IOW, the `blacken` nox session no longer touches `noxfile.py`

https://github.com/googleapis/gapic-generator-python/blob/f93a0eb9bc88bb20c7a1540a33cf9f2fc12764f6/.github/workflows/tests.yaml#L409-L414


chalmerlowe (approved):
LGTM

PR #2504, opened 2025-12-09
chore: specify verify_format=False when running pypandoc

### LINKED ISSUE #2503: Invoke pandoc more efficiently
Currently every call to pypandoc.convert_text runs pandoc three times:

- `pandoc --list-input-formats`
- `pandoc --list-output-formats`
- `pandoc --from... --to...` (actual conversion)

Invoking pandoc three times as often as we really need to is a significant performance hit.

jskeet:
Without this, *every* call to pypandoc.convert_text starts pandoc three times: once to list the input formats, once to list the output formats, and once to actually perform the conversion.

With `verify_format=False`, the first two invocations are skipped. For some APIs this can make a huge difference to the speed of generation (listerally about 45% in one test I've performed). We shouldn't need to verify that the formats are supported, as we always use commonmark and rst.

Fixes #2503.

### DIFF
--- a/gapic/utils/rst.py
+++ b/gapic/utils/rst.py
@@ -61,6 +61,7 @@ def rst(
                 text,
                 "rst",
                 format=source_format,
+                verify_format=False,
                 extra_args=["--columns=%d" % (width - indent)],
             )
             .strip()


jskeet:
I'm seeing a potential issue when generating dialogflow-cx, but it may be unrelated - trying to reproduce now. (I don't understand how this would cause a problem... if it works *with* the validation, it should definitely work without it...)

jskeet:
> I'm seeing a potential issue when generating dialogflow-cx, but it may be unrelated - trying to reproduce now. (I don't understand how this would cause a problem... if it works _with_ the validation, it should definitely work without it...)

I believe now that the error was due to a file under output/ that hadn't been deleted when it should have been (by me).

parthea (approved):

PR #2531, opened 2026-01-21
chore: cache proto context to optimize generator performance

ohmayr:
This PR implements a thread-safe cache for with_context calls, significantly reducing redundant calculations during template rendering.

Changes:

- Performance: Memoizes with_context results, improving generation speed significantly.

- Safety Strategy (Object Pinning): Since Proto objects are unhashable, the cache uses id(self) as the key. To prevent "Zombie ID" collisions (where Python reuses the memory address of a garbage-collected object), the cache explicitly pins the object instance in memory (cache[key] = (result, self)).

- Lifecycle: Managed via a generation_cache_context using threading.local, ensuring the cache is active only during the build/render phases and cleaned up immediately after.

Dialog-cx generation time (current)
```
cat /tmp/librarian-3410144122/timing.txt
Processed 1 libraries in 11m32.59s; average=11m32.59s
google-cloud-dialogflow-cx: 11m32.59s
```

Dialogflow-cx generation (time with this change):
```
cat /tmp/librarian-1164720254/timing.txt
Processed 1 libraries in 3m2.296s; average=3m2.296s
google-cloud-dialogflow-cx: 3m2.296s
```

### DIFF
--- a/gapic/cli/generate.py
+++ b/gapic/cli/generate.py
@@ -23,6 +23,7 @@ from google.protobuf.compiler import plugin_pb2
 from gapic import generator
 from gapic.schema import api
 from gapic.utils import Options
+from gapic.utils.cache import generation_cache_context
 
 
 @click.command()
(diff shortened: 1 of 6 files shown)


daniel-sanche (commented):
changes mostly LGTM, but I'd like to see a benchmark before merging, to make sure we are getting the speedup we expect

parthea (approved):

daniel-sanche (approved):

PR #2539, opened 2026-01-27
chore: replace isort and black with ruff

ohmayr:
This PR replaces isort and black with ruff. This work is targeted towards improving the generation performance of our client libraries via librarian.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -31,10 +31,12 @@ nox.options.error_on_missing_interpreters = True
 
 showcase_version = os.environ.get("SHOWCASE_VERSION", "0.35.0")
 ADS_TEMPLATES = path.join(path.dirname(__file__), "gapic", "ads-templates")
-BLACK_VERSION = "black==25.1.0"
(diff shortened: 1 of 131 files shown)


noxfile.py
```diff
@@ -80,6 +82,7 @@ def unit(session):
                 path.join("tests", "unit"),
             ]
         ),
+        external=True,
```
daniel-sanche: was this meant to be part of this PR?

noxfile.py
```diff
@@ -294,7 +297,7 @@ def showcase_library(
                     }
                 ]
                 update_service_yaml = _add_python_settings(tmp_dir, python_settings)
-                session.run("python", "-c" f"{update_service_yaml}")
+                session.run("python", f"-c{update_service_yaml}")
```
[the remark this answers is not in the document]
ohmayr: I think there's no comma between the two so it is still considered as just one string i.e. no space? I believe it should work like it did before. Or maybe it didn't work before as well.

noxfile.py
ohmayr: Good catch. I'll clean this up!

daniel-sanche (approved):
LGTM

PR #2553, opened 2026-01-30
chore: use import absolute.path as module syntax for pb2 imports

ohmayr:
This PR updates how we generate the import path for _pb2 files and api-core. This work is meant to unblock the migration from black/isort to ruff.

### DIFF
--- a/gapic/schema/imp.py
+++ b/gapic/schema/imp.py
@@ -26,11 +26,25 @@ class Import:
         return self.package == other.package and self.module == other.module
 
     def __str__(self) -> str:
-        answer = f"import {self.module}"
+        # Determine if we need to suppress type checking for this import.
+        # We do this for protobuf generated files (_pb2) and api_core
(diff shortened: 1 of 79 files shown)


gapic/schema/imp.py
```diff
@@ -26,11 +26,21 @@ def __eq__(self, other) -> bool:
         return self.package == other.package and self.module == other.module
 
     def __str__(self) -> str:
+        # Determine if we need to suppress type checking for this import.
+        # We do this for protobuf generated files (_pb2) and api_core
+        # internals where type information might be missing or incomplete.
```
daniel-sanche: Do we really need to do this in the generated code? Can't we configure it in mypy.ini?

Edit: Looking closer, it looks like this logic was already in place, you're just moving it around. Do you know if mypy.ini is an option though?

daniel-sanche (approved):
LGTM

PR #2562, opened 2026-02-07
chore: fix mypy issue for `__func__` type

ohmayr:
- `ignore_missing_imports = True`: Third-Party Protos: Many of our generated libraries depend on common protos (like google.iam.v1). These upstream packages often do not include a py.typed marker file (defined in PEP 561). Without this flag, MyPy refuses to use those imports, even if the package is correctly installed in the Nox environment.

- switch the deprecated `blacken` with the new `format` session to ensure that mypy runs successfully after we format our code with `ruff`.

- wrap `sa_info_func` and `sa_file_func` to ensure that `# type: ignore` correctly silences the mypy type warnings.

### DIFF
--- a/gapic/templates/mypy.ini.j2
+++ b/gapic/templates/mypy.ini.j2
@@ -1,13 +1,12 @@
 [mypy]
 python_version = 3.14
 namespace_packages = True
+ignore_missing_imports = False
 
-# FIX: Prevents failure when dependencies (like google-iam) lack type hints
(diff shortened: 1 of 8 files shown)


parthea:
`//tests/integration:redis_selective                                      FAILED in 0.0s`

@ohmayr PTAL

daniel-sanche (approved):
LGTM

PR #2577, opened 2026-03-10
fix(docs): async samples missing await

daniel-sanche:
b/488298938

There's a bug in the samples for async operations. There should be two awaits: 
1. awaiting the method that returns the operation
2. awaiting the operation.result


daniel-sanche:
> https://github.com/googleapis/googleapis-gen/blob/master/google/storage/control/v2/storage-control-v2-py/samples/generated_samples/storage_v2_generated_storage_control_rename_folder_async.py
> 
> We need a change for rename_folder HNS method sample as well, is this getting covered as part of this change?

Yes, that will come in the next stage. This is updating the generator, which will then re-generate the code for all libraries in the next update.

Is there a specific timeline you need this addressed by? It could take a couple weeks for all the automated releases to roll out, but we can try to escalate if needed

vchudnov-g (approved):

